PR #38, opened 2022-10-13
mriqc run and evaluate scripts

InesGP:
@nikhil153


workflow/proc_pipe/scripts/eval_mriqc_results.py
```diff
@@ -0,0 +1,58 @@
+import sys
+import json
+import glob
+import pandas as pd
+
+
+def eval(args): #config file
```
jbpoline, line 7: you are re-defining python eval builtin function ? is that a concious decision ? :)

PR #70, opened 2023-03-17
Workflow to update downstream dataset repos

michellewang:
Tentative workflow to make a PR to the main branch of dataset-specific repos. The workflow will be triggered whenever a commit is pushed to the `main` branch of the main `mr_proc` repo. Since the `main` branch is protected, this means that it will trigger when PRs are merged to `main`.

Need to test this on a toy repo before merging.


michellewang:
Successfully run in [toy repo](https://github.com/neurodatascience/mini_proc), see [created PRs](https://github.com/neurodatascience/mini_proc-ppmi/pulls?q=is%3Apr).

michellewang:
After some further experimentations with the toy repos, I think this should work okay now. The git history should be relatively clean as long as we use merge commits (instead of squashing or rebasing).

.github/workflows/update_dataset_repos.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          repository: ${{ matrix.fork }}
+      - name: Get upstream changes
+        run: |
+          git remote add upstream https://github.com/neurodatascience/mr_proc.git
+          git pull upstream main
+          git fetch upstream main:upstream-main
+          git reset --hard upstream-main
```
jbpoline, line 23: while a git reset --hard is always attractive, not sure I see if it is actually  needed just after the fetch upstream main:upstream-main : will that branch pre-exists and may contain some stagged or non committed files ? or some commits ahead of upstream main ?  or is it to make sure everything is clean there ?

nikhil153: I think it's because `neurodatascience/mr_proc-ppmi` and `neurodatascience/mr_proc-qpn` will have additional dataset  specific files (consequently commits) which are not in `neurodatascience/mr_proc`.
@michellewang correct me if that's not the reason!

michellewang: @jbpoline @nikhil153 I could be wrong (I haven't used GitHub Actions before and am still not too familiar), but I think the `actions/checkout@v3` action creates a fresh clone of the repository, so there shouldn't be any staged or non-committed files. There might be commits ahead of upstream main though because the forks (`neurodatascience/mr_proc-ppmi` etc.) will have additional dataset-specific files, like @nikhil153 said. In that case the `reset --hard` still makes sense (I think) because we only want to get the changes that were added to `neurodatascience/mr_proc`, not the PPMI/QPN forks. Then it will create/update a PR to add these changes to `neurodatascience/mr_proc-ppmi`/`neurodatascience/mr_proc-qpn`, and we will have to manually resolve merge conflicts if there are any, then merge to the main branch of the fork.

I've tried removing the `git reset --hard upstream-main` line and it didn't work, the PRs got closed automatically without merging (not 100% sure why). There might still be other ways to achieve the same results without `git reset --hard`.

jbpoline: I see (I think :) ) - the workflow makes sense to me !

michellewang:
I have added a step for automatic approval of the PR created by this action. Note that this requires a second Personal Access Token (which hasn't been added as a GitHub secret to this repo yet), because the person who created a PR is not allowed to approve it (on GitHub the button is visible but can't be clicked). Will look into auto-merging (provided no merge conflicts) later this week.

Test repo is still [`neurodatascience/mini_proc`](https://github.com/neurodatascience/mini_proc). New commits to `main` will automatically create/update PRs in [`mini_proc-ppmi`](https://github.com/neurodatascience/mini_proc-ppmi) and [`mini_proc-qpn`](https://github.com/neurodatascience/mini_proc-qpn).

bcmcpher:
As long as the request is merging onto `main-upstream` and not `main` directly I think removing that rule should work fine. Commits from upstream can still be deliberately merged into `main` from the upstream branch. I think that protection was deliberately redundant. 

Whatever modification can still be tested in the upstream branch before merging to `main`, lowering the risk of an automatic action breaking a working processing environment (I'm assuming everyone is processing on the `main` branch of the project repos when they clone them.).

michellewang:
> As long as the request is merging onto `main-upstream` and not `main` directly I think removing that rule should work fine.

@bcmcpher 
* `main-upstream` is updated directly to match `neurodatascience/mr_proc:main` exactly

michellewang:
> As long as the request is merging onto main-upstream and not main directly I think removing that rule should work fine.

@bcmcpher the PR would actually be merged onto `main`. `main-upstream` is the head branch from which the changes proposed by the PR come from. Here is what happens:
1. `neurodatascience/mr_proc-[dataset]/main-upstream` is updated (and created if necessary) directly to match `neurodatascience/mr_proc:main` exactly
2. A PR is created to push the changes on `neurodatascience/mr_proc-[dataset]/main-upstream` to `neurodatascience/mr_proc-[dataset]/main`
3. The PR is approved/merged automatically by GitHub Actions

I don't think merging to `main` automatically would be a big issue because even if the `main` branch on GitHub is updated, whatever's on people's local clones isn't updated automatically until they run `git pull`. Since the common code and the dataset-specific code shouldn't interact in any way it would be unlikely that the `mr_proc` prime code breaks anything (I think?). Although one exception that I can think of is if we make big changes to the directory structure: that might break things if there are hardcoded paths or something like that. 

> Whatever modification can still be tested in the upstream branch before merging to main, lowering the risk of an automatic action breaking a working processing environment

Testing on `main-upstream` would be the same as testing in `mr_proc` prime since it's an exact copy (it doesn't contain the dataset-specific changes). We could merge `main-upstream` to another branch that contains the dataset-specific code, then test it there, but this would require a manual merge to `main` eventually, and I think @nikhil153 wanted to avoid that because if we have a lot of dataset forks it would be tedious to merge the PR in all the forks.

nikhil153:
@michellewang Does disabling the branch protection rule **"Do not allow bypassing the above settings"**  will allow anyone in `neurodatascience` to push directly to mr_proc-[dataset]/main from their local clone? (e.g. Mohammad working on QPN analysis)? 
If that's not the case and it's just us three that can do that, then let's go ahead and disable it.

michellewang:
@nikhil153 it'll only be people with "Admin" role (so the three of us) according to the [GitHub docs](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization#permissions-for-each-role):

Screenshot 2023-03-22 at 6 36 18 PM

> If that's not the case and it's just us three that can do that, then let's go ahead and disable it.

Ok I've disabled it on the PPMI and QPN forks. `mr_proc(*)` still has that rule but there's no need to disable it right now.

nikhil153 (approved):

nikhil153:
@michellewang sounds good! I have reviewed the PR. i will let you merge it if you are happy with the final changes!

PR #73, opened 2023-04-01
refactored run script of fmriprep to remove shell script dependency

nikhil153:


michellewang:
@nikhil153 looks good at a first glance! Since the command takes the participant ID how are we supposed to loop over all subjects we want to process? I know the plan is to eventually use Dask jobqueue for job submissions, but would we still need another script/function to get the participant ID from $SGE_TASK_ID (or $SLURM_ARRAY_TASK_ID)?

nikhil153:
@michellewang - yes currently there are template sge / slurm scripts in the (hpc)[https://github.com/neurodatascience/mr_proc/tree/main/workflow/proc_pipe/fmriprep/scripts/hpc] subdir within fmriprep workflow. I will open another PR once I update these / dask-queue gets implemented.

jbpoline:
Pinging @jacobsanz97 ?

michellewang (approved):

PR #75, opened 2023-04-03
migrated old qpn branch-specific code to new qpn fork

nikhil153:


michellewang:
@nikhil153 looks like this PR's target is [neurodatascience/mr_proc:main](https://github.com/neurodatascience/mr_proc), not [neurodatascience/mr_proc-qpn:main](https://github.com/neurodatascience/mr_proc-qpn). Is this a mistake? I think GitHub automatically chooses the original repo for PRs, so we need to choose the right fork when creating the PR.

PR #86, opened 2023-05-11
add BIDS subject/session ID conversion functions

michellewang:
* For use when:
  * generating manifest
  * doing DICOM->BIDS conversion
  * etc.
* Adding this to `mr_proc` template repo since these can (should) be used for all datasets
* Should be merged before https://github.com/neurodatascience/mr_proc-ppmi/pull/2


michellewang:
Closing because the changes are included in the (bigger) PR #89.

PR #89, opened 2023-05-16
Refactor DICOM-to-BIDS conversion to use `doughnut.csv`

michellewang:
Keep track of DICOM download/conversion status in a separate file (`<ROOT>/scratch/raw_dicom/doughnut.csv`) instead of the manifest.

New workflow for adding/converting new subjects/sessions:
- Update manifest (populate all columns) (**dataset-specific**)
- Run `<ROOT>/workflow/dicom_org/check_dicom_status.py` to append rows to doughnut file
  - New rows have empty `participant_dicom_dir` and populated `dicom_id`/`bids_id` columns. The binary status columns (`downloaded`, `organized`, `converted`) all have `False`
- **Dataset-specific** script (ideally `<ROOT>/workflow/dicom_org/fetch_dicom_downloads.py`) to fetch/download new DICOMs. This script should modify the doughnut file by:
  - Populating the `participant_dicom_dir` field
  - Setting `downloaded` to `True` for newly fetched/downloaded DICOMs
- `<ROOT>/workflow/dicom_org/dicom_org.py` re-organizes the DICOMs into flat lists
  - The default is now to create symlinks (relative paths so that they work in a container)
  - Sets `organized` to `True` in doughnut file
- `<ROOT>/workflow/bids_conv/run_bids_conv.py` does BIDS conversion with Heudiconv
  - Sets `converted` to `True` in doughnut file after **Heudiconv Stage 2** (not tested yet)

It is also possible to (re-)generate the doughnut file using `<ROOT>/workflow/dicom_org/check_dicom_status.py --regenerate`. This looks at files on disk and can be very slow
- Since `participant_dicom_dir` is dataset-specific (cannot always be inferred from `participant_id`), the user needs to provide a conversion function `<ROOT>/workflow/dicom_org/dicom_dir_func/participant_id_to_dicom_dir`. If this function is not provided, the `downloaded` status will not be checked (but the `organized` and `converted` statuses will still be checked)
  - There is a `sample_dicom_dir_func.py` for reference. It can be used for cases where `participant_dicom_dir` is just `participant_id`

Should be merged before [neurodatascience/mr_proc-ppmi#2](https://github.com/neurodatascience/mr_proc-ppmi/pull/2) because the conversion functions in `<ROOT>/workflow/utils.py` are used to create the manifest


workflow/catalog.py
```diff
@@ -1,47 +1,38 @@
...
-DATATYPE = "datatype"
-DICOM_ID = "dicom_id"
-BIDS_ID  = "bids_id"
-
-def read_manifest(manifest_csv, session_id, logger):
+
+from workflow.utils import (
+    COL_BIDS_ID_MANIFEST,
+    COL_CONV_STATUS,
+    COL_DICOM_DIR,
```
nikhil153, line 8: I had the `PARTICIPANT_DICOM_DIR` to avoid confusion with the parent `dicom` dir in the mr_proc dir_tree. Is there a reason you prefer otherwise?

workflow/utils.py
```diff
@@ -0,0 +1,116 @@
...
+COL_DATATYPE_MANIFEST = 'datatype'
+COLS_MANIFEST = [COL_SUBJECT_MANIFEST, COL_BIDS_ID_MANIFEST, COL_VISIT_MANIFEST, 
+                 COL_SESSION_MANIFEST, COL_DATATYPE_MANIFEST]
+
+# status file columns
+COL_DICOM_DIR = 'participant_dicom_dir'
+COL_DICOM_ID = 'dicom_id'
+COL_DOWNLOAD_STATUS = 'downloaded'
+COL_ORG_STATUS = 'organized'
+COL_CONV_STATUS = 'converted'
```
nikhil153, line 37: why not make these `True` or `False`?

workflow/utils.py
```diff
@@ -0,0 +1,116 @@
...
+def subject_to_bids(subject_id):
+    # add BIDS prefix if it doesn't already exist
+    subject_id = str(subject_id)
+    if subject_id.startswith(BIDS_SUBJECT_PREFIX):
+        return subject_id
+    else:
+        return f'{BIDS_SUBJECT_PREFIX}{subject_id}'
+
+def participant_id_to_bids_id(participant_id):
+    return dicom_id_to_bids_id(participant_id_to_dicom_id(participant_id))
```
nikhil153, line 60: I think having both of these is a bit confusing in terms of whats `subject` and what's `participant`. 
We should either go with direct name transformations or the chained ones that forces participant_id --> dicom_id --> bids_id.

workflow/utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+def session_to_bids(session_id):
```
nikhil153, line 62: For consistency, I would name this as `session_id_to_bids_session`

nikhil153 (changes requested):

workflow/utils.py
michellewang: These are just the column names, the actual values are `True` or `False`

workflow/catalog.py
michellewang: No particular reason, I think I was just trying to shorten the variable name but since that can be confusing with the `dicom` directory then I'll change `COL_DICOM_DIR` to `COL_PARTICIPANT_DICOM_DIR`

workflow/utils.py
nikhil153: I think that could be confusing since we are trying to be as close to BIDS as possible. But if you have it in the code already, then we can keep it.

nikhil153 (approved):

PR #90, opened 2023-05-18
add run script / tracker files for TractoFlow

bcmcpher:
This is an initial working version.

This should function on the majority of the QPN / PPMI inputs.

There will probably need some adjustments for the defaults decisions to generalize well, but they should be easy to add.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "workflow/proc_pipe/tractoflow/tractoflow"]
+	path = workflow/proc_pipe/tractoflow/tractoflow
+	url = git@github.com:scilus/tractoflow.git
--- a/workflow/proc_pipe/tractoflow/run_tractoflow.py
+++ b/workflow/proc_pipe/tractoflow/run_tractoflow.py
@@ -1,0 +1,460 @@
+import argparse
+import os
+import shutil
+from pathlib import Path
+import json
+import subprocess
+import tempfile
+
+import numpy as np
+import nibabel as nib
+from bids import BIDSLayout
+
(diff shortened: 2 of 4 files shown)


workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -0,0 +1,460 @@
...
+    ## condition empty path
+    if rpe_out == None:
+        rpe_file = None
+    else:
+        rpe_file = Path(tmp_dir, rpe_out).joinpath()
+        
+    ## return the paths to the input files to copy
+    return(dmrifile, bvalfile, bvecfile, anatfile, rpe_file)
+
+def run_tractoflow(participant_id, global_configs, session_id, output_dir, use_bids_filter, logger=None):
```
nikhil153, line 317: Could we rename this to `run` - i think with `run_mr_proc.py` or `DaskQueue` calling these things in future, we may want standardized naming for the main "run" function that gets called externally.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "workflow/proc_pipe/tractoflow/tractoflow"]
```
nikhil153, line 1: Does this necessitate update in the `setup.cfg` for the repo?

nikhil153 (approved):

nikhil153:
@bcmcpher - Made a couple of small comments, but they can be pushed in the future commits. Feel free to merge this if this is ready to go on new datasets.

bcmcpher:
Merge through different cmmit

PR #91, opened 2023-05-24
Small leftover changes after refactoring of shell scripts into python run scripts and and trackers

nikhil153:


extractors/freesurfer/run_FS_utils.py
```diff
@@ -58,64 +59,107 @@ def run(FS_dir, participants_list, out_file, meas, fwhm, template):
...
     """
     parser = argparse.ArgumentParser(description=HELPTEXT)
-
-    parser.add_argument('--FS_dir', type=str, help='list of participants to process')
-    parser.add_argument('--participants_csv', type=str, help='list of participants to process')
-    parser.add_argument('--group', type=str, help='filter participants based on a specific group value in the csv')
-    parser.add_argument('--output_dir', type=str, help='out_file path for the processed / aggregated output')
+    parser.add_argument('--global_config', type=str, help='path to global config file for your mr_proc dataset', required=True)
+    parser.add_argument('--session_id', type=str, help='session_id')
+    parser.add_argument('--visit_id', type=str, help='visit_id')
```
michellewang, line 64: Are `session_id` and `visit_id` required? There is no `required=True` flag so if the user doesn't provide them the script will crash. Sometimes when I want to run a script for the first time I'll call it without arguments so that `argparse` prints the usage message

michellewang (approved):

extractors/freesurfer/run_FS_utils.py
nikhil153: Fixed! Thanks!

PR #101, opened 2023-06-19
Update sample global configs

### LINKED ISSUE #100: Update `sample_global_configs.json`
- [x] Add `"VISITS"` key
- [x] Add info for tabular "pipelines". Something like this:
```json
"TABULAR": {
    "DEMOGRAPHICS": {
        "SEX": {
            "VERSION": "",
            "FILENAME": ""
        },
        "DATE_OF_BIRTH": {
            "VERSION": "",
            "FILENAME": ""
        }
    },
    "ASSESSMENTS": {
        "UPDRS_3": {
            "VERSION": "",
            "FILENAME": ""
        },
        "MOCA": {
            "VERSION": "",
            "FILENAME": ""
        },
        "GROUP": {
            "VERSION": "",
            "FILENAME": ""
        }
    }
}
```

We should probably also update the documentation somewhere to clarify the difference between a "visit" and a "session"

michellewang:
- Add `"VISITS"` key
- Add `"TABULAR"` key with subsections `"DEMOGRAPHICS"` and `"ASSESSMENTS"` (matches tree structure)
  - Should be used to generate phenotypic data bagels (and maybe the manifest)

Closes #100.

### DIFF
--- a/sample_global_configs.json
+++ b/sample_global_configs.json
@@ -10,6 +10,7 @@
     "TEMPLATEFLOW_DIR": "",
 
     "SESSIONS": ["1","2"],
+    "VISITS": ["V01", "V02"],
 
     "BIDS": {
         "heudiconv": {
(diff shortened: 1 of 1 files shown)


sample_global_configs.json
```diff
@@ -10,6 +10,7 @@
     "TEMPLATEFLOW_DIR": "",
 
     "SESSIONS": ["1","2"],
+    "VISITS": ["V01", "V02"],
```
nikhil153, line 13: I wonder if we should stick with integers here or actual `V01`, `V02` strings etc.. since for PPMI at least we will match the `sessions` and `visits`. Thoughts?

nikhil153 (approved):

sample_global_configs.json
michellewang: I used different codes for visits just to show that they could be different, but if you want I can change it to be the same as sessions. I don't have a strong preference.

nikhil153: Good point. Okay let's keep them different

PR #111, opened 2023-07-11
Fix `dicom_org` issues and update `bids_conv`

michellewang:
`dicom_org`:
- fix typo in `check_dicom_status.py`
- fix logic error with `skip_dcm_check` in `utils.py`
- add shell scripts for DICOM downloading/unzipping/squashing in `shell_utils` (currently used for PPMI, may be used for other datasets in the future)

`run_bids_conv.py`:
- add `--overlay` flag for mounting SquashFs overlays (needed for e.g. PPMI)
- add `--fpaths_to_copy` flag for specifying files to copy to `<DATASET_ROOT>/proc` in addition to `heuristic.py` (needed e.g., because PPMI heuristic requires checking another PPMI file)
- some cleanup

Other:
- fix bug where `participant_dicom_dir` is `NA` for new participants (`catalog.py`)


workflow/dicom_org/shell_utils/make_squash.sh
nikhil153, line 1: This is really cool and very well written! 
Would it be easy to redirect the print statements to a log file?

nikhil153 (approved):

PR #121, opened 2023-08-01
fix subjects in derivatives bagel

michellewang:
Previously the trackers assumed that all sessions had the same participants. Now it uses the `session` column in the manifest to only select participants who have imaging data available for the relevant session


michellewang:
Closing: I merged this branch into #95 since that one is also about trackers.

PR #132, opened 2023-08-16
[MAINT] add test to run in ci for fmriprep workflow 

Remi-Gau:
Set up test with pytest to check that the correct command is created by run_fmriprep.

Had to do quite a a few changes to make this work.

- set up all the packaging using pyproject.toml (delete setup.py, setup.cfg)
- use hatchling instead of setup tools to install the package
- add extra dependencies in pyproject.toml for testing
- reorganize into a [source layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/) and adapt import statements accordingly: was needed to facilitate testing
- add tests for fmriprep workflow
- add GHA testing workflow


.github/workflows/tests.yml
```diff
@@ -0,0 +1,38 @@
...
+  cancel-in-progress: true
+
+jobs:
+
+  test:
+
+    runs-on: ubuntu-latest
+
+    # only trigger on upstream repo
+    # if: github.repository_owner == 'neurodatascience' && github.repository == 'nipoppy'
```
Remi-Gau, line 21:
```suggestion
    if: github.repository_owner == 'neurodatascience' && github.repository == 'nipoppy' 
```

tests/test_workflow_fmriprep.py
```diff
@@ -0,0 +1,122 @@
...
+    pth = pth / "freesurfer"
+    pth.mkdir(parents=True, exist_ok=True)
+    license_file = pth / "license.txt"
+    license_file.write_text("dummy content")
+
+
+def dummy_bids_filter_file() -> Path:
+    """TODO probably don't want the bids filter file to be in the module directory"""
+    pth = Path(fmriprep_module.__file__).parent
+    return pth / "bids_filter.json"
```
Remi-Gau, line 24: the default path for the bids_filter.json will probably have to be changed because if someone install the package with pip install but not in editable mode, it will be hard for this to work.

Remi-Gau:
- Good news: no need to install any data for testing
- Bad news: a lot of file moving require to make the testing happen

.gitmodules
```diff
@@ -1,3 +1,4 @@
-[submodule "workflow/proc_pipe/tractoflow/tractoflow"]
-	path = workflow/proc_pipe/tractoflow/tractoflow
-	url = git@github.com:scilus/tractoflow.git
+[submodule "nipoppy/workflow/proc_pipe/tractoflow/tractoflow"]
+	path = nipoppy/workflow/proc_pipe/tractoflow/tractoflow
+	url = https://github.com/scilus/tractoflow.git
+	datalad-url = https://github.com/scilus/tractoflow.git
```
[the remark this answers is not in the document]
bcmcpher: `main.nf` w/in the tractoflow repo is the analysis script that nextflow invokes when it runs. Nextflow is basically a generalized nipype w/ it's own syntax to wrap the commands. I have to point at `main.nf` and the container for this version - the newest release has the `*.nf` files w/in the container. This is the version running on UKBB, so preserving for posterity is desirable. Let me know if there's a more convenient way to package this together.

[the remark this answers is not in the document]
Remi-Gau: best to be sure that it will keep on working is to... write a test for it... 😉 

let me check but I am sure we can package git submodule into python packages #inception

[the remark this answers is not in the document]
Remi-Gau: @bcmcpher https://github.com/neurodatascience/nipoppy/pull/132/commits/f828e0118280e4d7f72965c0f2a7df12987e1dbc should fix it

tests/test_workflow_fmriprep.py
```diff
@@ -0,0 +1,122 @@
...
+    log_file = tmp_path / "fmriprep.log"
+
+    bids_dir = "bids_dir"
+    fmriprep_dir = tmp_path / "fmriprep_dir"
+    fs_dir = "fs_dir"
+    templateflow_dir = "templateflow_dir"
+    participant_id = "01"
+    singularity_container = "fmriprep.simg"
+
+    cmd = run_fmriprep(
```
Remi-Gau, line 77: to be fair I was surprised that no error occurred when just feeding it dummy values with no data and no container installed but it does simplify testing for now

nipoppy/workflow/bids_conv/run_bids_conv.py
michellewang, line 1: Import statements here need to be changed (`workflow` -> `nipoppy.workflow`)

nipoppy/workflow/dicom_org/check_dicom_status.py
michellewang, line 1: Here also `workflow` should be changed to `nipoppy.workflow` in the import statements

nipoppy/workflow/bids_conv/run_bids_conv.py
Remi-Gau: done

nipoppy/workflow/dicom_org/check_dicom_status.py
Remi-Gau: done

tests/test_workflow_fmriprep.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.mark.parametrize("anat_only", [True, False])
+def test_run_fmriprep(tmp_path, use_bids_filter, anat_only):
+    log_file = tmp_path / "fmriprep.log"
+
+    bids_dir = "bids_dir"
+    fmriprep_dir = tmp_path / "fmriprep_dir"
+    fs_dir = "fs_dir"
+    templateflow_dir = "templateflow_dir"
+    participant_id = "01"
+    singularity_container = "fmriprep.simg"
```
nikhil153, line 75: Should we add actual container names with versions that we are using. For instance `fmriprep_20.2.7.sif`

tests/test_workflow_fmriprep.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.mark.parametrize("use_bids_filter", [True, False])
+@pytest.mark.parametrize("anat_only", [True, False])
+def test_run_fmriprep(tmp_path, use_bids_filter, anat_only):
+    log_file = tmp_path / "fmriprep.log"
+
+    bids_dir = "bids_dir"
+    fmriprep_dir = tmp_path / "fmriprep_dir"
+    fs_dir = "fs_dir"
+    templateflow_dir = "templateflow_dir"
+    participant_id = "01"
```
nikhil153, line 74: Do expect to have bids formatted id?

nikhil153 (approved):
Minor suggestions for naming subjects and containers.

tests/test_workflow_fmriprep.py
Remi-Gau: well I do not think there are any checks on that at the moment, no?

tests/test_workflow_fmriprep.py
Remi-Gau: makes sense let me try someting

Remi-Gau: OK used your suggestion

tests/test_workflow_fmriprep.py
nikhil153: No because fmriprep doesn't require the `sub-` prefix. But I think we should standardize all "run script" calls to comply with the bids subject id naming.

Remi-Gau: probably a separate PR to make sure it is participant_id is alphanumeric

PR #133, opened 2023-08-17
Use `bids_id` in doughnut file only instead of manifest and doughnut

### LINKED ISSUE #116: Decide what to do about `bids_id` column in `manifest.csv`
The `check_dicom_status.py` script expects the `bids_id` column in the manifest to be populated, but @nikhil153 was under the impression that we wanted to add it ourselves. The reasoning is probably that we do not want users to have to generate `bids_id` themselves. However, I don't think we should change the manifest except to add new participants/sessions/datatypes. Maybe `participant_id` to `bids_id` conversion can just always be done on the fly (and written to bagels but nowhere else).

michellewang:
Closes #116.


nipoppy/workflow/dicom_org/check_dicom_status.py
```diff
@@ -52,7 +53,7 @@ def run(global_config_file, regenerate=False, empty=False):
     # load manifest
     fpath_manifest = dpath_dataset / FPATH_MANIFEST_RELATIVE
     df_manifest = load_manifest(fpath_manifest)
-    df_status = df_manifest.loc[~df_manifest[COL_BIDS_ID_MANIFEST].isna()].copy()
+    df_status = df_manifest.loc[~df_manifest[COL_SESSION_MANIFEST].isna()].copy()
```
nikhil153, line 56: Does this imply that `session_id` is the only way to tell if subject has imaging data in the manifest? 
It might be good to have additional check for `datatype` as well? And raise warnings if either of them are empty?

michellewang: Yeah I was also thinking of adding the check for `datatype`. At one point we said we only wanted `session_id` to be defined if the subject had imaging data -- is that still the case?

For the warning, does that mean there should be a warning if a subject doesn't have imaging data? Because that can happen fairly often in PPMI, especially for non-baseline visits/sessions

nikhil153 (approved):

PR #145, opened 2023-08-21
Fix bug where `doughnut.csv` loses rows after BIDS conversion

### LINKED ISSUE #131: `doughnut.csv` loses sessions at one point during DICOM-to-BIDS conversion
Need to identify where that happens + fix it.

michellewang:
Closes #131.


nipoppy/workflow/catalog.py
```diff
@@ -11,13 +11,14 @@
     load_status,
 )
 
-def read_and_process_status(status_csv, session_id, logger):
+def read_and_process_status(status_csv, session_id=None, logger=None):
```
nikhil153, line 14: Why would `session_id` ever be `None`?

nipoppy/workflow/dicom_org/check_dicom_status.py
```diff
@@ -170,6 +170,7 @@ def run(global_config_file, regenerate=False, empty=False):
         print(f'\nAdded {len(df_status_new_rows)} rows to existing status file')
 
     df_status = df_status[COLS_STATUS].drop_duplicates(ignore_index=True)
```
nikhil153, line 172: drop_duplicates with ignore index makes me a bit worried. I assume this is why you now have sort_values call after. It's a bit hard for me to see what's going on here exactly. But I trust your testing of this. But out of curiosity, would it be better if we use `.loc` based filtering instead?

nipoppy/workflow/catalog.py
michellewang: The bug that this PR is supposed to fix is due to `run_bids_conv.py` loading the doughnut for only a single session, then updating it and saving it (which makes it lose all the rows with other sessions). But I just realized that I should have used `nipoppy.workflow.utils.load_status()` instead 😅

nipoppy/workflow/dicom_org/check_dicom_status.py
michellewang: So at this point in the script the index is just row numbers (i.e. not `participant_id` or `session`). The reason why we need `ignore_index=True` is just so that the resulting dataframe has index 0 to N-1 (where N is number of rows) because I compare the new doughnut with the most recent one (to avoid writing exactly the same doughnut), and that comparison checks the index. So if we sort all the rows and set the index to 0...N-1 then it makes it possible to compare the two doughnuts, otherwise the row order (or arbitrary index) might not be the same.

About the `drop_duplicates()`, we might not actually need it. I added it when I noticed duplicate rows in the doughnut, but it turns out that this was because of a bug in my manifest generation script (will be fixed by [neurodatascience/nipoppy-ppmi#47](https://github.com/neurodatascience/nipoppy-ppmi/pull/47)). I can remove the `drop_duplicates()` call, but we might want to discuss if we want the script to fix things automatically when there's an issue with the manifest

nikhil153: Hmm I see, and we can't do this with pandas builtin functions: [equals](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.equals.html#pandas.DataFrame.equals) or [compare](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.compare.html)?

michellewang: Yes, that’s what I do on L176 (`pd.DataFrame.equals`), but the two dataframes need to have rows in the same order and the same index, which is why I sort and reset the index

nikhil153: Ahh I see. Okay this is becoming clearer to me.

nikhil153 (approved):
Maybe it can be another PR, but should we also change the default expectation of `dicom_dir` name - which currently depends on this [import](https://github.com/neurodatascience/nipoppy/blob/b0285b70f68881015af058362a5f32231c1bb829/nipoppy/workflow/dicom_org/check_dicom_status.py#L117)?

PR #146, opened 2023-08-21
Add GitHub Actions workflow to prevent downstream forks changing files tracked in main repo

### LINKED ISSUE #138: Add GitHub Actions workflow to make sure dataset forks do not change files that exist in template repo
To avoid future headaches

Outline:
- Do not run on `neurodatascience/nipoppy`
- Checkout dataset fork repo
- List all files that were modified in current PR (https://stackoverflow.com/questions/424071/how-do-i-list-all-the-files-in-a-commit)
- ~Make a temporary branch that is hard-synced to `neurodatascience/nipoppy:main`~
  - ~Or: use `git clone`? Not sure what would be the better option~
  - Checkout `neurodatascience/nipoppy:main`
- For each of these files, check if it exists in `neurodatascience/nipoppy:main`
  - If yes, fail run

michellewang:
Closes #138.

- Workflow will only run on repos under `neurodatascience` that **are not** called `nipoppy`
- The list of changed files is obtained by taking the diff between the base/target branches of the PR on the fork, then the script checks if any of those files exist in `neurodatascience/nipoppy` and fails if that is the case
- Will get failures if upstream changes from `neurodatascience/nipoppy` have been applied to the PR target branch but not the base branch; however, that shouldn't be a problem (in general) since the forks are synced to the template repo by another workflow. If the sync fails (due to merge conflicts) then it means there are issues that need to be addressed as soon as possible anyway, so putting other PRs on hold in the meantime seems reasonable


.github/workflows/check_changed_files.yml
```diff
@@ -0,0 +1,51 @@
...
+    check-changed-files:
+
+        runs-on: ubuntu-latest
+
+        env:
+            TEMPLATE_REPO: neurodatascience/nipoppy
+            PATH_CURRENT_REPO: repo_to_check
+            PATH_TEMPLATE_REPO: template_repo
+
+        if: github.repository_owner == 'neurodatascience' && github.event.repository.name != 'nipoppy'
```
nikhil153, line 19: are we matching exact repo name string or anything that begins with "nipoppy"?

.github/workflows/check_changed_files.yml
```diff
@@ -0,0 +1,51 @@
...
+          - name: Checkout template repository
+            uses: actions/checkout@v3
+            with:
+                repository: ${{ env.TEMPLATE_REPO }}
+                ref: main
+                path: ${{ env.PATH_TEMPLATE_REPO }}
+
+          - name: Check changed files
+            run: |
+                CHANGED_FILES=$(cd ${PATH_CURRENT_REPO}; git diff-tree --no-commit-id --name-only -r ${{ github.event.pull_request.head.sha }} ${{ github.event.pull_request.base.sha }})
```
nikhil153, line 38: Since I don't speak github actions dialect, just want to check if this looks for entire file path or just the filenames? Because we have multiple `READMEs` :)

.github/workflows/check_changed_files.yml
michellewang: Matching the exact repo string: I tested in a dummy repo (`testpoppy`) and the workflow is skipped in [testpoppy](https://github.com/neurodatascience/testpoppy/actions) but runs on [testpoppy-ppmi](https://github.com/neurodatascience/testpoppy-ppmi/actions)

.github/workflows/check_changed_files.yml
michellewang: This part is just like a regular shell script, `git diff-tree` gives full paths (relative to the root of the `git` repository) so it shouldn't be a problem. Tested it [here](https://github.com/neurodatascience/testpoppy-ppmi/pull/3) with the READMEs and it works fine :)

nikhil153 (approved):

michellewang:
Well, it turns out that didn't quite work because the auto-update workflow creates a PR on `main` in each of the forks 😅. We need to make the workflow only run on PRs that don't originate from the `main-upstream` branch

PR #162, opened 2023-08-24
Make backups of `bagel.csv` and check it when tracking pipelines (+ refactor status -> doughnut)

### LINKED ISSUE #159: Restore changes in `run_tracker.py`
> I think something went wrong in #95 when I merged from `main` (or another point) at one point, probably merge conflicts that I didn't resolve correctly. There are some changes in `run_tracker.py` that somehow got lost (at different stages)...

_Originally posted by @michellewang in https://github.com/neurodatascience/nipoppy/issues/148#issuecomment-1690937774_

- See [here](https://github.com/michellewang/nipoppy/blob/0bbfeaa3c4b847e741b5d5acd26c7e24583576b8/trackers/run_tracker.py) previous version that gets a session-specific participant list instead of the same one for all sessions
- See [here](https://github.com/michellewang/nipoppy/blob/399db9bfc690efe570bf06cbd1652dcda5298979/trackers/run_tracker.py) for previous version that loads existing bagel file and adds/updates the specified pipelines

Blocked by #148.

michellewang:
This is not quite done yet but I'm creating the PR so that I don't forget about it.

Restore lost changes:
- Load old `bagel.csv` if it exists and replace rows as needed
  - [x] Currently this only works at the pipeline level, need to extend it so that it replaces only the specified session (since we now have a `session_id` argument when before we always tracked all sessions)
- Fix bug where all subjects were assumed to have all sessions, which is not the case

Closes #159.

### DIFF
--- a/nipoppy/trackers/run_tracker.py
+++ b/nipoppy/trackers/run_tracker.py
@@ -137,7 +137,8 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
                         logger.info(f"task_name: {name}, status: {status}")
                         _df.loc[bids_id,name] = status
                         _df.loc[bids_id,"pipeline_starttime"] = get_start_time(subject_dir)
-                        _df.loc[bids_id,"pipeline_endtime"] = get_end_time(subject_dir)
+                        # TODO only check files listed in the tracker config
+                        _df.loc[bids_id,"pipeline_endtime"] = UNAVAILABLE # get_end_time(subject_dir)
                 else:
                     logger.error(f"Output for pipeline: {pipeline} not found for bids_id: {bids_id}, session: {session}")
                     for name in status_check_dict.keys():                    


nipoppy/workflow/utils.py
```diff
@@ -11,8 +11,10 @@
 # directory/file names
 DNAME_BACKUPS_MANIFEST = '.manifests'
 DNAME_BACKUPS_STATUS = '.doughnuts'
+DNAME_BACKUPS_BAGELS = '.bagels'
 FNAME_MANIFEST = 'manifest.csv'
 FNAME_STATUS = 'doughnut.csv'
+FNAME_BAGEL = 'bagel.csv'
```
nikhil153, line 17: maybe rename `STATUS` to `DOUGHNUT` to be consistent?

nipoppy/trackers/run_tracker.py
```diff
@@ -43,47 +48,76 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
...
-            status_check_dict = pipe_tracker.get_pipe_tasks(tracker_configs, PIPELINE_STATUS_COLUMNS, pipeline, version)
-
-            dash_col_list = list(schema["GLOBAL_COLUMNS"].keys()) 
-            
-            logger.info(f"Checking session: {session_id}")    
-            _df = pd.DataFrame(index=bids_participants, columns=dash_col_list)          
-            _df["session"] = session
-            _df["pipeline_name"] = pipeline        
+        # Grab BIDS participants from the doughnut
+        doughnut_file = f"{DATASET_ROOT}/scratch/raw_dicom/doughnut.csv"
```
nikhil153, line 61: import doughnut.csv name from utils?

nipoppy/trackers/run_tracker.py
```diff
@@ -43,47 +48,76 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
...
-    
-    proc_status_dfs = [] # list of dataframes
-    for session_id in session_ids:
+    logger.info(f"tracking session_ids: {session_ids}")    
 
-        session = session_id_to_bids_session(session_id)
+    for pipeline in pipelines:
+        pipe_tracker = Tracker(global_configs, dash_schema_file, pipeline) 
+        
+        mr_proc_root_dir, _, version = pipe_tracker.get_global_configs()
```
nikhil153, line 56: rename `mr_proc` :)

nipoppy/trackers/run_tracker.py
```diff
@@ -43,47 +48,76 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
...
+            n_participants_session = len(participants_session)
+            logger.info(f"n_participants_session: {n_participants_session}")
+
+            _df = pd.DataFrame(index=participants_session, columns=dash_col_list)
+            _df[COL_SESSION_MANIFEST] = session
+            _df["pipeline_name"] = pipeline
             _df["pipeline_version"] = version
+            _df["has_mri_data"] = TRUE # everyone in the status file has MRI data
+
+            tracker_csv = Path(mr_proc_root_dir, 'derivatives', FNAME_BAGEL)
```
nikhil153, line 91: rename `mr_proc` :)

nipoppy/trackers/run_tracker.py
```diff
@@ -43,47 +48,76 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
...
+            if tracker_csv.exists():
+                old_proc_status_df_full = load_bagel(tracker_csv)
+
+                old_proc_status_df_session = old_proc_status_df_full.loc[old_proc_status_df_full[COL_SESSION_MANIFEST] == session]
+                old_participants_session = set(old_proc_status_df_session[COL_BIDS_ID_MANIFEST])
+                old_pipelines_session = set(old_proc_status_df_session['pipeline_name'])
+                
+                # make sure the number of participants is consistent across pipelines
```
nikhil153, line 99: Not sure if this is before or after our discussion on querying doughnut regardless a full run or selective subset of participant run. But to me it should be possible to ensure number of participants is consistent across pipelines automatically. 
Th runtime error is likely to confuse the user more and trigger slow full runs unnecessarily.

nipoppy/trackers/run_tracker.py
```diff
@@ -97,29 +131,57 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
...
+
+            # add old rows from other pipelines/sessions and sort for consistent order
+            proc_status_df: pd.DataFrame = pd.concat([old_proc_status_df, _df], axis='index')
+            proc_status_df = proc_status_df.sort_values(["pipeline_name", "pipeline_version", COL_BIDS_ID_MANIFEST], ignore_index=True)
 
-    proc_status_df = pd.concat(proc_status_dfs, axis='index')
+            # don't write a new file if no changes
+            try:
+                if len(proc_status_df.compare(old_proc_status_df_full)) == 0:
+                    logger.info(f'\nNo change for pipeline {pipeline}, session {session}')
```
nikhil153, line 155: No change in status? Or `participant_id` x `session` combinations?

nipoppy/trackers/run_tracker.py
michellewang: 😅

nipoppy/trackers/run_tracker.py
michellewang: This was from before the discussion yesterday. I didn't add the feature where the user can specify specific subjects to run yet because I think it's better to have it in another PR. I think we can address the consistent participant thing in that other PR also? The only thing that that doesn't sit too well with me is that we might have out-of-sync tracking info for different pipelines for the same participants (ex: fMRIPrep is up-to-date but MRIQC is not because for some reason it wasn't run). So that's why I had the error, though I could change it to a warning for now

michellewang:
@nikhil153 addressed all your comments (I think). Also refactored some doughnut-related variable/function names which is why there are more files changes now

nipoppy/trackers/run_tracker.py
```diff
@@ -97,29 +133,57 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id=1,
                 
                 if dir_status:                
                     for name, func in status_check_dict.items():
-                        status = func(subject_dir, session, run_id)
+                        status = func(subject_dir, session_id, run_id)
                         logger.info(f"task_name: {name}, status: {status}")
                         _df.loc[bids_id,name] = status
                         _df.loc[bids_id,"pipeline_starttime"] = get_start_time(subject_dir)
-                        _df.loc[bids_id,"pipeline_endtime"] = UNAVAILABLE # TODO
+                        _df.loc[bids_id,"pipeline_endtime"] = get_end_time(subject_dir)
```
nikhil153, line 140: This could get very slow depending on number of files. Maybe we should either only check the files listed in the tracker config or parse the pipeline log to get completion timestamp.

nikhil153 (approved):

PR #165, opened 2023-09-07
add `session_id` to `participant_id_to_dicom_dir()`

### LINKED ISSUE #164: Handle `session_id` in `participant_id_to_dicom_dir.py` helper function used in `check_dicom_status.py`
@michellewang, 
Its unclear to me how the mapping function [participant_id_to_dicom_dir](https://github.com/neurodatascience/nipoppy/blob/main/nipoppy/workflow/dicom_org/check_dicom_status.py#L129-L131) is supposed to handle multiple sessions from the same participant. 

My guess is that either we have to pass the `session_id` in this call or change the lambda function approach.

michellewang:
Closes #164.

### DIFF
--- a/nipoppy/workflow/dicom_org/check_dicom_status.py
+++ b/nipoppy/workflow/dicom_org/check_dicom_status.py
@@ -126,8 +126,13 @@ def run(global_config_file, regenerate=False, empty=False):
                 'See sample_dicom_dir_func.py for an example.'
             )
 
-        df_status[COL_PARTICIPANT_DICOM_DIR] = df_status[COL_SUBJECT_MANIFEST].apply(
-            lambda participant_id: participant_id_to_dicom_dir(participant_id, global_config)
+        df_status[COL_PARTICIPANT_DICOM_DIR] = df_status.apply(
+            lambda row: participant_id_to_dicom_dir(
+                getattr(row, COL_SUBJECT_MANIFEST),
+                getattr(row, COL_SESSION_MANIFEST),
+                global_config,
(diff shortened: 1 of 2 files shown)


michellewang:
@nikhil153 the doughnut has `session` (e.g. `ses-01`), but I guess we really want the `session_id`? If so we need to strip the `ses-` prefix from the session

nipoppy/workflow/dicom_org/check_dicom_status.py
```diff
@@ -126,8 +127,13 @@ def run(global_config_file, regenerate=False, empty=False):
                 'See sample_dicom_dir_func.py for an example.'
             )
 
-        df_status[COL_PARTICIPANT_DICOM_DIR] = df_status[COL_SUBJECT_MANIFEST].apply(
-            lambda participant_id: participant_id_to_dicom_dir(participant_id, global_config)
+        df_status[COL_PARTICIPANT_DICOM_DIR] = df_status.apply(
+            lambda row: participant_id_to_dicom_dir(
```
nikhil153, line 131: Cool, didn't know you could just pass in a row!

nikhil153 (approved):

nipoppy/workflow/dicom_org/check_dicom_status.py
michellewang: Yes, turns out `apply` is not only for `Series`: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html

PR #166, opened 2023-09-14
update to tractoflow run for bval parsing

bcmcpher:
I checked the parsing with the bval/bvec files. This was ready to go, but I've improved the comments / logging so it's clear what's going on.

There was a bug in the max lmax check, but it wasn't impacting anything.

### DIFF
--- a/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
+++ b/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
@@ -459,7 +459,7 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
     ## build paths to working inputs
     tractoflow_input_dir = f"{tractoflow_dir}/input"
     tractoflow_nxtf_inp = f"{tractoflow_input_dir}/{participant_id}_ses-{session_id}"
-    tractoflow_subj_dir = f"{tractoflow_input_dir}/{participant_id}_ses-{session_id}/input"
+    tractoflow_subj_dir = f"{tractoflow_input_dir}/{participant_id}_ses-{session_id}/{participant_id}"
     tractoflow_work_dir = f"{tractoflow_dir}/work/{participant_id}_ses-{session_id}"
 
     ## check if working values already exist
@@ -586,7 +586,7 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
     NEXTFLOW_CMD=f"nextflow run /scilus_flows/tractoflow/main.nf -work-dir {tractoflow_work_dir} -with-trace {LOGDIR}/{participant_id}_ses-{session_id}_nf-trace.txt -with-report {LOGDIR}/{participant_id}_ses-{session_id}_nf-report.html"
     
     ## compose tractoflow arguments
-    TRACTOFLOW_CMD=f""" --input {tractoflow_nxtf_inp} --output_dir {tractoflow_out_dir} --run_gibbs_correction --eddy_cmd eddy_openmp --encoding_direction {phase} --readout {readout} --dti_shells "0 {dti_use}" --fodf_shells "0 {odf_use}" --sh_order {sh_order} --profile {profile} --processes {ncore}"""
+    TRACTOFLOW_CMD=f""" --input {tractoflow_nxtf_inp} --output_dir {tractoflow_out_dir} --run_gibbs_correction --encoding_direction {phase} --readout {readout} --dti_shells "0 {dti_use}" --fodf_shells "0 {odf_use}" --sh_order {sh_order} --profile {profile} --processes {ncore}"""
 
     ## TractoFlow arguments can be printed multiple ways that appear consistent with the documentation but are parsed incorrectly by nextflow.
     ## .nexflow.log (a run log that documents what is getting parsed by nexflow) shows additional quotes being added around the dti / fodf parameters. Something like: "'0' '1000'"
--- a/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
+++ b/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
@@ -447,6 +447,7 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
 
     ## create merged list of requested shells
     rshell = np.unique([ dti_use, odf_use ])
+    #rshell = np.unique([ dti_use + odf_use ]) # depends on how the data is passed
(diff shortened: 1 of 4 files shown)


nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -472,14 +473,15 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
         tdir = np.unique(tvec, axis=0)
 
         ## compute and print the maximum shell
-        plmax = int(np.floor((-3 + np.sqrt(1 + 8 * tdir.shape[1])) / 2.0))
+        plmax = int(np.floor((-3 + np.sqrt(1 + 8 * tdir.shape[1]) / 2.0)))
```
nikhil153, line 476: I think we need a bit more detail / explanation on this formula.

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -489,23 +491,25 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
...
-        plmax = min(mlmax)
-        logger.info(f"The maximum lmax across shells is: {plmax}")
+        ## the max lmax within any 1 shell is used
+        plmax = max(mlmax)
+        logger.info(f"The maximum lmax for any one shell is: {plmax}")
         
     ## if lmax too large, reset with warning
     if sh_order <= plmax:
         logger.info(f"Running model with requested lmax: {sh_order}")
     else:
```
nikhil153, line 508: This block only has logging?

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -489,23 +491,25 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
...
         logger.info(f"The requested lmax ({sh_order}) exceeds the recommended capabilities of the data ({plmax})")
+        logger.info(f"Generally, you do not want to fit an lmax in excess of any one shell's ability in the data.")
         logger.info(f" -- You should redo with sh_order set to: {plmax}")
-        logger.info(f"Running model with overrode lmax: {sh_order}")
+        #logger.info(f"Running model with overrode lmax: {sh_order}")
         #sh_order = plmax
         
     ## hard coded inputs to the tractoflow command in mrproc
```
nikhil153, line 515: Are these `global vars` or `run args`? Would be nice to either move them to the top or include in the argparse?

nikhil153 (commented):
This is great and a lot of work! 
I think it would be useful to have a `README` to document all the pain points and things to be aware of in the current setup for this tractoflow version. For instance the non-BIDS file structure, symlinks, nextflow issues etc?

nikhil153 (commented):
Would it possible to pass a bids_db_file instead of [creating the layout](https://github.com/neurodatascience/nipoppy/pull/166/files#diff-8e74aed6e7d6e1fdaf61127259eea4a211f1de73c29d8dd7e6b580854e250ea4R34) during every run?

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
bcmcpher: I've had a hard time finding the reference, but I can clarify it. 

It identifies the highest lmax possible based on the number of directions (`tdir.shape[1]`).

If you try and fit an lmax higher than is supported, it will be under specified in it's estimates if it runs at all.

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
bcmcpher: Yeah. This is where I would have overrode a bad lmax. But the logic wasn't working with the new data, so I decided it was easier to just print some useful debugging in case it is an issue later on.

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
bcmcpher: `sh_order` is a `run_args` passed by the user that has a default in the run script. We could make the default adjustable in global, but that might be more trouble than it's worth. 

`plmax` has to be determined by the data - it's something the user _can_ figure out beforehand so they can set it correctly, but in my experience it's worth parsing and checking this upfront.

What I'm doing with these steps is verifying the data have enough directions to fit that parameter, otherwise it will fail once it starts running.

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -21,22 +22,59 @@
...
-def parse_data(bids_dir, participant_id, session_id, logger=None):
+def parse_data(global_configs, bids_dir, participant_id, session_id, use_bids_filter=False, logger=None):
     """ Parse and verify the input files to build TractoFlow's simplified input to avoid their custom BIDS filter
     """
 
+    ## load from global configs
+    DATASET_ROOT = global_configs["DATASET_ROOT"]
+
     ## because why parse subject ID the same as bids ID?
     subj = participant_id.replace('sub-', '')
```
nikhil153, line 33: Not urgent, but there are helper functions and global variables now for different ID conversions in [utils.py](https://github.com/neurodatascience/nipoppy/blob/main/nipoppy/workflow/utils.py)

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -21,22 +22,59 @@
...
+    
+    ## parse bids directory with subject filter
+    layout = BIDSLayout(bids_dir, indexer=bidx)
+    ## check if DB exists on disk first? BIDSLayout(database_path=var)? where is this saved?
+    ## should this be made / updated as part of BIDS-ification of dicoms?
+
+    ## load the bids filter if it's called
+    if use_bids_filter:
+
+        bidf_path = Path(f"{DATASET_ROOT}", 'proc', 'bids_filter_tractoflow.json') ## is this where it will always be?
```
nikhil153, line 51: This is where it gets copied from CWD of `run_tractoflow`. I find it a bit annoying to open and edit it outside of the repo and I think it could be helpful to commit these to dataset specific forks.

nikhil153 (approved):

bcmcpher:
Down to minor refactor issues.
merging working version.

PR #177, opened 2023-11-02
Minor fixes to trackers

### LINKED ISSUE #176: Bring back `INCOMPLETE` status in derivatives bagel
So we will have 4 statuses, with the logic being:
* `UNAVAILABLE`: The subject **does not have** the required MRI datatype for this session (based on the manifest)
* If the right MRI datatype(s) are theoretically available for this subject-session pair:
  * `INCOMPLETE`: The expected parent directory for the derivatives **does not exist** -> the pipeline hasn't been run yet
    * The directory to check must be **specific to both the subject AND the session**. For example for fMRIPrep it should be `.../sub-<SUBJECT>/ses-<SESSION>`, and for Freesurfer it should be `.../ses-<SESSION>/sub-<SUBJECT>`
  * If the expected parent directory for the derivatives exists:
    * `SUCCESS`: **all files** specified in the tracker exist
    * `FAIL`: at least one file specified in the tracker **does not exist**

@alyssadai @nikhil153 does that look okay?

nikhil153:
resolves #176 
- [x] fixed tractoflow subject dir path issue
- [x] added INCOMPLETE status based on subject-session-dir path
- [ ] incorporate manifest datatype check


nipoppy/trackers/run_tracker.py
```diff
@@ -161,19 +161,23 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
 
                 if pipeline == "heudiconv":
                     subject_dir = f"{DATASET_ROOT}/bids/{bids_id}"
```
michellewang, line 163: Would it be better to only have `subject_ses_dir` (i.e. no `subject_dir`) and just pass `subject_ses_dir` to the trackers? Then all the trackers would have to be modified but it might be cleaner in the end.

nipoppy/trackers/run_tracker.py
```diff
@@ -161,19 +161,23 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
...
                 else:
                     logger.error(f"unknown pipeline: {pipeline}")
                     
-                dir_status = Path(subject_dir).is_dir()
-                logger.debug(f"subject_dir:{subject_dir}, dir_status: {dir_status}")
+                subject_ses_dir_status = Path(subject_ses_dir).is_dir()
+                logger.debug(f"subject_dir:{subject_ses_dir_status}, dir_status: {subject_ses_dir_status}")
                 
-                if dir_status:                                 
+                # TODO incorporate manifest datatype availability
```
michellewang, line 179: Are we going to use the manifest here *instead* of the doughnut or *in addition* to the doughnut? I feel like it would replace the doughnut since the doughnut tells us who was successfully converted to BIDS, but we'll just use `INCOMPLETE` for people who haven't been successfully converted to BIDS (if I understand correctly). If we don't want to differentiate between "no BIDS data yet" and "with BIDS but not run" then we only need the manifest I think

michellewang: Also, the `datatype` for the manifest only gives `anat`/`func`/`dwi`/etc. level of information, it doesn't say if someone has `anat` but only T2w (which is a case in PPMI...). So it might be a little accurate for these edge cases but I think that should still be fine (?)

michellewang (commented):
Looks good! Just some questions about using the manifest for datatype availability which we might want to clarify before you implement it

nipoppy/trackers/run_tracker.py
nikhil153: I was thinking the same, but we have to pass the `session_id` to pipelines with BIDS naming to construct the expected filename. So not sure if it would help much.

nipoppy/trackers/run_tracker.py
nikhil153: All valid questions which basically made me procrastinate on making these changes until recently :) 

But yes, I think only using `manifest` as expected truth makes sense to me. `doughnuts` to me are more or less are like `bagels` informing disk status of organized raw data. 

Now that we have `bids_tracker`, we can assign `INCOMPLETE` for a `participant x session`, if the manifest specifies a `datatype` for it but no corresponding directory is found on disk. Along with pipeline trackers, this would help pinpoint what "expected" outputs are incomplete in general. 

Does this make sense?

michellewang: Yes I think so!

michellewang (approved):
We decided to merge this then make a new PR for replacing the doughnut by the manifest

PR #178, opened 2023-11-03
Refactor runscripts to use Boutiques

### LINKED ISSUE #59: Harmonize `proc_pipe` runscripts
* Make sure that each individual pipeline's runscript takes a subject ID and a session ID
* Add a driver script that loops over all subject sessions and calls the pipeline (for use in job arrays)

Will wait after TractoFlow code is added to the repo.

michellewang:
Closes #59.

Some notes:
* Not quite done writing tests yet... still need `ProcpipeRunner`, `FmriprepRunner` and `MriqcRunner`
* Planning to add docstrings too...
* Other modules not directly related to the runners:
  * `base.py`: a simple `Base` class with nice string representation + `GlobalConfigs` class that loads the global configs and provides an interface to the underlying JSON dict
  * `parser.py`: `CustomParser` class that can be used for common parser arguments (`--global_configs`, `--subject`, etc.). Will load the global configs into a `GlobalConfigs` and create a logger if logging arguments are added
  * `logger.py`: create logger that streams to stdout and (optionally) a file. Basically the same as `workflow.logger` but made the file optional and in the parent directory (with the idea being that we will use that instead of `workflow.logger`).
  *  `io.py`: convenience functions for loading JSONs. Thinking of eventually adding functions for loading/writing CSVs/TSVs
* I also fixed the old fMRIPrep tests failing
* Do we want to keep the old runscripts for now or should I delete them?
* I'm not sure about this file directory structure yet, open to suggestions


michellewang:
Closing because this is no longer relevant after the refactor in #205.

PR #179, opened 2023-11-03
Quick fix for tractoflow data parsing missing valid T1s

bcmcpher:
- fixed parsing issues for some T1 images - just dropped `sag` filter.
 - updated logging for more verbose reporting / traceback of what files get selected as forward / reverse phase encodings
 -  clarified the language of interpreting `lmax` selection to be more clear


nikhil153 (approved):
I wonder if it would be worth externalizing many of these heuristic checks on filetypes, scan, shell and bval sizes in a dataset-specific config file for provenance sake.

PR #181, opened 2023-11-07
Update trackers to use manifest and handle tar archives

### LINKED ISSUE #97: Fetch bagel schema automatically when running trackers
Right now, the user is expected to provide the [schema file from the dashboard repo](https://github.com/neurobagel/proc_dash/blob/main/schemas/bagel_schema.json) when calling `run_tracker.py`. This can be inconvenient because if we need to re-download the file every time the schema changes.

@alyssadai discussed with Rémi and came up with a possible solution: use `requests` to fetch the contents of the file from GitHub ([StackOverflow example](https://stackoverflow.com/questions/14120502/how-to-download-and-write-a-file-from-github-using-requests)) every time we run the tracker. Of course, this only works if the user has Internet, so we can still let the user specify a path as fallback. Maybe we can save the schema locally after it is fetched so that people have a more-or-less up-to-date version on their computer.

### LINKED ISSUE #174: Handle tar archives in trackers
The upcoming Boutiques PR will add the option to tar result directories for processing pipelines to limit inode usage. The trackers need to be updated to handle tar files.

For now I have updated my local copy of `run_tracker.py` to set `pipeline_complete` to `'SUCCESS'` if the directory is tarred, i.e. no files inside the archive are checked. This works because tarring is only done upon successful processing (if the job failed, then nothing is tarred).

Ideally we would run `tar -tf <PATH>` to get the list of relative paths inside the archive and check those for `pipeline_complete` and each phase/stage as usual.

In the future, we might want to consider adding custom fields to the Boutiques descriptors for knowing which file(s) to track (instead of having individual `fmriprep_tracker.py`, `mriqc_tracker.py`, etc.). The only issue is that it might not be super flexible since complex logic cannot be implemented in JSON. We might only be able to check file suffixes and/or use regexes and/or Nipoppy string replacement (e.g. `[[NIPOPPY_SUBJECT]]`). Things like @bcmcpher's complex input checking would be more difficult to do though.

### LINKED ISSUE #180: Replace doughnut by manifest in `run_tracker.py`
We decided to merge this then make a new PR for replacing the doughnut by the manifest

_Originally posted by @michellewang in https://github.com/neurodatascience/nipoppy/pull/177#pullrequestreview-1713579466_

michellewang:
Closes #180, closes #174, closes #97 (except for the `--check-for-update` flag)

With a minor bug fix that made the script write new bagel files even if they didn't change from the previous version.

Some notes:
- Added more logic to handle the fMRIPrep 20.2.7 results being in `output/fmriprep` instead of just `output`. So @nikhil153 depending on your setup for QPN/NIMHANS we might need to put things back into an `fmriprep` subdirectory
- I have a personal preference for using the version name without a preceding `v`, which is how I have coded the Boutiques thing to do for now. This PR makes the tracker compatible with the current Boutiques runner outputs. So this will break things with older runscripts unless we rename the version directories (e.g. `v20.2.7` -> `20.2.7`). @nikhil153 let me know what you think about that

### DIFF
--- a/nipoppy/trackers/run_tracker.py
+++ b/nipoppy/trackers/run_tracker.py
@@ -243,12 +243,12 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
             _df = _df.reset_index(drop=True)
 
             # add old rows from other pipelines/sessions and sort for consistent order
-            df_bagel: pd.DataFrame = pd.concat([df_bagel_old, _df], axis='index')
+            df_bagel: pd.DataFrame = pd.concat([df_bagel_old, _df], axis='index', ignore_index=True)
             df_bagel = df_bagel.sort_values(["pipeline_name", "pipeline_version", COL_BIDS_ID_MANIFEST, COL_SESSION_MANIFEST], ignore_index=True)
 
             # don't write a new file if no changes
             try:
-                if len(df_bagel.compare(df_bagel_old_full)) == 0:
+                if (df_bagel.shape == df_bagel_old_full.shape) and (set(df_bagel.columns) == set(df_bagel_old_full.columns)) and (len(df_bagel.compare(df_bagel_old_full)) == 0):
                     logger.info(f'No change in bagel file for pipeline {pipeline}, session {session}')
                     continue
             except Exception as exception:
--- a/nipoppy/trackers/run_tracker.py
+++ b/nipoppy/trackers/run_tracker.py
@@ -39,6 +39,7 @@ PIPELINE_REQUIRED_DATATYPES = {
     "mriqc": ["anat"],
     "tractoflow": ["anat", "dwi"],
 }
+ALL_DATATYPES = sorted(["anat", "dwi", "func", "fmap"])
 BIDS_PIPES = ["mriqc","fmriprep"]
 NO_TRACKER_PIPES = ["maget_brain"]
 
@@ -100,6 +101,8 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
         # for prefixed columns we need to generate the column name
         dash_col_list = list(key for key, value in schema["GLOBAL_COLUMNS"].items() if not value["IsPrefixedColumn"])
         # status_check_dict will typically only have minimal pipeline_complete key
+        for datatype in ALL_DATATYPES:
+            dash_col_list.append(f"HAS_DATATYPE__{datatype}")
         dash_col_list = dash_col_list + list(status_check_dict.keys())
 
         for session in sessions:
@@ -194,6 +197,10 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
                 ]
                 subject_ses_tar_status = any([path.exists() for path in subject_ses_tar_paths])
                 logger.debug(f"subject_ses_dir: {subject_ses_dir}, dir_status: {subject_ses_dir_status}, subject_ses_tar_status: {subject_ses_tar_status}")
+
+                # populate HAS_DATATYPE__ columns
+                for datatype in ALL_DATATYPES:
+                    _df.loc[bids_id, f"HAS_DATATYPE__{datatype}"] = datatype in datatypes
                 
                 if subject_ses_tar_status:
                     logger.debug(f"subject_ses_dir: {subject_ses_dir} is a tar file")
--- a/nipoppy/trackers/run_tracker.py
+++ b/nipoppy/trackers/run_tracker.py
@@ -11,19 +11,16 @@ import nipoppy.workflow.logger as my_logger
 from nipoppy.trackers.tracker import Tracker, get_start_time, get_end_time, SUCCESS, UNAVAILABLE, INCOMPLETE, TRUE
 from nipoppy.trackers import bids_tracker, fs_tracker, fmriprep_tracker, mriqc_tracker, tractoflow_tracker
 from nipoppy.workflow.utils import (
-    BIDS_SUBJECT_PREFIX,
     BIDS_SESSION_PREFIX,
     COL_DATATYPE_MANIFEST,
     COL_SUBJECT_MANIFEST,
     COL_BIDS_ID_MANIFEST,
     COL_SESSION_MANIFEST,
-    COL_CONV_STATUS, 
     DNAME_BACKUPS_BAGEL,
     FNAME_BAGEL,
     FNAME_MANIFEST,
     load_manifest,
     save_backup,
-    session_id_to_bids_session,
 )
 
 # Globals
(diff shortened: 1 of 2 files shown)


nikhil153:
Both of these issues are I think user- vs dev- friendliness tradeoff :) 

- From user perspective, I think it's better to standardize root output dir across pipelines. Which is why I was manually moving the data in the legacy `fmriprep20.2.7` which creates confusing `fmriprep/v6.0.1/output/fmriprep` path. If manual moving is annoying (which it is) maybe we should just check and support both output paths?
 
- Why do you not like the `v` prefix? I don't have strong preference here. I was going with what I usually see in the user facing doc pages (see [example](https://fmriprep.org/en/stable/)) to make it intuitive for new users. But it shouldn't be difficult to change it.

nipoppy/trackers/run_tracker.py
```diff
@@ -34,6 +32,13 @@
     "mriqc": mriqc_tracker.tracker_configs,
     "tractoflow": tractoflow_tracker.tracker_configs,
 }
+PIPELINE_REQUIRED_DATATYPES = {
+    "heudiconv": [],
+    "freesurfer": ["anat"],
+    "fmriprep": ["anat"],
+    "mriqc": ["anat"],
+    "tractoflow": ["anat", "dwi"],
+}
```
nikhil153, line 41: Should this go in global_config?

nipoppy/trackers/run_tracker.py
```diff
@@ -154,30 +156,41 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
...
                     subject_ses_dir = f"{subject_dir}/{session}"
-                elif pipeline in ["freesurfer","tractoflow"]:
-                    subject_dir = f"{DATASET_ROOT}/derivatives/{pipeline}/v{version}/output/{session}/{bids_id}" 
+                elif pipeline in ["freesurfer", "tractoflow"]:
+                    subject_dir = f"{DATASET_ROOT}/derivatives/{pipeline}/{version}/output/{session}/{bids_id}" 
                     subject_ses_dir = subject_dir
                 elif pipeline in BIDS_PIPES:
-                    subject_dir = f"{DATASET_ROOT}/derivatives/{pipeline}/v{version}/output/{bids_id}" 
+                    subject_dir = f"{DATASET_ROOT}/derivatives/{pipeline}/{version}/output/{bids_id}" 
+                    if pipeline == "fmriprep" and version == "20.2.7":  # NOTE temporary solution while we refactor tracker configs to be version-specific
```
nikhil153, line 172: I think it would be more robust if we just check both output paths instead of being very specific to a given fmriprep version string.

nipoppy/trackers/run_tracker.py
```diff
@@ -154,30 +156,41 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
...
+                        subject_dir = f"{DATASET_ROOT}/derivatives/{pipeline}/{version}/output/fmriprep/{bids_id}"
                     subject_ses_dir = f"{subject_dir}/{session}"
                 elif pipeline in NO_TRACKER_PIPES:
                     logger.warning(f"pipeline: {pipeline} does not have a tracker yet...")
                 else:
                     logger.error(f"unknown pipeline: {pipeline}")
                     
                 subject_ses_dir_status = Path(subject_ses_dir).is_dir()
-                logger.debug(f"subject_dir:{subject_ses_dir_status}, dir_status: {subject_ses_dir_status}")
+                subject_ses_tar_status = Path(subject_ses_dir).with_suffix('.tar').is_file()
```
nikhil153, line 181: Will always be `tar` or `tar.gz` is also possible?

nikhil153: Also if it doesn't slow down things too much, might be worth adding the `tarfile.is_tarfile()` check.

nipoppy/trackers/run_tracker.py
michellewang: I was thinking this should go in the individual trackers instead? It would be the datatypes required for the `pipeline_complete` part of the pipeline to run

nipoppy/trackers/run_tracker.py
michellewang: Oh I didn't know about `tarfile.is_tarfile()` 😅  but I just tried it and it seems like it is slowing things down a fair bit 😞 (PPMI fMRIPrep: ~6 minutes with `is_tarfile`, ~1 minute with just checking file existence)

michellewang:
@nikhil153 added the additional check for `.tar.gz` file extension

> From user perspective, I think it's better to standardize root output dir across pipelines. Which is why I was manually moving the data in the legacy fmriprep20.2.7 which creates confusing fmriprep/v6.0.1/output/fmriprep path. If manual moving is annoying (which it is) maybe we should just check and support both output paths?

Ok, I changed it to check both output paths for now, but it is very messy (since it affects the path to the subject directory, the path to the session directory, and the path to the potential tarred archive). So I still think we should eventually have a more user-configurable way to get paths to track for different versions.

> Why do you not like the v prefix? I don't have strong preference here. I was going with what I usually see in the user facing doc pages (see [example](https://fmriprep.org/en/stable/)) to make it intuitive for new users. But it shouldn't be difficult to change it.

I think it might be a Pypi thing to have the `v` prefix... I just don't think it's necessary 😕 

Also added the `HAS_DATATYPE__<datatype>` columns since that's easy to do now that we have the manifest. Mentioned by @alyssadai today when talking about the dashboard.

michellewang:
@nikhil153 could you take another look? The Neurobagel team will try to get a someone at the Douglas to run the BIDS tracker soon, and it would be good if they only need the manifest file (not the doughnut) for that

nipoppy/trackers/run_tracker.py
nikhil153: Individual trackers makes sense!

nipoppy/trackers/run_tracker.py
```diff
@@ -192,23 +226,33 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
...
-                    for name in status_check_dict.keys():                    
-                        _df.loc[bids_id,name] = INCOMPLETE
-                        _df.loc[bids_id,"pipeline_starttime"] = UNAVAILABLE
-                        _df.loc[bids_id,"pipeline_endtime"] = UNAVAILABLE
+                    for name in status_check_dict.keys():
+
+                        _df.loc[bids_id, name] = INCOMPLETE
+
+                        # check that all required datatypes are available
+                        for datatype in PIPELINE_REQUIRED_DATATYPES[pipeline]:
```
nikhil153, line 234: This branching feels a bit odd - shouldn't the "availability" of participant be verified before checking for "incomplete" status?

nikhil153:
>Ok, I changed it to check both output paths for now, but it is very messy (since it affects the path to the subject directory, the path to the session directory, and the path to the potential tarred archive). So I still think we should eventually have a more user-configurable way to get paths to track for different versions.

Yes made a comment about that. Other than that it's good to go!

michellewang:
@nikhil153 I changed the branching logic for `UNAVAILABLE` status based on the manifest's `datatypes` column like we discussed this morning

nikhil153 (approved):

PR #185, opened 2023-11-21
use doughnut file to get BIDS ID in `run_tracker.py` (create if needed)

### LINKED ISSUE #184: Fix the bids_id issue from run_tracker
Currently, `run_tracker` [expects](https://github.com/neurodatascience/nipoppy/blob/main/nipoppy/trackers/run_tracker.py#L91) `bids_id` in the manifest after recent refactor that replaced doughnut dependency with manifest. 

We need to revert to doughnut since `bids_id` is not part of manifest and cannot be inferred easily when `participant_id` is poorly named. Since `doughnut` is not automatically available for datasets without dicoms, we will have to add a check and generate one with a relevant warning.

michellewang:
Closes #184.

### DIFF
--- a/nipoppy/trackers/run_tracker.py
+++ b/nipoppy/trackers/run_tracker.py
@@ -10,6 +10,7 @@ import pandas as pd
 import nipoppy.workflow.logger as my_logger
 from nipoppy.trackers.tracker import Tracker, get_start_time, get_end_time, SUCCESS, UNAVAILABLE, INCOMPLETE, TRUE
 from nipoppy.trackers import bids_tracker, fs_tracker, fmriprep_tracker, mriqc_tracker, tractoflow_tracker
+from nipoppy.workflow.make_doughnut import run as run_make_doughnut
 from nipoppy.workflow.utils import (
     BIDS_SESSION_PREFIX,
     COL_DATATYPE_MANIFEST,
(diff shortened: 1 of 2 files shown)


nipoppy/trackers/run_tracker.py
```diff
@@ -48,6 +51,24 @@ def run(global_configs, dash_schema_file, pipelines, session_id="ALL", run_id="1
...
+    df_doughnut = load_doughnut(fpath_doughnut)
+
+    # load the manifest
+    fpath_manifest = f"{DATASET_ROOT}/tabular/{FNAME_MANIFEST}"
+    df_manifest = load_manifest(fpath_manifest)
+
+    # add BIDS ID column to the manifest
+    df_manifest = df_manifest.merge(
+        df_doughnut,
+        on=[COL_SUBJECT_MANIFEST, COL_SESSION_MANIFEST],
```
nikhil153, line 68: might be an issue if there is `bids_id` present in the manifest which would create duplicate names with `x` and `y` prefixes in pandas.

michellewang: Good catch! 😅 In that case should we use the `bids_id` column from the manifest or the doughnut?

nikhil153 (approved):

PR #189, opened 2024-01-23
Add dcm2bids conversion

danyluik:


nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,233 @@
...
+    COL_DICOM_ID,
+    DNAME_BACKUPS_DOUGHNUT,
+    FNAME_DOUGHNUT, 
+    load_doughnut,
+    save_backup,
+    session_id_to_bids_session,
+)
+
+#Author: nikhil153
+#Date: 07-Oct-2022
```
nikhil153, line 29: I didn't know I wrote this!

michellewang:
Thanks Matt! Just FYI: the test failure is unrelated to the changes you made, so don't worry about it

danyluik:
okay :)

nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python
+
+import argparse
+import json
+import glob
+import os
```
michellewang, line 6: Looks like `os` isn't used, can probably remove

nipoppy/workflow/bids_conv/run_dcm2bids.py
michellewang: I think what Nikhil meant but that comment is that you can delete those lines :)

nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python
```
michellewang, line 1: Could you change file permissions to make it executable (`chmod u+x run_dbm2bids.py`)? I don't think that's the case right now

nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,233 @@
...
+        logger.info(f"No new participants found for bids conversion...")
+
+    logger.info("-"*50)
+    logger.info("")
+
+if __name__ == '__main__':
+    # argparse
+    HELPTEXT = """
+    Script to perform DICOM to BIDS conversion using dcm2bids
+    """
```
michellewang, line 200: Could you add somewhere (either here or in the help message for `--stage`) that, if running Stage 2, the config file is expected to be found at `<DATASET_ROOT>/proc/dcm2bids_config.json`?

michellewang (commented):
Thanks a lot @danyluik, this looks pretty good! I only had some minor comments, but let's wait for @nikhil153 to also take a look too before approving/merging. 

Some things that were in the Heudiconv script are probably not needed here (e.g. binding `tabular`, copying files to `proc`) since the config file for `dcm2bids` is a JSON file (for Heudiconv we could have a function that read another file), but I think it's fine to leave them for now.

I guess one thing to maybe discuss with @nikhil153 is the config file living in the dataset directory (`proc`), instead of being in the code repo and being copied to `dataset/proc` at runtime, like we do with Heudiconv. The reason we had that was that we would have the config file with the rest of dataset-specific scripts on GitHub for the datasets we're processing, but in my opinion we probably want to move away from that in the future (and have `nipoppy` as a standalone package).

danyluik:
Thanks for taking a look! I made the requested changes. Will leave the location of the config file up to you guys.

nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,230 @@
...
+        for dname in ['bids', 'tabular', 'proc']:
+            flag_bind += f'--bind {DATASET_ROOT}/{dname}:{SINGULARITY_WD}/{dname} '
+    else:
+        flag_bind = f'--bind {DATASET_ROOT}:{SINGULARITY_WD}'
+
+    # Singularity CMD 
+    SINGULARITY_CMD=f"{SINGULARITY_PATH} run {flag_overlay} {flag_bind} {SINGULARITY_DCM2BIDS} "
+
+    # dcm2bids CMD
+    if stage == 1:
```
nikhil153, line 61: maybe we should name the stages with more informative keyword e.g. "prepare-data", "convert-data" (or something better!)

nipoppy/workflow/bids_conv/run_dcm2bids.py
```diff
@@ -0,0 +1,230 @@
...
+            -o {SINGULARITY_BIDS_DIR} "
+
+    else:
+        logger.error(f"Incorrect dcm2bids stage: {stage}")
+
+    CMD_ARGS = SINGULARITY_CMD + dcm2bids_CMD 
+    CMD = CMD_ARGS.split()
+
+    logger.info(f"CMD:\n{CMD_ARGS}")
+    dcm2bids_proc_success = True
```
nikhil153, line 83: Shouldn't this be inside `try` ?

nikhil153 (approved):
Made minor suggestions.

PR #194, opened 2024-02-09
Fix MRIQC (and fMRIPrep) runscript bugs 

### LINKED ISSUE #193: Bugs discovered during CRIUGM work session
- MRIQC (and probably fMRIPrep) throw obscure "cross-disk symlink" error if `--bids-database-dir` is a path that does not exist
  - Fix: create the directory if it does not exist
- MRIQC Singularity call needs `--cleanenv` for it to work on Compute Canada, otherwise it throws an error that `REQUESTS_CA_BUNDLE` cannot be found
  - Fix: add `--cleanenv` argument

michellewang:
Closes #193.

- If the `bids_db_dir` does not exist, write warning in log and create it
  - There was a note saying that `run_nipoppy.py` created those directories, but since we don't require that users run `run_nipoppy.py`, we should handle (within the runscript) the case where the `bids_db_dir` does not exist
  - @nikhil153 `bids_db_dir` is not an argument in `run_fmriprep.run()`, but it is in `run_mriqc.run()`. It is unclear to me what users should provide if they provide a custom path, and it will break things if that path is not already accessible inside the container. So I suggest doing what we do in `run_fmriprep.run()` and just not have it as an argument that the user sets (i.e., always use `<DATASET_ROOT>/proc/<bids_db_mriqc>`)
- Add `--cleanenv` argument to the Singularity `run` call for MRIQC

### DIFF
--- a/nipoppy/workflow/proc_pipe/fmriprep/run_fmriprep.py
+++ b/nipoppy/workflow/proc_pipe/fmriprep/run_fmriprep.py
@@ -137,7 +137,7 @@ def run(participant_id: str,
     log_dir = f"{DATASET_ROOT}/scratch/logs/"
 
     if logger is None:
-        log_file = f"{log_dir}/fmriprep1.log"
+        log_file = f"{log_dir}/fmriprep.log"
         logger = my_logger.get_logger(log_file)
 
     logger.info("-"*75)
--- a/nipoppy/workflow/proc_pipe/fmriprep/run_fmriprep.py
+++ b/nipoppy/workflow/proc_pipe/fmriprep/run_fmriprep.py
@@ -16,6 +16,7 @@ CWD = os.path.dirname(os.path.abspath(fname))
 SINGULARITY_FS_DIR = "/fsdir/"
 SINGULARITY_TEMPLATEFLOW_DIR = "/templateflow"
 SINGULARITY_FS_LICENSE = "/fsdir/license.txt"
+DNAME_BIDS_DB = 'bids_db_fmriprep'
 os.environ['SINGULARITYENV_SUBJECTS_DIR'] = SINGULARITY_FS_DIR
 os.environ['SINGULARITYENV_FS_LICENSE'] = SINGULARITY_FS_LICENSE
 os.environ['SINGULARITYENV_TEMPLATEFLOW_HOME'] = SINGULARITY_TEMPLATEFLOW_DIR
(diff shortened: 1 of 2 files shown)


nikhil153:
@michellewang - I think the tricky part with `bids_db` is identifying the stale state and creating a fresh instance (i.e. new subjects) if needed. I think in the current proposed change, it will not regenerate the `bids_db` if it's already there. That creates a problem for newly added subjects. This is I delegated this task to `run_nipoppy.py` which identifies new subjects, cleans old db and regenerates a new instances only for the new subjects. Currently, I think mriqc (and fmriprep as well) won't be able to process subset of subjects, so one will have to manually delete the bids_db and then these pipelines will generate a new one for all subjects. 
Let's discuss this and come-up with a better plan for possible workflows for incremental updates in manifest?

michellewang:
@nikhil153 okay, I see what you mean. I think a potential solution is to make a `bids_db` on-the-fly for each processing run (i.e. a mini `bids_db` specific to a single subject and session). I did something like that in my initial Boutiques refactor (see [here](https://github.com/neurodatascience/nipoppy/blob/863a5ba9297dcfc44ccbf98b5c77c6907d416392/nipoppy/workflow/runner.py#L410)) and I think it's a solution that makes sense (it also doesn't take very long to create a mini `bids_db`)

michellewang:
Although, with the way things are right now, merging this PR doesn't affect anything that is run with `run_nipoppy.py`. It just prevents that error from happening if there is no `bids_db` folder. But if users don't use `run_nipoppy.py` then there is no cleanup of the `bids_db` happening which will be an issue if the dataset changes, as you have pointed out. Another solution would be to expose the `--bids-database-wipe` flag, but it looks like only MRIQC has it

nikhil153:
How about we add a `--regenerate_bids_db` flag as we do with the `doughnut`? This would simply delete the current `bids_db` file only to be regenerated by the mriqc / fmriprep for the entire dataset? 
But yes, I like the idea of generating this on the fly for each subject. However I believe this requires unique naming and cleanup when submitted as a hpc job array.

michellewang:
> How about we add a --regenerate_bids_db flag as we do with the doughnut? This would simply delete the current bids_db file only to be regenerated by the mriqc / fmriprep for the entire dataset?

Yes, that should work! I'll add that. 

> But yes, I like the idea of generating this on the fly for each subject. However I believe this requires unique naming and cleanup when submitted as a hpc job array.

Yes, but it should be fairly straightforward, something like:
1. Call the `bids_db_dir` something unique, for example `bids_db-<pipeline>_<subject>_<session>`
2. Set the flag `--bids-database-dir` to be `bids_db-<pipeline>_<subject>_<session>`
3. Generate the BIDS layout (potentially with custom "ignore" regex patterns)
4. Run the pipeline
5. Delete `bids_db-<pipeline>_<subject>_<session>` (even if the pipeline fails)

What do you think @nikhil153? I would only include this in the refactor though, it would be harder to integrate with the current runscripts

nikhil153:
Sounds like a plan! 
Adding subject-specific `bids_db` in the upcoming big refactor makes sense.

nikhil153 (approved):
Looks good!

PR #196, opened 2024-02-12
fixed bugs related to maget-brain data prep and tracking options

nikhil153:


nipoppy/extractors/maget_brain/prepare_data.py
```diff
@@ -50,15 +50,17 @@ def get_masked_image(img_path, mask_path, masked_img_path):
 fmriprep_dir = f"{DATASET_ROOT}/derivatives/fmriprep/{fmriprep_version}/output/"
 maget_dir = f"{DATASET_ROOT}/derivatives/maget_brain/{maget_version}/output/"
 maget_preproc_T1w_nii_dir = f"{maget_dir}/ses-{session_id}/preproc_T1w_nii/"
+maget_proc_list_file = f"{maget_preproc_T1w_nii_dir}proc_participant.csv"
```
michellewang, line 53: This script is writing a CSV file of participants to process? Who have been processed?

nipoppy/trackers/run_tracker.py
```diff
@@ -41,6 +41,7 @@
     "fmriprep": ["anat"],
     "mriqc": ["anat"],
     "tractoflow": ["anat", "dwi"],
+    "maget_brain": ["anat"]
```
michellewang, line 44: So we have a tracker for `maget_brain` but it is not tracked by `git`?

We also don't have a runscript for it... because it's not containerized?

michellewang (commented):
How does `maget_brain` fit with the rest of the processing pipelines? It's not containerized (yet), but do we still want to track the runscripts/trackers in the `git` repo? What are your thoughts on this?

nipoppy/trackers/run_tracker.py
nikhil153: I guess we can ignore this now as we are switching to CLI soon?

nikhil153:
@michellewang - Just need to merge extractor changes

michellewang (approved):
Why did you add back the `v` to the version-level folder? 😢 The runscripts don't use it anymore

michellewang:
Also in the future you could have different branches in your fork so that you can have different PRs that do different things!

nikhil153:
> Also in the future you could have different branches in your fork so that you can have different PRs that do different things!

Yes I keep forgetting to what PRs are for what :( 

>Why did you add back the v to the version-level folder? 😢 The runscripts don't use it anymore

Because of presence of hardlinks, I cannot change that for tractofolow :( - so want to be consistent for QPN pipelines. Otherwise I will need to reprocess entire tractoflow - which would take a while. Also I believe Brent mentioned it's not ideal to have directory names starting with a number, so I thought we dropped the plan.. is that not the case?

michellewang:
> Also I believe Brent mentioned it's not ideal to have directory names starting with a number, so I thought we dropped the plan.. is that not the case?

I don't think having file/directory names start with a number is a big issue -- from a quick online search it seems like something people do, so if it does cause issues it must be pretty niche. Also I should point out that we have plenty of directories starting with numbers in PPMI since the names are just the participant IDs!

nikhil153:
Yeah which is why I didn't mind the change as it's personal preference. But it's creating a mess for QPN / tractoflow data, so then it no longer seems worth to force this minor change... 
On a quick look, even BIDS [recommends](https://bids-specification.readthedocs.io/en/stable/common-principles.html#storage-of-derived-datasets) usage of `v` (although as part of `<pipeline>-<variant>` format) 
Maybe we can make this a user's choice, by making it a specification in the `global_config`?

PR #200, opened 2024-03-18
Remove symlinks from tractoflow output

bcmcpher:
This should clean up completed runs of tractoflow as part of the processing. 

Be sure to do a smaller test in the first batch - it _should_ be working, but tf is a real challenge to test conclusively.

### DIFF
--- a/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
+++ b/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
@@ -6,7 +6,7 @@ import json
 import subprocess
 import tempfile
 import re
-    
+
 import numpy as np
 import nibabel as nib
 from bids import BIDSLayout, BIDSLayoutIndexer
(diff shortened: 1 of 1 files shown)


nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -636,41 +661,64 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
...
+
+            # find every file in the output path (this should grab symlinks)
+            out_files = Path(tractoflow_out_dir).rglob("*")
+
+            # convert symlinks to real files
+            for out in out_files:
+                move_target_to_symlink(out)
+
+            # remove tractoflow_input_dir
+            shutil.rmtree(tractoflow_input_dir)
```
michellewang, line 717: Is this removing the entire `derivatives/tractoflow/<version>/input` directory? Probably not the best thing to do since it might break parallel runs, unless I'm misunderstanding this.

I see that there is also a `tractoflow_subj_dir` -- might be better to remove this one (and its parent) instead.

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
```diff
@@ -636,41 +661,64 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
...
     logger.info(f"Command passed to system:\n\n{CMD}\n")
     logger.info("+"*75)
-    
+
     ## there's probably a better way to try / catch the .run() call here
     try:
         logger.info('Attempting Run')
         tractoflow_proc = subprocess.run(CMD, shell=True)
+
+        # do clean-up
```
michellewang, line 700: I would probably move this entire thing outside the `try`/`except` block

michellewang (commented):
Thanks for the PR! I think the symlink function should work. One concern about removing the entire `input` directory, and a ~minor comment about nesting the cleanup inside the `try`/`catch` block

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
bcmcpher: Good catch, that should have been `tractoflow_nxtf_inp`

They have to be nested deeply to prevent conflicts with tractoflows parsing "help".

nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
bcmcpher: Moved out.

michellewang (approved):
Thanks, looks good now. Will do a test on a single subject once this is merged and let you know if there is any issue

PR #202, opened 2024-03-19
Fix TractoFlow symlink bugs

### LINKED ISSUE #201: Tractoflow symlink cleanup bugs
Found 3 bugs but managed to get it working:
1. Path to `.trk` file was wrong (missing `participant_id` parent directory)
2. Script errored out because of `Readme` files which are not symlinks
3. `nibabel.streamlines.load` only takes file objects or strings, not `pathlib.Path`

@bcmcpher they're really quick fixes but just wanted to let you know, will make a PR soon.

michellewang:
Closes #201 (see issue for descriptions of the 3 minor bugs).

Also changed the Singularity command so that the bound path is `derivatives/tractoflow/<VERSION>` (resolved), instead of just `derivatives` because I am using symlinks pointing to other places on disk where I have more storage.

### DIFF
--- a/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
+++ b/nipoppy/workflow/proc_pipe/tractoflow/run_tractoflow.py
@@ -716,7 +716,8 @@ def run(participant_id, global_configs, session_id, output_dir, use_bids_filter,
 
         # convert symlinks to real files
         for out in out_files:
-            move_target_to_symlink(out)
+            if out.is_symlink():
+                move_target_to_symlink(out)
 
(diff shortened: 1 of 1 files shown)


bcmcpher (approved):
I'd thought the function handled symlinks not being links, but this works too.

michellewang:
@bcmcpher the function does handle it by throwing a `ValueError` but that causes the whole script to crash, which is not what we want because not everything in `output` is a symlink` (e.g. `Readme` files or files that have already been converted to symlinks). So it's better to ignore whatever that is not a symlink instead of throwing an error.

PR #215, opened 2024-05-08
[DOC] make install command more robust

Remi-Gau:
I cannot remember if this is a powershell or just another type of shell issue, but I know that fr some users you need to type:

`pip install -e '.[dev]'`


michellewang (approved):
Thanks @Remi-Gau!

PR #216, opened 2024-05-08
[DOC] fix edit button and unpin doc dependencies

Remi-Gau:


nipoppy_cli/pyproject.toml
```diff
@@ -37,16 +37,16 @@ version = "1.0.0" # TODO eventually use dynamic versioning
...
+    "furo>=2024.1.29",
+    "pygments-csv-lexer>=0.1.3",
+    "sphinx>=7.2.6",
+    "sphinx-argparse>=0.4.0",
+    "sphinx-autoapi>=3.0.0",
+    "sphinx-copybutton>=0.5.2",
+    "sphinx-jsonschema>=1.19.1",
+    "sphinx-togglebutton>=0.3.2",
+    "mdit-py-plugins>=0.4.0",
+    "myst-parser>=2.0.0",
```
Remi-Gau, line 49: not sure if there was a reason to pin those to those versions but I think that >= should be enough, no?

michellewang:
I pinned them because this is [what Read the Docs recommended](https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html) but maybe that was overkill. We can leave as `>=` for now and handle things if they break in the future

nipoppy_cli/docs/source/conf.py
```diff
@@ -31,13 +31,20 @@
...
 
 html_theme = "furo"
 html_static_path = ["_static"]
 
+# -- Furo configuration ------------------------------------------------------
+#  https://pradyunsg.me/furo/customisation/#customisation
+html_theme_options = {
+    "source_repository": "https://github.com/neurodatascience/nipoppy",
+    "source_branch": "main",
+    "source_directory": "docs/",
```
michellewang, line 45: Wrong docs directory
```suggestion
    "source_directory": "nipoppy_cli/docs/",
```
Though I guess we could move it outside of `nipoppy_cli` since I removed the old `docs` directory (which only had a README), but that will require a number of path changes in the docs and the root README

Remi-Gau:
feels like an overkill and also feels strange to be more stringent on the way we handle the dependencies for the build of the doc but not of the library itself.

michellewang (approved):
Just need to fix the path to the `docs` repo. Other than that feel free to merge (same for your other docs PR), thanks!

Remi-Gau:
Also notived some errors and warnings in the doc build: 

Some examples

https://beta.readthedocs.org/projects/nipoppy/builds/24314765/#229666038--141
https://beta.readthedocs.org/projects/nipoppy/builds/24314765/#229666038--201


I do not think this is a problem but it may make it harder when new warnings or errors we may care about get lost in the middle of all those.

michellewang:
Yeah I have seen those too -- not sure why they happen but was planning to take a look eventually 😅

Remi-Gau:
> Yeah I have seen those too -- not sure why they happen but was planning to take a look eventually 😅

no worries: I do not think this is urgent.

Remi-Gau:
OK the edit button still gives a 404:

https://github.com/neurodatascience/nipoppy/edit/main/docs/index.md

and the link should be

https://github.com/neurodatascience/nipoppy/edit/main/nipoppy_cli/docs/index.md

Remi-Gau:
link fixed: merging

PR #220, opened 2024-05-08
[MAINT] use official fids

Remi-Gau:
just released FIDS to pypi so this can now go through


michellewang (approved):
Looks good, thanks @Remi-Gau!

PR #229, opened 2024-05-14
[ENH] Add DICOM header checking and `DicomDirMap` to `DicomReorgWorkflow`

### LINKED ISSUE #225: Update DICOM reorg workflow
The current `DicomReorgWorkflow` does not yet do a few things that the [old script](https://github.com/neurodatascience/nipoppy/blob/821121222384e52e97b24d1c7a3d00983f8af4f7/nipoppy/workflow/dicom_org/run_dicom_org.py) could:
- The DICOM check to see if some DICOM series are derivatives
  - Since it takes a lot of time to check the files and not all datasets have derivatives DICOMs, I think we can make this an "opt-in" thing instead of an "opt-out" thing
- The mapping from participant ID to DICOM dir -- we can let users specify a file (either on the command-line or in the config) for that
- Joblib parallel computing things -- though I would prefer doing this on a bigger scale instead of just for the DICOM reorg workflow

michellewang:
Closes #225

Remaining TODOs
- [x] Add tests for `DicomDirMap`
- [x] Add tests for new validation functions
- [x] Add yet more tests

Summary:
- Add the `pydicom` header check for whether `ImageType` contains `"DERIVED"`
- Add new fields to global config for configuring DICOM reorganization (i.e. in `<DATASET_ROOT>/scratch/raw_dicom`):
  - `DICOM_DIR_MAP_FILE`: custom CSV with `participant_id`, `session` and `participant_dicom_dir` columns
  - `DICOM_DIR_PARTICIPANT_FIRST`: alternative option for choosing between `<DATASET_ROOT>/scratch/raw_dicom/<PARTICIPANT_ID>/<SESSION>` (default) or `<DATASET_ROOT>/scratch/raw_dicom/<SESSION>/<PARTICIPANT_ID>`
  - (These options are mutually exclusive, by the way)
- Some changes to logger to fix testing (previously wasn't able to make the `caplog` fixture work)
- Added some validators for user-provided tabular/config files (i.e. `participant_id` cannot have `sub-` prefix, and `session` must have `ses-` prefix)

Note that this ONLY allows the contents of `raw_dicom` to be somewhat arbitrary. For `sourcedata` I think it makes more sense to enforce `<PARTICIPANT_ID>/<SESSION>`, though that might break some of the existing PPMI/QPN data. But I think it is better for things to already be more BIDS-like after the reorg step.


nipoppy_cli/nipoppy/config/main.py
```diff
@@ -31,6 +59,27 @@ class Config(ModelWithContainerConfig):
...
+            raise ValueError(
+                "Cannot specify both DICOM_DIR_MAP_FILE and DICOM_DIR_PARTICIPANT_FIRST"
+            )
+        # otherwise set the default if needed
+        elif self.DICOM_DIR_PARTICIPANT_FIRST is None:
+            self.DICOM_DIR_PARTICIPANT_FIRST = True
+
+        return self
+
     def _check_no_duplicate_pipeline(self) -> Self:
```
nikhil153, line 83: What about duplicates within `BIDS_PIPES` or `PROC_PIPES`?

nipoppy_cli/nipoppy/tabular/dicom_dir_map.py
```diff
@@ -0,0 +1,138 @@
...
+                    {
+                        cls.col_participant_id: participant,
+                        cls.col_session: session,
+                        cls.col_participant_dicom_dir: participant_dicom_dir,
+                    }
+                )
+            dicom_dir_map = cls(data=data_dicom_dir_map)
+            if validate:
+                dicom_dir_map.validate()
+            return dicom_dir_map
```
nikhil153, line 126: Not exactly sure how the convention works, but isn't this function suppose to return `Self`?

nipoppy_cli/nipoppy/tabular/doughnut.py
```diff
@@ -145,29 +146,25 @@ def get_bidsified_participants_sessions(
 
 def generate_doughnut(
     manifest: Manifest,
+    dicom_dir_map: DicomDirMap,
     dpath_downloaded: Optional[str | Path] = None,
     dpath_organized: Optional[str | Path] = None,
     dpath_bidsified: Optional[str | Path] = None,
```
nikhil153, line 152: Do these need to change to support python3.9?

nikhil153 (approved):
Minor comments related to conventions

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -16,12 +23,33 @@ class Config(ModelWithContainerConfig):
...
+    DICOM_DIR_PARTICIPANT_FIRST: Optional[bool] = Field(
+        default=None,
+        description=(
+            "Whether subdirectories under the raw dicom directory (default: "
+            f"{DEFAULT_LAYOUT_INFO.dpath_raw_dicom}) follow the pattern "
+            "<PARTICIPANT>/<SESSION> (default) or <SESSION>/<PARTICIPANT>. Note: "
+            "this field and and DICOM_DIR_MAP_FILE cannot both be specified"
+        ),
+    )
     BIDS: dict[str, dict[str, dict[str, PipelineConfig]]] = Field(
```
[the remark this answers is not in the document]
michellewang: This has become `BIDS_PIPELINES` in #238

nipoppy_cli/nipoppy/config/main.py
michellewang: Right now that is not possible because BIDS pipelines have the `step` field. But will add that check after we change it (in #238)

nipoppy_cli/nipoppy/tabular/doughnut.py
michellewang: They will be once I get the latest changes from `main`!

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -16,12 +23,33 @@ class Config(ModelWithContainerConfig):
...
+        description=(
+            "Path to a CSV file mapping participant IDs to DICOM directories"
+            ", to be used in the DICOM reorg step. Note: this field and "
+            "DICOM_DIR_PARTICIPANT_FIRST cannot both be specified"
+            f'. The CSV should have three columns: "{DicomDirMap.col_participant_id}"'
+            f' , "{DicomDirMap.col_session}"'
+            f', and "{DicomDirMap.col_participant_dicom_dir}"'
+        ),
+    )
+    DICOM_DIR_PARTICIPANT_FIRST: Optional[bool] = Field(
```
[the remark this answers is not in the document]
michellewang: This is a boolean (not path) for easy switching between non-custom layouts, either participant-first (i.e. `<participant>/<session>`) or session-first (i.e. `<session>/<participant>`). If neither of those work for the dataset, then the user has to provide their custom mapping file

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -40,6 +89,7 @@ def _check_no_duplicate_pipeline(self) -> Self:
                 "Cannot have the same pipeline under BIDS and PROC_PIPELINES"
                 f", got {bids_pipelines} and {proc_pipelines}"
             )
+        return self
 
     def _propagate_container_config(self) -> Self:
         """Propagate the container config to all pipelines."""
```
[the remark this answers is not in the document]
michellewang: It is returning `self`...

nipoppy_cli/nipoppy/tabular/dicom_dir_map.py
michellewang: The `Self` (capitalized) type annotation just means it will return a `DicomDirMap` instance, which is what is happening here :)

PR #237, opened 2024-05-15
[FIX] Make sure Python >=3.9 is supported

### LINKED ISSUE #236: Self import from typing does not work with Python 3.10
```from typing import Self```

Does not work with Python 3.10 as it is only available in Python 3.11+. 

This import is directly stated in the following files:

nipoppy_cli/nipoppy/tabular/base.py
nipoppy_cli/nipoppy/tabular/doughnut.py
nipoppy_cli/nipoppy/config/main.py
nipoppy_cli/nipoppy/tabular/manifest.py

A possible option is to use 

```from typing_extensions import Self```

 instead, as mentioned in this [stack overflow solution](https://stackoverflow.com/a/77247460).

Or mention in the Docs to use Python 3.11+ instead of Python 3.9+.

### PR DESCRIPTION
**Pull request by michellewang**
Closes #236

- Update CI workflow for tests
- Update `pyproject.toml` to 
- Add `from __future__ import annotations` to files that use `|` for `typing.Union` (only available in Python >= 3.10 otherwise)
- Use `from typing_extensions import Self` instead of `from typing import Self` (only available in Python >= 3.11) 
- Combine `str | pathlib.Path` into `StrOrPathLike` typevar

### DIFF
--- a/nipoppy_cli/nipoppy/workflows/base.py
+++ b/nipoppy_cli/nipoppy/workflows/base.py
@@ -1,5 +1,7 @@
 """Workflow utilities."""
 
+from __future__ import annotations
+
 import logging
 import os
 import shlex
@@ -8,7 +10,7 @@ import subprocess
 from abc import ABC, abstractmethod
 from functools import cached_property
 from pathlib import Path
-from typing import Optional, Sequence, Union
+from typing import Optional, Sequence
 
 from nipoppy.base import Base
 from nipoppy.config.main import Config
@@ -65,7 +67,7 @@ class BaseWorkflow(Base, ABC):
 
     def generate_fpath_log(
         self,
-        dnames_parent: Optional[Union[str, list[str]]] = None,
+        dnames_parent: Optional[str | list[str]] = None,
         fname_stem: Optional[str] = None,
     ) -> Path:
         """Generate a log file path."""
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

File: nipoppy_cli/nipoppy/workflows/dicom_reorg.py [hunk 1]
```diff
@@ -13,7 +14,7 @@ class DicomReorgWorkflow(BaseWorkflow):
 
     def __init__(
         self,
-        dpath_root: Path | str,
+        dpath_root: StrOrPathLike,
         copy_files: bool = False,
         fpath_layout: Optional[Path] = None,
```
nikhil153, line 19: in `pipeline.py` this is set as: `fpath_layout: Optional[StrOrPathLike] = None,` 
But several other places it's still `fpath_layout: Optional[Path] = None`

nikhil153 (approved):
Just a couple of minor comments.

nipoppy_cli/nipoppy/workflows/dicom_reorg.py
michellewang: Good catch, fixed!

nipoppy_cli/nipoppy/workflows/base.py
```diff
@@ -101,7 +103,7 @@ def run_command(
 
         Parameters
         ----------
-        command_or_args : Sequence[str] | str
+        command_or_args : Sequence[str]  |  str
```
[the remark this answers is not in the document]
michellewang: This will affect how the autodocs look but I don't think that matters (for now at least)

PR #238, opened 2024-05-16
[ENH] Update global config specification

### LINKED ISSUE #223: Discuss/finalize global config spec
Some things to address:
1. Do we use singular "config" or plural "configs"? At one point it was a mix of the two. I think singular would make more sense but not sure what others think
2. Pipeline config spec
    - Right now we have nested dictionaries for pipelines/versions, which makes it easy to find pipelines and versions that exist, but it would be probably better to have a list of pipeline configs instead, where each config has a `NAME` and `VERSION` field (and optionally `VARIANT`/`STEP`). Would require some checks that no conflicting configs exist, but that shouldn't be too hard to do
4. New `GLOBAL_VARIABLES` field at the top that is processed on load to replace some strings in the config, e.g. for things that should be the same but appear in multiple places like Templateflow directory, Freesurfer license path, etc.

### LINKED ISSUE #251: [BUG] Specifying `DICOM_DIR_MAP_FILE` causes `ValidationError`
An error is only supposed to be raised if **both** `DICOM_DIR_MAP_FILE` and `DICOM_DIR_PARTICIPANT_FIRST` are specified

michellewang:
Closes #223, closes #251

Summary:
1. `PipelineConfig` changed to have an additional `NAME` and `VERSION` field, both of which are required
2. New `BidsPipelineConfig` which is the same as `PipelineConfig` except that it has an additional required field `STEP`
3. New `GLOBALS` field in the config, to address feedback on having to replace the same value multiple times when starting from the sample config file. Also: can basically serve as a top-level/env config things, which we discussed before
4. The `BIDS` field has been renamed to `BIDS_PIPELINES`, to better parallel `PROC_PIPELINES`
5. Update layout to use `global_config.json` (singular) instead of `global_configs.json` (plural)
6. Update `sample_global_config.json` to only have a single version for each processing pipeline


nipoppy_cli/nipoppy/config/main.py
```diff
@@ -22,41 +23,63 @@ class Config(ModelWithContainerConfig):
...
+        _check_pipeline_infos(
+            self.PROC_PIPELINES,
+            pipeline_type="processing",
+            info_func=lambda x: (x.NAME, x.VERSION),
+        )
+        _check_pipeline_infos(
+            self.BIDS_PIPELINES,
+            pipeline_type="BIDS conversion",
+            info_func=lambda x: (x.NAME, x.VERSION, x.STEP),
+        )
```
nikhil153, line 71: Not sure if I understand the purpose here...

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -22,41 +23,63 @@ class Config(ModelWithContainerConfig):
...
+        def _propagate(pipeline_configs: list[PipelineConfig]):
+            for pipeline_config in pipeline_configs:
                 container_config = pipeline_config.get_container_config()
                 if container_config.INHERIT:
                     container_config.merge_args_and_env_vars(self.CONTAINER_CONFIG)
-            else:
-                for (
-                    child_pipeline_or_pipeline_dicts
-                ) in pipeline_or_pipeline_dicts.values():
-                    _propagate(child_pipeline_or_pipeline_dicts)
```
nikhil153, line 81: Not sure how this works either.. currently there are no definitions found for `get_container_config()`

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -122,6 +160,14 @@ def save(self, fpath: str | Path, **kwargs):
             file.write(self.model_dump_json(**kwargs))
 
     @classmethod
-    def load(cls, path: str | Path) -> Self:
+    def load(cls, path: str | Path, apply_globals_replacement=True) -> Self:
```
nikhil153, line 163: Should this change to `StrOrPathLike` type?

nipoppy_cli/nipoppy/config/pipeline.py
```diff
@@ -4,36 +4,75 @@
...
-    DESCRIPTION: Optional[str] = None
-    CONTAINER: Optional[Path] = None
-    URI: Optional[str] = None
-    DESCRIPTOR: Optional[dict] = None
-    DESCRIPTOR_FILE: Optional[Path] = None
-    INVOCATION: Optional[dict] = None
-    INVOCATION_FILE: Optional[Path] = None
-    PYBIDS_IGNORE: list[re.Pattern] = []
-    TRACKER_CONFIG: dict[str, list[str]] = {}
+    """Model for processing pipeline configuration."""
```
nikhil153, line 13: I am hesitant using the term `model` here. It's very overloaded and creates confusion. I presume this could be called `template` or `recipe` ?

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -0,0 +1,161 @@
...
+    "SESSIONS": [
+        "ses-<SESSION_LABEL>"
+    ],
+    "CONTAINER_CONFIG": {
+        "ARGS": [
+            "--cleanenv"
+        ]
+    },
+    "GLOBALS": {
+        "[[HEUDICONV_HEURISTIC_FILE]]": "<PATH_TO_HEURISTIC_FILE>",
```
nikhil153, line 15: Why is it here and also in `BIDS_PIPELINES` :-/

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -0,0 +1,161 @@
...
+                "overwrite": true,
+                "minmeta": true
+            },
+            "CONTAINER_CONFIG": {
+                "ARGS": [
+                    "--bind",
+                    "[[HEUDICONV_HEURISTIC_FILE]]"
+                ]
+            }
+        },
```
nikhil153, line 48: It feels odd to have multiple steps listed as separate pipelines... I understand why you are doing it, but this can't happen with bigger processing pipelines. This also has separate `CONTAINER` AND `CONTAINER_CONFIG` which looks confusing. 
Why not make `STEPS` as nested block that can share `NAME`, `VERSION` etc. but have independent `INVOCATION`

nikhil153 (commented):
First pass before discussion on `global_config.json` schema. 
I will rely on @Remi-Gau look at test code closely :-)

nipoppy_cli/nipoppy/config/main.py
michellewang: This checks that there are no duplicate configs for the same pipeline name/version(/step)

nipoppy_cli/nipoppy/config/main.py
michellewang: That is added in #237 (which might already handle it), but otherwise I can't change it before the other PR is merged

nipoppy_cli/nipoppy/config/pipeline.py
michellewang: Schema maybe?

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
michellewang: This is the template/placeholder string that will be replaced by whatever the user puts on the right side!

nikhil153:
Re: `global_config` substitutions, I was thinking following alternative to avoid loading, regexing, and regenerating json objects. It could be my obsolete way of thinking, but may offer a cleaner way than regex matching. 
The idea would be to use [eval() function](https://realpython.com/python-eval-function/). We put the top level "SUBSTITUTION VARS" inside the `globals` and the variables in the json block holding a given variable inside `locals`. 

`sample_json`
```
{
"SUBS" : {
        "PIPE" : "heudiconv"
    },

    "BIDS": {
        "heudiconv": {
            "VERSION": "0.11.6",    
            "CONTAINER": "f'{pipe}_{VERSION}.sif'",
            "URL": ""
        },
}
```

`sample eval() code`
```
# read global configs
global_config_file = "sample_global_configs.json"
with open(global_config_file, 'r') as f:
    global_configs = json.load(f)

SUBS = global_configs["SUBS"]

bids_configs = global_configs["BIDS"]
heudiconv_configs = bids_configs["heudiconv"]

CONTAINER = heudiconv_configs["CONTAINER"]

# dynamic eval
eval(CONTAINER, SUBS, heudiconv_configs)
```

This may require try...catch, but shouldn't be too difficult. 
Thoughts?

nipoppy_cli/nipoppy/config/pipeline.py
```diff
@@ -2,70 +2,114 @@
...
+        default=None,
+        description=(
+            "Path to the tracker configuration file associated with the pipeline"
+            ". This file must contain a list of tracker configurations"
+            ", each of which must be a dictionary with a NAME field (string)"
+            " and a PATHS field (non-empty list of strings)"
+        ),
+    )
 
     model_config = ConfigDict(extra="forbid")
```
nikhil153, line 40: I think I need to understand this logic better in person!

nipoppy_cli/nipoppy/config/pipeline.py
```diff
@@ -2,70 +2,114 @@
...
     @model_validator(mode="after")
-    def check_fields(self):
+    def validate_after(self):
         """
-        Check that <FIELD> and <FIELD>_FILE fields are not both set.
+        Validate the pipeline configuration after creation.
 
-        Also add an empty invocation if none is provided.
+        Specifically:
+        - If STEPS has more than one item, make sure that each step has a unique name.
```
nikhil153, line 48: Will it be easier to have a generic "duplication" check method in workflow that does this for multiple things e.g. pipelines, versions, steps, trackers etc.. ?

nipoppy_cli/nipoppy/config/pipeline.py
```diff
@@ -2,70 +2,114 @@
...
+                        "Found at least one step with undefined NAME field"
+                        f" for pipeline {self.NAME} {self.VERSION}"
+                        ". Pipeline steps must have names except "
+                        "if there is only one step"
+                    )
+                if step.NAME in step_names:
+                    raise ValueError(
+                        f'Found at least two steps with NAME "{step.NAME}"'
+                        f" for pipeline {self.NAME} {self.VERSION}"
+                        ". Pipeline names must be unique"
```
nikhil153, line 64: Typo - Step names must be unique?

nipoppy_cli/nipoppy/config/pipeline_step.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class PipelineStepConfig(SchemaWithContainerConfig):
+    """Schema for processing pipeline step configuration."""
+
+    NAME: Optional[str] = Field(
+        default=None,
+        description="Step name, required if the pipeline has multiple steps",
+    )
+    DESCRIPTOR_FILE: Optional[Path] = Field(
```
nikhil153, line 20: I thought descriptor files were specific to pipeline (i.e. container) and not steps?

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -0,0 +1,144 @@
...
+        "<VISIT_LABEL>"
+    ],
+    "SESSIONS": [
+        "ses-<SESSION_LABEL>"
+    ],
+    "CONTAINER_CONFIG": {
+        "ARGS": [
+            "--cleanenv"
+        ]
+    },
```
nikhil153, line 13: Is there a reason / time when `--cleanenv` is not needed when using containers? If having this flag will never hurt, then might as well hardcode it pipeline class and save space here!

nipoppy_cli/nipoppy/workflows/pipeline.py
```diff
@@ -250,7 +255,7 @@ def process_template_json(
 
         return template_json_str if return_str else json.loads(template_json_str)
 
-    def get_boutiques_config(self, participant: str, session: str):
+    def get_boutiques_config(self):
```
nikhil153, line 258: Now that you have got me interested in `cached_properties` I am wondering why is this not one ...

nikhil153 (commented):
Overall looks good and most things now make sense to me :-) 
Will discuss a few comments in person and then should be ready to merge!

nipoppy_cli/nipoppy/config/pipeline_step.py
michellewang: We clarified that there should be one descriptor per argparse parser (or equivalent, e.g. usage message). This will also be needed for e.g. Micapipe

nikhil153 (approved):
Made a few minor comments. A lot of important decisions here! It took a while, but glad we spent time understanding the rationale. Hopefully no more global changes! :-) :-)

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -89,16 +96,17 @@
...
             "NAME": "fmriprep",
             "VERSION": "20.2.7",
             "CONTAINER_INFO": {
-                "PATH": "[[NIPOPPY_DPATH_CONTAINERS]]/fmriprep_20.2.7.sif",
-                "URI": "docker://nipreps/fmriprep:20.2.7"
+                "FILE": "[[NIPOPPY_DPATH_CONTAINERS]]/[[NIPOPPY_PIPELINE_NAME]]_[[NIPOPPY_PIPELINE_VERSION]].sif",
+                "URI": "docker://nipreps/[[NIPOPPY_PIPELINE_NAME]]:[[NIPOPPY_PIPELINE_VERSION]]"
             },
             "CONTAINER_CONFIG": {
                 "ENV_VARS": {
```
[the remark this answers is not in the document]
michellewang: So by default we have this Templateflow directory for fMRIPrep. For Freesurfer license it doesn't strictly have to be an env var since fMRIPrep has the `--fs-license-file` argument. It doesn't really make a difference (especially with the replacement things we are doing) so if you prefer having Freesurfer license file also as an env var instead of directly passed to fMRIPrep we can do that

nipoppy_cli/nipoppy/logger.py
```diff
@@ -56,3 +58,26 @@ def add_logfile(logger: logging.Logger, fpath_log: StrOrPathLike) -> None:
     logger.addHandler(file_handler)
     logger.info(f"Writing the log to {fpath_log}")
     return logger
+
+
+def capture_warnings(logger: logging.Logger) -> logging.Logger:
```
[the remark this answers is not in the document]
michellewang: If you are asking why we are capturing warnings at all, it's so that they seen by the logger and written wherever the log is written (otherwise it would appear in the console but not in log files, for example).

If you are asking why this is not in the same function as e.g. `get_logger`, it's because I couldn't get it to work otherwise 😅

nipoppy_cli/nipoppy/workflows/base.py
```diff
@@ -35,6 +35,10 @@ class BaseWorkflow(Base, ABC):
     log_prefix_run_stderr = "[RUN STDERR]"
     validate_layout = True
 
+    # hack to avoid errors when loading/processing the default config
```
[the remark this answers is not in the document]
michellewang: For workflow objects that don't have `pipeline_name` and/or `pipeline_version` init arguments (`DicomReorgWorkflow` and `DoughnutWorkflow`), loading the config will fail because it will try to do the string replacement in the pipeline config, but there is no pipeline name/version to replace them.

Eventually if we come up with a better way to do this replacement thing, it might be worth implementing so that not all pipeline name/version placeholders in the config get replaced by the same thing...

nipoppy_cli/nipoppy/workflows/dataset_init.py
```diff
@@ -54,6 +55,14 @@ def run_main(self):
 
         self.logger.info(f"Created an empty dataset at {self.dpath_root}")
 
+        # copy descriptor files
```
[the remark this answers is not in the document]
michellewang: Not 100% sure what you're asking. This is copying the default descriptor files (that are in the Nipoppy codebase) to the user's dataset. If they don't change the `DESCRIPTOR_FILE` entries in the config then they will point to these copied files. It's the same things for the invocations and tracker config files.

nipoppy_cli/tests/data/config1.json
```diff
@@ -19,7 +19,7 @@
             "NAME": "heudiconv",
             "VERSION": "0.11.6",
             "CONTAINER_INFO": {
-                "PATH": "heudiconv_0.11.6.sif",
+                "FILE": "heudiconv_0.11.6.sif",
```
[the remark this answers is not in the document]
michellewang: It really doesn't matter because the test is not testing the substitutions :)

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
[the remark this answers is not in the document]
nikhil153: No it's fine. I was just checking if there was a way to move this somewhere else if it's going to be empty by default, or offer some useful example for people seeing it for the first time.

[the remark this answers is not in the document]
michellewang: It's not empty though? For fMRIPrep we have the `TEMPLATEFLOW_HOME` env var. Maybe this is just being cut off by GitHub because those lines haven't been changed in this PR

nipoppy_cli/nipoppy/workflows/base.py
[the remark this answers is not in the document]
nikhil153: I see - always messy to deal with mix of standard and non-standard workflows :-/

nipoppy_cli/nipoppy/workflows/dataset_init.py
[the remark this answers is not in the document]
nikhil153: Ahh i see, this makes sense.

nipoppy_cli/nipoppy/data/examples/sample_global_config-all_pipelines.json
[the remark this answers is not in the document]
nikhil153: Yes, you are right. It cuts off exactly the `TEMPLATEFLOW_HOME` env var. The lines around it were still there despite not being changed in the PR.. so got confused. Should be good!

PR #243, opened 2024-05-22
[DOC] Add user guide

michellewang:
- Closes #228
- Closes #301

Preview: https://nipoppy--243.org.readthedocs.build/en/243/

TODO
- [x] Add `"DICOM_DIR_PARTICIPANT_FIRST": true` to default global config
- [x] Add something in ~Quickstart guide~ (update: added to the BIDS conversion/processing pages) explaining the substitutions

PR #263, opened 2024-06-11
[ENH] Rename tabular file columns to `visit_id` and `session_id`

### LINKED ISSUE #241: Rename `<DATASET_ROOT>/scratch/raw_dicom` -> `<DATASET_ROOT>/scratch/raw_imaging`
Since there is the possibility of starting with raw NIfTIs (that still need to be organized, then BIDSified) instead of raw DICOMs

### LINKED ISSUE #242: Rename doughnut columns
Currently the columns are `downloaded`, `organized`, and `bidsified`, which does not make it really clear where we expect the data to be.

Tentative new naming scheme:
- `downloaded` -> `in_raw_dicom` (maybe soon-to-be `in_raw_imaging`): raw data dump of whatever the user is starting with for their analysis (potentially arbitrary structure)
- `organized` -> `in_sourcedata`: still non-BIDS data, but organized following `<participant_id>/<session>` structure
- `bidsified` -> `in_bids`: after BIDS conversion

### LINKED ISSUE #245: Rename manifest columns and drop BIDS prefixes
In the manifest (and by extension the doughnut and other tabular files):
- `participant_id`: without any prefix, but otherwise BIDS-compliant (no illegal characters), e.g., `01`
- `visit_id`: free for all
- `session_id`: without any prefix, e.g., `BL`

Studies with weird participant IDs (e.g. that contain identifying information like actual first/last names, or with characters not allowed in BIDS) should have a `recruitment.csv` file (inside `tabular`?) with a `recruitment_id` field. This file is optional.
- To think about: combine `recruitment.csv` with the `dicom_dir_map` file in #229

In directory names, for imaging data (in particular `sourcedata`): use the prefixes, to be closer to BIDS.

In the CLI, set the flags as `--participant-id` etc. (kebab style, not snake) but strip the prefix if it is provided. Note that argparse does argument prefix matching by default (see video [here](https://www.youtube.com/watch?v=jrs0l3HJ-Ow)) so just typing `--participant` should work

michellewang:
Closes #245, closes #242, closes #241.

Summary:
- New column names in manifest, doughnut, dicom directory map
  - Specifically: `session` -> `session_id` and `visit` -> `visit_id`, where `session_id` **does not** have the BIDS prefix (breaking change)
  - Sample manifest with new columns:
     Screenshot 2024-06-12 at 6 15 19 PM
- Global config fields also updated accordingly `SESSIONS` -> `SESSION_IDS`, `VISITS` -> `VISIT_IDS`
- CLI arguments updated too
- Big variable renaming `participant` -> `participant_id`, `session` -> `session_id`, `visit` -> `visit_id`
   - I might have missed some instances but I think I got most of them
- Doughnut columns renaming to be more explicit about where we expect the data to be:
    - `downloaded` -> `in_raw_imaging`
    - `organized` -> `in_sourcedata`
    - `bidsified` -> `in_bids`
- Renaming of `scratch/raw_dicom` -> `scratch/raw_imaging` as previously discussed

Should be merged before #243 which I will update to reflect the changes here


nipoppy_cli/nipoppy/tabular/bagel.py
```diff
@@ -56,19 +68,26 @@ def check_status(cls, value: str):
...
 
 class Bagel(BaseTabular):
     """A file to track data availability/processing status."""
 
     # column names
     col_participant_id = "participant_id"
-    col_bids_id = "bids_id"
-    col_session = "session"
+    col_bids_participant = "bids_participant"
+    col_session_id = "session_id"
```
nikhil153, line 90: this if probably minor and doesn't affect anything, but in some other files you also name this as `bids_session_id`?

nipoppy_cli/nipoppy/utils.py
```diff
@@ -36,87 +36,97 @@
...
 FIELD_DESCRIPTION_MAP = {
-    "bids_id": "BIDS-compliant participant identifier (e.g., sub-01)",
-    "participant_id": "Participant identifier",
-    "session": "BIDS-compliant identifier imaging session (e.g., ses-1)",
-    "visit": "Visit identifier",
+    "participant_id": "Participant identifier, without the BIDS prefix",
+    "session_id": "Imaging session identifier, without the BIDS prefix",
+    "bids_participant": "Participant identifier with BIDS prefix (e.g., sub-01)",
+    # only for bagel
+    "bids_session": "Participant identifier with BIDS prefix (e.g., sub-01)",
```
nikhil153, line 43: we define `bids_session` only for bagel but not for other places? Elsewhere we are using `session` ?
Also a typo in the example.

nipoppy_cli/nipoppy/utils.py
```diff
@@ -167,18 +177,18 @@ def get_pipeline_tag(
     pipeline_name: str,
     pipeline_version: str,
     pipeline_step: Optional[str] = None,
-    participant: Optional[str] = None,
-    session: Optional[str] = None,
+    participant_id: Optional[str] = None,
+    session_id: Optional[str] = None,
     sep="-",
 ):
     """Generate a tag for a pipeline."""
```
nikhil153, line 184: Tag?

nipoppy_cli/tests/data/dicom_dir_map2.csv
```diff
@@ -1,4 +1,4 @@
-participant_id,session,participant_dicom_dir
-01,ses-BL,all_BL_files_are_together
-01,ses-M12,all_M12_files_are_together
-02,ses-BL,all_BL_files_are_together
+participant_id,session_id,participant_dicom_dir
+01,BL,all_BL_files_are_together
```
nikhil153, line 2: why is this a valid map?

nipoppy_cli/tests/test_tabular_bagel.py
```diff
@@ -11,15 +11,15 @@
     [
         {
             Bagel.col_participant_id: "1",
-            Bagel.col_bids_id: "sub-1",
-            Bagel.col_session: "ses-01",
+            Bagel.col_bids_participant: "sub-1",
+            Bagel.col_session_id: "01",
```
nikhil153, line 15: Doesn't really matter, but until now you were using `session_id` without zero padding.

nikhil153 (approved):
Only minor comments, should be good to merge!

nipoppy_cli/nipoppy/config/main.py
```diff
@@ -137,21 +129,18 @@ def _propagate(pipeline_configs: list[PipelineConfig]):
     @classmethod
     def check_input(cls, data: Any):
         """Validate the raw input."""
-        key_sessions = "SESSIONS"
-        key_visits = "VISITS"
+        key_sessions = "SESSION_IDS"
+        key_visits = "VISIT_IDS"
         if isinstance(data, dict):
-            # if sessions are not given, infer from visits
+            # if sessions are not given, set to be the same as visits
```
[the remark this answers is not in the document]
michellewang: Yes, I think this was part of things I missed during renaming operation

nipoppy_cli/nipoppy/tabular/bagel.py
michellewang: I don't see `bids_session_id` anywhere, but I do use `bids_session` (instead of `session`) now in other parts of the code. The bagel file is a bit trickier because we still want things to be compatible with the dashboard, so cannot rename all columns immediately.

nipoppy_cli/nipoppy/utils.py
michellewang: Like a string that fully identifies the pipeline run (name + version + step + participant + session). Couldn't think of a better name

nipoppy_cli/nipoppy/utils.py
michellewang: Removed the comment as it was not entirely accurate -- initially I kept the entry as `session` but I think I changed it in the `FIELD_DESCRIPTION_MAP` (so now it doesn't match the bagel column, but will make things easier when we eventually drop `session` from the bagel)

nipoppy_cli/tests/data/dicom_dir_map2.csv
michellewang: Because the test doesn't care about the actual paths, just about general validity of the dataframe :)

nipoppy_cli/tests/test_tabular_bagel.py
michellewang: Yeah I don't think we have to enforce a specific style for the tests. More variety is usually good I think

Issue #258, opened 2024-06-05
Have one tracker config file per pipeline step (instead of one per pipeline)?


michellewang:
Mentioned in discussion with @nikhil153 yesterday. This is doable but would require adding a `pipeline_step` column to the bagel file. Though this could potentially cause issues/confusion since the step name is not mandatory if there is only one step (default name is `None`), and changing step names would require regenerating the bagel.

PR #264, opened 2024-06-12
Implement RABIES (DRAFT)

david-roper:
The following pull request is a draft to provide a centralized point for discussion and tracking of the implementation of the RABIES pipeline into nipoppy. 

More information about the RABIES fmri pipeline is provided at the following sites:

Github (https://github.com/CoBrALab/RABIES)
RABIES documentation (https://rabies.readthedocs.io/en/stable/)


nipoppy_cli/nipoppy/data/descriptors/RABIES-0.5.1-confound-correction.json
```diff
@@ -0,0 +1,252 @@
...
+            "description": "Select the file format for figures generated by RABIES.",
+            "optional": true,
+            "type": "String",
+            "command-line-flag": "--figure_format",
+            "value-key": "[FIGURE_FORMAT]"
+        },
+        {
+            "id": "verbose",
+            "name": "verbose",
+            "description": "Select the file format for figures generated by RABIES.",
```
Gab-D-G, line 206: description is wrong

nipoppy_cli/nipoppy/data/descriptors/RABIES-0.5.1-analysis.json
```diff
@@ -0,0 +1,266 @@
+{
+    "name": "RABIES",
+    "description": "tool description",
+    "tool-version": "v0.5.1",
+    "schema-version": "0.5",
+    "command-line": "[[NIPOPPY_CONTAINER_COMMAND]] [[NIPOPPY_FPATH_CONTAINER]] [INCLUSION_IDS] [EXCLUSION_IDS] [PLUGIN] [LOCAL_THREADS] [SCALE_MIN_MEMORY] [MIN_PROC] [FIGURE_FORMAT] [VERBOSE] [FORCE] analysis [SCAN_LIST] [PRIOR_MAPS] [PRIOR_BOLD_IDX] [PRIOR_CONFOUND_IDX] [DATA_DIAGNOSIS] [OUTLIER_THRESHOLD]  [SEED_LIST] [SEED_PRIOR_LIST] [SEED_LIST] [FC_MATRIX] [ROI_TYPE] [FC_MATRIX] [ROI_CSV] [GROUP_ICA] [DR_ICA] [NPR_TEMPORAL_COMP] [NPR_SPATIAL_COMP] [NETWORK_WEIGHTING] [INPUT_DIR] [OUTPUT_DIR]",
+    "inputs": [
+        {
+            "id": "scan_list",
```
Gab-D-G, line 9: that's no longer an option in 0.5.1 I believe

Gab-D-G:
was this ever tested on a sample dataset?

david-roper:
@Gab-D-G This implementation was tested with the sample dataset provided in the [rabies tutorial ](https://github.com/CoBrALab/RABIES_tutorial), specifically the one available on [zenodo](https://zenodo.org/records/8349029)

Gab-D-G:
> @Gab-D-G This implementation was ran with the sample dataset provided in the [rabies tutorial ](https://github.com/CoBrALab/RABIES_tutorial), specifically the one available on [zenodo](https://zenodo.org/records/8349029)

Great! If all ran without errors, then it should be all good.

nipoppy_cli/nipoppy/data/examples/sample_invocations/RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,7 @@
+{
+    "plugin": "MultiProc",
+    "local_threads": 4,
+    "anatomical_resampling": "0.3x0.3x0.3",
```
Gab-D-G, line 4: Is this supposed to be the defaults? This is to speed test the pipeline, but in real application we would lost loads of data. Do not provide any additional defaults for preprocessing.

Gab-D-G (changes requested):
Just make sure that the RABIES default are not overridden. If its not the case, then all is good on my end.

nipoppy_cli/nipoppy/data/examples/sample_invocations/RABIES-0.5.1-confound-correction.json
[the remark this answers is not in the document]
david-roper: This was simply for the test case and for testing reasons. Since this pipeline will be used for real applications I will get rid of these and switch it back to the default values.

Gab-D-G (approved):

Gab-D-G (approved):

david-roper:
@michellewang @Gab-D-G I was wondering how I should go about creating the pipeline tracker configuration file for RABIES and if it is necessary to implement? The final output of the pipeline can have 60+ files, which could lead to quite a large tracker config file if they were all to be tracked. In addition, some files will only exist if a certain argument is passed (ex: the --data_diagnosis tag). Any advice how to tackle this would be greatly appreciated, thanks!

Gab-D-G:
you are right @david-roper that RABIES outputs lots of files, and the set of file varies according to the input parameters. It would be preferable that it is not necessary to track every file manually, since nipype is doing so internal when running RABIES... Not sure I can advise further on this end.

michellewang:
Hi @david-roper, the tracker configuration file is not required, so it's fine if there isn't one for RABIES. Though just to clarify, the tracker doesn't need to include an exhaustive list of all expected output files. If there is a set of files (or a single file) that is generally important for downstream analysis (and are produced no matter the input arguments), then you could include it in the tracker config file, though again that is not required.

Another thing: trackers currently are specified by pipeline, not pipeline step (we might change that, see #258). So if the steps in RABIES are sequential then it might make sense to track only the expected outputs of the last step.

FYI we're currently thinking about how to manage the growing library of supported pipelines (see #316), as it doesn't necessarily make sense to have the JSON descriptors/invocations all in the main code repo. I think we might want to wait until we settle on long-term solution for this before merging this PR.

In the meantime if you have any feedback about the [docs](https://nipoppy.readthedocs.io/en/latest/) or anything else, please let us know!

david-roper:
Ok great thank you both for the advice. I will hold merging the descriptor and invocation files for now until a long term solution (Zenodo upload, etc.) is set (#316). I will also keep the tracker config empty for now but will do a deeper dive into the important files from the final output of RABIES (i.e. the analysis step) that would be beneficial to track. I found the current docs quite helpful and if I have any extra suggestions for them I will let you know!

michellewang:
@david-roper sorry for the long delay on this PR! If this is ready for review I will take a look next week. As you have probably noticed, there have been some changes to where we put the pipeline config files and possibly to the global config file since you last worked on this -- I'll let you know if there's anything else that should be changed. Thanks a lot for your patience!

michellewang:
By the way, the CI checks have found some [possible typos](https://results.pre-commit.ci/run/github/466274304/1741383413.CB4BnB1CQ8yOviRLbK5JbQ), you might want to take a look:
```
nipoppy_cli/nipoppy/data/descriptors/RABIES-0.5.1-confound-correction.json:91: begining ==> beginning
nipoppy_cli/nipoppy/data/descriptors/RABIES-0.5.1-confound-correction.json:131: targetted ==> targeted
nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-analysis.json:11: imbedded ==> embedded
nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-confound-correction.json:91: begining ==> beginning
nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-confound-correction.json:131: targetted ==> targeted
nipoppy_cli/nipoppy/data/descriptors/RABIES-0.5.1-analysis.json:11: imbedded ==> embedded
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-confound-correction.json
```diff
@@ -0,0 +1,366 @@
...
+            "command-line-flag": "--figure_format",
+            "value-key": "[VERBOSE]",
+            "default-value": "1"
+        },
+        {
+            "id": "force",
+            "name": "force",
+            "description": "The pipeline will not stop if previous outputs are encountered. Previous outputs will be overwritten.",
+            "optional": true,
+            "type": "String",
```
michellewang, line 253:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[BIDS_FILTER]",
+            "command-line-flag": "--bids_filter",
+            "default-value": "{\"func\": {\"suffix\": [\"bold\", \"cbv\"]}, \"anat\": {\"suffix\": [\"T1w\", \"T2w\"]}}"
+        },
+        {
+            "id": "bold_only",
+            "name": "bold_only",
+            "description": "Apply preprocessing with only EPI scans. Commonspace registration is executed directly using the corrected EPI 3D reference images. The commonspace registration simultaneously applies distortion correction, this option will produce only commonspace outputs.",
+            "optional": true,
+            "type": "String",
```
michellewang, line 23:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[BOLD_ONLY]",
+            "command-line-flag": "--bold_only",
+           "default-value": "False"
+        },
+        {
+            "id": "anat_autobox",
+            "name": "anat_autobox",
+            "description": "Crops out extra space around the brain on the structural image using AFNI's 3dAutobox",
+            "optional": true,
+            "type": "String",
```
michellewang, line 33:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[ANAT_AUTOBOX]",
+            "command-line-flag": "--anat_autobox",
+           "default-value": "False"
+        },
+        {
+            "id": "bold_autobox",
+            "name": "bold_autobox",
+            "description": "Crops out extra space around the brain on the structural image using AFNI's 3dAutobox",
+            "optional": true,
+            "type": "String",
```
michellewang, line 43:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[OBLIQUE2CARD]",
+            "command-line-flag": "--oblique2card",
+            "default-value": "none"
+        },
+        {
+            "id": "apply_despiking",
+            "name": "apply_despiking",
+            "description": "Applies AFNI's 3dDespike",
+            "optional": true,
+            "type": "String",
```
michellewang, line 63:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[HMC_OPTION]",
+            "command-line-flag": "--HMC_option",
+            "default-value": "optim"
+        },
+        {
+            "id": "isotropic_HMC",
+            "name": "isotropic_HMC",
+            "description": "Whether to resample the EPI to isotropic resolution (taking the size of the axis with highest resolution) for the estimation of motion parameters.",
+            "optional": true,
+            "type": "String",
```
michellewang, line 83:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[ISOTROPIC_HMC]",
+            "command-line-flag": "--isotropic_HMC",
+           "default-value": "False"
+        },
+        {
+            "id": "voxelwise_motion",
+            "name": "voxelwise_motion",
+            "description": "Whether to output estimates of absolute displacement and framewise displacement at each voxel. This will generate 4D nifti files representing motion timeseries derived from the 6 motion parameters. This is handled by antsMotionCorrStats",
+            "optional": true,
+            "type": "String",
```
michellewang, line 93:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[VOXELWISE_MOTION]",
+            "command-line-flag": "--voxelwise_motion",
+           "default-value": "False"
+        },
+        {
+            "id": "apply_slice_mc",
+            "name": "apply_slice_mc",
+            "description": "Whether to apply a slice-specific motion correction after initial volumetric HMC. This can correct for interslice misalignment resulting from within-TR motion. With this option, motion corrections and the subsequent resampling from registration are applied sequentially since the 2D slice registrations cannot be concatenate with 3D transforms. ",
+            "optional": true,
+            "type": "String",
```
michellewang, line 103:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[APPLY_SLICE_MC]",
+            "command-line-flag": "--apply_slice_mc",
+            "default-value": "False"
+        },
+        {
+            "id": "detect_dummy",
+            "name": "detect_dummy",
+            "description": "Detect and remove initial dummy volumes from the EPI.",
+            "type": "String",
```
michellewang, line 112:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[ANATOMICAL_RESAMPLING]",
+            "command-line-flag": "--anatomical_resampling",
+            "default-value": "inputs_defined"
+        },
+        {
+            "id": "apply_STC",
+            "name": "apply_STC",
+            "description": "Select this option to apply the STC step.",
+            "optional": true,
+            "type": "String",
```
michellewang, line 223:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "default-value": "/home/rabies/.local/share/rabies/DSURQE_40micron_labels.nii.gz"
+        },
+        {
+            "id": "bids_dir",
+            "name": "bids_dir",
+            "description": "The root folder of a BIDS valid dataset (sub-XXXXX folders should be found at the top level in this folder).",
+            "optional": false,
+            "type": "String",
+            "value-key": "[BIDS_DIR]",
+            "default-value": "[[NIPOPPY_DPATH_BIDS]]"
```
michellewang, line 325: Should put this in invocation instead
```suggestion
            "value-key": "[BIDS_DIR]"
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+        },
+        {
+            "id": "output_dir",
+            "name": "output_dir",
+            "description": "The directory where the output files should be stored. If you are running group level analysis this folder should be prepopulated with the results of the participant level analysis.",
+            "optional": false,
+            "type": "String",
+            "value-key": "[OUTPUT_DIR]",
+            "default-value": "[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]"
```
michellewang, line 334: Should put this in invocation instead
```suggestion
            "value-key": "[OUTPUT_DIR]"
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[FIGURE_FORMAT]",
+            "default-value": "png"
+        },
+        {
+            "id": "verbose",
+            "name": "verbose",
+            "description": "Select the file format for figures generated by RABIES.",
+            "optional": true,
+            "type": "Number",
+            "command-line-flag": "--figure_format",
```
michellewang, line 414: Wrong values for `verbose`

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,587 @@
...
+            "value-key": "[VERBOSE]",
+            "default-value": "1"
+        },
+        {
+            "id": "force",
+            "name": "force",
+            "description": "The pipeline will not stop if previous outputs are encountered. Previous outputs will be overwritten.",
+            "optional": true,
+            "type": "String",
```
michellewang, line 423:
```suggestion
            "type": "Flag",
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/invocation-RABIES-0.5.1-preprocess.json
```diff
@@ -0,0 +1,6 @@
+{
+    "plugin": "Linear",
```
michellewang, line 2:
```suggestion
    "bids_dir": "[[NIPOPPY_DPATH_BIDS]]",
    "output_dir": "[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]",
    "plugin": "Linear",
```

michellewang (changes requested):
Hi @david-roper, here are some comments for the PR:

1. `nipoppy_cli` has been renamed to just `nipoppy`, so all the files in `nipoppy_cli` should be moved to a new location. I think you have already done that for the descriptors and sample invocation files, but the tracker config file should be moved to `nipoppy/data/examples/sample_pipelines/rabies-0.5.1/tracker_config.json`
2. The RABIES 0.4.8 folder seems to only contain descriptor files (no invocations), and the `tool-version` field says 0.5.1 which is probably wrong. Also there is a `RABIES-0.4.8.json` which has `tool-version` v0.1.0 and subcommand `analysis`, not sure if this was included by mistake
3. The sample global config files have been moved to `nipoppy/data/examples/sample_global_config-{all,latest}_pipelines.json`, so you should copy-paste the RABIES configs to the new files
4. Could you add the RABIES pipeline(s) to the test file `tests/test_default_config.py` with the other pipelines? This will check the pipeline config and descriptor/invocation files in the CI https://github.com/nipoppy/nipoppy/blob/a35ae3a5808042f32a87597b0e9dc56a1dba81c3/tests/test_default_config.py#L196-L211
5. I also left some comments for the 0.5.1 descriptor files -- they may or may not be valid, so feel free to ignore if I am mistaken.

In general, how deeply have you tested the descriptor files? You can run `bosh exec simulate -i <INVOCATION.JSON> <DESCRIPTOR.JSON>` to see the command that is generated by Boutiques (without running it).

Let me know if you have any questions/comments!

david-roper:
@michellewang Thank you for the update! I will get to updating this pull request with you comments and suggestions.

david-roper:
Also checking in with the code the most up to date rabies container is the 0.5.1 version. I can safely remove all 0.4.8 related files.

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/invocation-analysis.json
```diff
@@ -0,0 +1,8 @@
+{
+    "plugin": "Linear",
+    "local_threads": 24,
+    "data_diagnosis": false,
+    "force": true,
+    "input_dir": "[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]",
+    "output_dir": "[[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]"
```
michellewang, line 7: There is an extra opening square bracket here that should be removed:
```suggestion
    "output_dir": "[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]"
```

michellewang (changes requested):
Thanks @david-roper!

I have some additional comments concerning the global config and the Nipoppy paths in the descriptors/invocations. There are some failing tests too, see my proposed fix.

It seems to me like some inputs in the descriptors are still missing a `"list": true` field, but I could be wrong.

I tested the Boutiques command generation locally and am getting these commands, let me know if they look right to you (the binds paths and other container things are auto-added by Nipoppy):

**`preprocess` step**
```console
# CONTAINER THINGS
singularity run --cleanenv --bind /path/to/dataset_root/bids --bind /path/to/dataset_root/derivatives/rabies/0.5.1/output --bind /path/to/dataset_root/scratch/work/rabies-0.5.1/rabies-0.5.1 --bind /path/to/dataset_root/scratch/pybids_db/rabies-0.5.1 /path/to/container_store/nipoppy/rabies_0.5.1.sif
# ACTUAL RABIES COMMAND
-p Linear \
--local_threads 24 \
--force \
preprocess  \
--anat_inho_cor \
method=SyN,otsu_thresh=2,multiotsu=false \
/path/to/dataset_root/bids \
/path/to/dataset_root/derivatives/rabies/0.5.1/output
```

**`confound-correction` step**
Currently failing because the invocation is missing a required field, but after fixing it I get:
```console
# CONTAINER THINGS
singularity run --cleanenv --bind /path/to/dataset_root/bids --bind /path/to/dataset_root/derivatives/rabies/0.5.1/output --bind /path/to/dataset_root/scratch/work/rabies-0.5.1/rabies-0.5.1 --bind /path/to/dataset_root/scratch/pybids_db/rabies-0.5.1 /path/to/container_store/nipoppy/rabies_0.5.1.sif 
# ACTUAL RABIES COMMAND
-p Linear \
--local_threads 24 \
--force \
confound_correction \
--frame_censoring FD_censoring=false,FD_threshold=0.05,DVARS_censoring=false,minimum_timepoint=3 \
/path/to/dataset_root/derivatives/rabies/0.5.1/output \
/path/to/dataset_root/derivatives/rabies/0.5.1/output
```

**`analysis` step**
After removing an extra `[` in the invocation for `output_dir`:
```console
# CONTAINER THINGS
singularity run --cleanenv --bind /path/to/dataset_root/bids --bind /path/to/dataset_root/derivatives/rabies/0.5.1/output --bind /path/to/dataset_root/scratch/work/rabies-0.5.1/rabies-0.5.1 --bind /path/to/dataset_root/scratch/pybids_db/rabies-0.5.1 /path/to/container_store/nipoppy/rabies_0.5.1.sif
# ACTUAL RABIES COMMAND
-p Linear \
--local_threads 24 \
--force \
analysis  \
/path/to/dataset_root/derivatives/rabies/0.5.1/output \
/path/to/dataset_root/derivatives/rabies/0.5.1/output
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/tracker_config.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "NAME": "pipeline_complete",
+        "PATHS": [
+        ]
+    }
+]
```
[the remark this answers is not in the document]
david-roper: @michellewang The analysis outputs only show up when the `--data_diagnosis` tag is used. I'll add the file paths of these output mentioned in the documentation while having the `--data_diagnosis` tag set as true in the invocation file. However this may cause tracking issues in the future if users decide to not use the tag for their file.

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/tracker_config.json
michellewang, line 1: @david-roper have you tried running `nipoppy track`? Just to make sure that it's not a problem if the paths start with `/`

nipoppy/data/examples/sample_global_config-latest_pipelines.json
michellewang, line 1: The tests are failing because we need to add the `ANALYSIS_LEVEL` and `GENERATE_PYBIDS_DATABASE` fields to match the other sample global config file

michellewang (commented):
Thanks @david-roper! 

`sample_global_config-latest_pipelines.json` is missing some changes (which is why the tests are failing). Otherwise the last thing would just be to check that the tracker config works (if you haven't already).

Then this should be ready to merge!

nipoppy/data/examples/sample_global_config-latest_pipelines.json
david-roper: @michellewang I've added the suggested changes and the tests seem to be passing atm. I also have a question about the tracker file. Is it possible to have files conditionally tracked in the tracker config? For example, the current files I have in my `tracker_config.json` should only be tracked when the `--data_diagnosis` tag is true in my invocation file, or else they will not exist. Thank you for all the help!

michellewang: @david-roper I will answer on the issue you opened about the conditional tracking later today, but for my understanding of RABIES specifically: what does the `analysis` step do if the `--data_diagnosis` flag is not given? Because if it only prints some information to the screen then we not want a default tracker for it. 

Related: would it be relevant to track outputs of the `preprocessing` and `confound-correction` steps?

michellewang: @david-roper I just realized that the config as it is now is not ideal, because from what I understand RABIES handles the looping over participants internally, so from the Nipoppy perspective it should run at the the `group`analysis level (otherwise for e.g. the `analysis` step there will be unnecessary looping on the Nipoppy side). However, the tracking can only be done at the `participant_session` level (this was a design choice for reasons mostly related to what Neurobagel can handle as input). So in the case of RABIES we might need something like that in the pipeline config:
- 3 steps with invocation and descriptor files, with `ANALYSIS_LEVEL="group"`
- 3 additional steps with tracker files and default `ANALYSIS_LEVEL` (`"participant_session"`)
  - They could be named maybe something `preprocess-tracking`, `confound-correction-tracking` and `analysis-tracking`

This is probably not ideal but is the only way I can think of that would avoid `nipoppy run` looping around every single participant and session and rerunning the same RABIES command for no reason (while still having tracking at the participant-session level).

david-roper: @michellewang Thank you for the suggestions. It is true that Rabies handles the looping over participants internally so running it at the group level would be best. 

I will work on making the tracking steps in the pipeline as well as a separate tracker file for each of them.

david-roper: @michellewang I've completed the tracker changes. There are now a total of 9 new files in the rabies pipeline folder (3 invocation, 3 descriptor, 3 tracker_config files). I've run through the code and the trackers seem to run successfully and the outputs are consistent. If there's any other suggestion or sections of code I need to add let me know. Thanks!!

michellewang: @david-roper thanks! I'll take a look later this week

nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -331,11 +331,35 @@
...
-                    "GENERATE_PYBIDS_DATABASE": false
+                    "GENERATE_PYBIDS_DATABASE": false,
+                    "ANALYSIS_LEVEL": "group"
+                },
+                {
+                    "NAME": "preprocess-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "ANALYSIS_LEVEL": "participant_session",
+                    "GENERATE_PYBIDS_DATABASE": false,
```
michellewang, line 342: To avoid confusion we can remove all these fields, this way the `preprocess-tracking` step will only work with `nipoppy track` (not `nipoppy run`)

nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -331,11 +331,35 @@
...
+                    "TRACKER_CONFIG_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/[[STEP_NAME]]_tracker_config.json"
+                },
+                {
+                    "NAME": "confound-correction-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "ANALYSIS_LEVEL": "participant_session",
+                    "GENERATE_PYBIDS_DATABASE": false,
```
michellewang, line 350: Same comment as above
```suggestion
```

nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -331,11 +331,35 @@
...
+                    "TRACKER_CONFIG_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/[[STEP_NAME]]_tracker_config.json"
+                },
+                {
+                    "NAME": "analysis-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "GENERATE_PYBIDS_DATABASE": false,
+                    "ANALYSIS_LEVEL": "participant_session",
```
michellewang, line 358: Same comment as above

```suggestion
```

nipoppy/data/examples/sample_global_config-latest_pipelines.json
```diff
@@ -225,21 +225,51 @@
...
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "GENERATE_PYBIDS_DATABASE": false,
+                    "ANALYSIS_LEVEL": "group"
+                },
+                {
+                    "NAME": "preprocess-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "ANALYSIS_LEVEL": "participant_session",
+                    "GENERATE_PYBIDS_DATABASE": false,
```
michellewang, line 252: Same comment as for `sample_global_config-all_pipelines.json`
```suggestion
```

nipoppy/data/examples/sample_global_config-latest_pipelines.json
```diff
@@ -225,21 +225,51 @@
...
+                    "TRACKER_CONFIG_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/[[STEP_NAME]]_tracker_config.json"
+                },
+                {
+                    "NAME": "confound-correction-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "ANALYSIS_LEVEL": "participant_session",
+                    "GENERATE_PYBIDS_DATABASE": false,
```
michellewang, line 260: Same comment as above

```suggestion
```

nipoppy/data/examples/sample_global_config-latest_pipelines.json
```diff
@@ -225,21 +225,51 @@
...
+                    "TRACKER_CONFIG_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/[[STEP_NAME]]_tracker_config.json"
+                },
+                {
+                    "NAME": "analysis-tracking",
+                    "INVOCATION_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/invocation-[[STEP_NAME]].json",
+                    "DESCRIPTOR_FILE": "[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/descriptor-[[STEP_NAME]].json",
+                    "GENERATE_PYBIDS_DATABASE": false,
+                    "ANALYSIS_LEVEL": "participant_session",
```
michellewang, line 268: Same comment as above

```suggestion
```

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/analysis-tracking_tracker_config.json
michellewang, line 1: You might want to use `*` in the paths so that the tracking is more flexible. Otherwise if the user's BIDS data doesn't have exactly these BIDS entities the tracking will fail. 

For example, for the first path you could have something like this instead:
```
"analysis_datasink/dual_regression_nii/_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]*_bold/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]*_bold_RAS_combined_cleaned_DR_maps.nii.gz"
```

But I am not familiar with the pipeline, so if you expect all outputs to have `_task-rest_acq-EPI_run-01` bit in the paths then feel free to ignore my comment

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/confound-correction-tracking_tracker_config.json
michellewang, line 1: Same comment as above re: BIDS entities

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-analysis-tracking.json
michellewang, line 1: I think this can be deleted given the change I proposed in the sample global config files

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-confound-correction-tracking.json
michellewang, line 1: Same here: I think this can be deleted given the change I proposed in the sample global config files

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/descriptor-preprocess-tracking.json
michellewang, line 1: Same here also: I think this can be deleted given the change I proposed in the sample global config files

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/invocation-analysis-tracking.json
michellewang, line 1: Should also be deleted

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/invocation-confound-correction-tracking.json
michellewang, line 1: Should also be deleted

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/invocation-preprocess-tracking.json
michellewang, line 1: Should also be deleted

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/preprocess-tracking_tracker_config.json
michellewang, line 1: Same comment as above re: BIDS entities

michellewang (changes requested):
Thanks @david-roper! Two comments:
1. We don't need invocation/descriptor files for the `*-tracking` steps. It is okay (and preferred in my opinion) to have only the tracker config files for them
2. It might be better to allow for globbing in the tracker paths (i.e. use `*` instead of hardcoded BIDS entities like `run`) 

Otherwise looks good! Let me know if you have any questions!

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/analysis-tracking_tracker_config.json
david-roper: Yeah this would makes sense to do. Although the output file paths seem to be consistent having the flexibility is still good.

david-roper:
@michellewang Thank you for the suggestions! I have implemented them into code. Let me know if there's anything else needed. Thanks again for all the help on this pr!

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/preprocess-tracking_tracker_config.json
```diff
@@ -1,14 +1,14 @@
...
-        "transforms_datasink/bold_to_anat_inverse_warp/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session3_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w/_run_1/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_task-rest_acq-EPI_run-1_bold_RAS_inho_cor_output_1InverseWarp.nii.gz",
-        "transforms_datasink/bold_to_anat_warp/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session3_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w/_run_1/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_task-rest_acq-EPI_run-1_bold_RAS_inho_cor_output_1Warp.nii.gz"
-    ]
+  "PATHS": [
+    "anat_datasink/anat_preproc/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session3_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w_RAS_inho_cor.nii.gz",
```
michellewang, line 3: Are the `T2w` files always expected to have `acq-RARE`? If not we can replace by `*` here as well

michellewang, line 3: Also, is the `session3` hardcoded in RABIES or does it depend on the data?

michellewang (commented):
I just have some remaining questions regarding the tracker config for the `analysis` step. Once these are addressed we can finally merge this :)

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/preprocess-tracking_tracker_config.json
david-roper: Yes that can be replaced with a `*`

david-roper: This section of the file path is dependent on the session number of the subject. So it will be using the `[[NIPOPPY_BIDS_SESSION_ID]]` replacement for it. Good catch!

david-roper:
@michellewang I have added the last changes to the pr. However it seems that the docs build test is failing within the CI now. Would pulling the most recent changes from main resolve this? Thanks!

nipoppy/data/examples/sample_pipelines/rabies-0.5.1/preprocess-tracking_tracker_config.json
```diff
@@ -1,14 +1,14 @@
...
-    "anat_datasink/anat_preproc/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session3_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w_RAS_inho_cor.nii.gz",
-    "bold_datasink/commonspace_bold/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session3_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_acq-RARE_T2w/_run_1/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]*_bold_RAS_combined.nii.gz",
+    "anat_datasink/anat_preproc/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session[[NIPOPPY_BIDS_SESSION_ID]]_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_*_T2w/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_*_T2w_RAS_inho_cor.nii.gz",
+    "bold_datasink/commonspace_bold/_scan_info_subject_id[[NIPOPPY_BIDS_PARTICIPANT_ID]].session[[NIPOPPY_BIDS_SESSION_ID]]_split_name_sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]_*_T2w/_run_1/sub-[[NIPOPPY_BIDS_PARTICIPANT_ID]]_ses-[[NIPOPPY_BIDS_SESSION_ID]]*_bold_RAS_combined.nii.gz",
```
michellewang, line 4: @david-roper so `[[NIPOPPY_BIDS_PARTICIPANT_ID]]` and `[[NIPOPPY_BIDS_SESSION_ID]]` all include the BIDS prefixes (`sub-` and `ses-` respectively). If you don't want the prefixes then it should be `[[NIPOPPY_PARTICIPANT_ID]]` and `[[NIPOPPY_SESSION_ID]]`. Just pointing it out because if the session ID is `3` then `session[[NIPOPPY_BIDS_SESSION_ID]]` will become `sessionses-3` which is probably not what you want. So you should really make sure to run `nipoppy track` on a participant that has completed the pipeline successfully and check that all expected paths are correct

david-roper: @michellewang Good catch is also noticed in the preprocess tracker i have paths where `[[NIPOPPY_BIDS_PARTICIPANT_ID]]` would not provide me the correct tag as I would need their id without the `sub-` attached to it. In these cases I have replaced it with `*` instead of `[[NIPOPPY_BIDS_PARTICIPANT_ID]]`

david-roper:
@michellewang Everything seems to be in working order now. The tracker configs all have successful readings in the `imaging_bagel.tsv` when run. As the all the regular steps of the rabies pipeline run successfully. If there is anything else needed to be added let me know. Thank you yet again for all the help on this PR!

michellewang (approved):
Looks good @david-roper! I will go ahead and merge this, though I also wanted to let you know that we are working making a big change to the way pipeline configs work with Nipoppy. Briefly:
- Currently, pipeline updates or contributions require a PR to the `nipoppy/nipoppy` repo, and new changes will only be obtainable by regular users when Nipoppy makes a new release, which is not ideal
- So we want to decouple the sample Nipoppy pipelines from the core Nipoppy CLI code
- Instead of being in the `sample_pipelines` directory, the currently supported pipelines will be uploaded to Zenodo and look something like [this](https://sandbox.zenodo.org/records/197945). The Nipoppy CLI will have new commands for uploading/downloading pipelines from Zenodo
- This means that for example if you ever need to change the RABIES config, you can update the Zenodo record without going through this PR process (though we will still be happy to give feedback on pipeline configs if needed), and the new record will be usable immediately by other Nipoppy users

This is actively being worked on in the [`dev-catalog`](https://github.com/nipoppy/nipoppy/tree/dev-catalog) branch and will be a breaking change once merged/released. For RABIES specifically you will need to change 3 things:
- Move the RABIES part of the global config into a `config.json` file inside the `rabies-0.5.1` directory
- Update the paths to the descriptor/invocation/tracker config files in `config.json` so that they are relative to `rabies-0.5.1/` (i.e. remove the `[[NIPOPPY_DPATH_PIPELINES]]/[[PIPELINE_NAME]]-[[PIPELINE_VERSION]]/` part).
Then run something like `nipoppy pipeline upload /path/to/rabies-0.5.1` with a Zenodo API token
- Add a field `"PIPELINE_TYPE": "processing"` to `config.json`

Let me know if this makes sense and/or if you have any questions. I will let you know when the Zenodo things are ready (hopefully by the end of the month)

Issue #239, opened 2024-05-17
Refactor container-related components of Boutiques descriptors


michellewang:
- Right not the Boutiques descriptors cannot be run outside of Nipoppy because of the hardcoded `[[NIPOPPY_CONTAINER_COMMAND]] [[NIPOPPY_FPATH_CONTAINER]]` in the `"command-line"` field.
- It would be better to change them to actual Boutiques inputs (e.g., `[CONTAINER_COMMAND]` and `FPATH_CONTAINER`, with defaults being `[[NIPOPPY_CONTAINER_COMMAND]]` and `[[NIPOPPY_FPATH_CONTAINER]]`)
- This would not change the current behaviour at all, but would allow people to reuse the descriptors for other purposes (except that they might have to specify empty strings for the extra arguments)

michellewang:
We also decided to no longer have Nipoppy-specific default values in the descriptor files themselves -- instead they will be stored in the default/sample invocation files

michellewang:
It turns out that this is trickier than initially thought, because Boutiques will surround the container command with quotes (for security purposes). 

Tentative descriptor modifications (for future reference):
```json
{
    "command-line": "[CONTAINER_WRAPPER] dcm2bids [DICOM_DIR] [PARTICIPANT] [SESSION] [CONFIG] [OUTPUT_DIR] [AUTO_EXTRACT_ENTITIES] [BIDS_VALIDATE] [FORCE_DCM2BIDS] [SKIP_DCM2NIIX] [CLOBBER] [LOG_LEVEL] [VERSION]",
    "inputs": [
        {
            "id": "container_wrapper",
            "name": "container_wrapper",
            "description": "The full container command to run the tool, for example 'apptainer run container.sif'.",
            "optional": true,
            "type": "String",
            "value-key": "[CONTAINER_WRAPPER]",
            "default-value": ""
        }
    ]
}
```

Corresponding invocation:
```json
{
    "container_wrapper": "[[NIPOPPY_CONTAINER_COMMAND]] [[NIPOPPY_FPATH_CONTAINER]]",
    "dicom_dir": [
        "[[NIPOPPY_DPATH_SOURCEDATA]]/[[NIPOPPY_PARTICIPANT]]/[[NIPOPPY_SESSION]]"
    ],
    "participant": "[[NIPOPPY_PARTICIPANT]]",
    "session": "[[NIPOPPY_SESSION_SHORT]]",
[...]

PR #265, opened 2024-06-13
[ENH] Add default tracker config files for fMRIPrep and FreeSurfer, and update default invocation files

### LINKED ISSUE #252: Add more default tracker config files
For all currently supported pipelines, not just MRIQC

Also add empty PyBIDS ignore files (?) just so that users can see it in the default config

michellewang:
- Closes #252

Summary:
- Added default tracker configs based on old `fmriprep_tracker.py` and `fs_tracker.py` files
  - Note: the global config file now had pipeline configurations for Freesurfer, though it is only supposed to be used for tracking since Freesurfer is run as part of fMRIPrep
- Updated invocations so that Nipoppy-specific things are generally in the invocation file only
  - The exception is the container-related template strings, see #239


nipoppy_cli/nipoppy/data/descriptors/heudiconv-0.12.2-prepare.json
michellewang, line 1: Deleted since Heudiconv only has 1 CLI command, and now the descriptor is the same for `prepare` and `convert` since we removed the defaults

nipoppy_cli/nipoppy/data/examples/sample_invocations/heudiconv-0.12.2-prepare.json
```diff
@@ -1,3 +1,9 @@
 {
+    "dicom_dir_template": "[[NIPOPPY_DPATH_SOURCEDATA]]/[[NIPOPPY_PARTICIPANT_ID]]/[[NIPOPPY_BIDS_SESSION]]/*",
+    "subjs": "[[NIPOPPY_PARTICIPANT_ID]]",
+    "converter": "none",
```
nikhil153, line 4: why put this `converter` flag in invocation and not in descriptor like for the `convert` step? 
Same for `heuristic` flag below.

nikhil153 (approved):
It wasn't letting me add an in-line comment but I think it would be good to have [output_spaces](https://github.com/neurodatascience/nipoppy/pull/265/files#diff-bc027192d84e362c8c5e03b1aae3308d1a48351c45d4c7fa446763607cdf72e3R229) in invocations for fmriprep.

michellewang:
@nikhil153 what should we populate `output_spaces` with?

nipoppy_cli/nipoppy/data/examples/sample_invocations/heudiconv-0.12.2-prepare.json
michellewang: The `convert` step uses the default `converter` (`dcm2niix`), which is already in the descriptor by default. For `prepare` we explicitly set it to `none` so that it doesn't do the conversion. I can add it for the default if you think it would be better but I don't think users would want to change it since the only two options are `dcm2niix` (do the conversion) or `none` (don't do the conversion).

For the `heuristic` flag, it is optional and not used in the `prepare` step so shouldn't be specified then. And usually at that stage in the process we don't even have a heuristic file.

michellewang:
@nikhil153 I'm thinking of renaming the descriptors:
- `dcm2bids-3.1.0-convert` -> `dcm2bids-3.1.0.json`
- `dcm2bids-3.1.0-prepare` -> `dcm2bids_helper-3.1.0.json`

Mostly because I'm looking at BIDSCOIN now and it has the `bidsmapper`, `bidseditor` and `bidscoiner` CLIs, and I'm wondering if it would make more sense to just name the descriptors based on the actual command name. 

However, I still think the invocations should probably still have the step name in them, so it might be confusing to have e.g. the descriptor called `dcm2bids_helper-3.1.0.json` and the invocation called `dcm2bids-3.1.0-prepare.json`?

Do you have any thoughts/preference?

nipoppy_cli/nipoppy/data/examples/sample_invocations/heudiconv-0.12.2-prepare.json
nikhil153: If we don't want users to change anything mistakenly and has only 1 possible option per step - wouldn't it better to keep these two in descriptor that maps 1-1 onto the step?

nikhil153:
> @nikhil153 what should we populate `output_spaces` with?

The default one, which i think is this: `MNI152NLin2009cAsym`. The reason we should explicitly add this into invocation is better visibility. Often people forget what space / resolution they use or how to change it before launching large number of jobs. This would help, hopefully!

nikhil153:
> @nikhil153 I'm thinking of renaming the descriptors:
> 
> * `dcm2bids-3.1.0-convert` -> `dcm2bids-3.1.0.json`
> * `dcm2bids-3.1.0-prepare` -> `dcm2bids_helper-3.1.0.json`
> 
> Mostly because I'm looking at BIDSCOIN now and it has the `bidsmapper`, `bidseditor` and `bidscoiner` CLIs, and I'm wondering if it would make more sense to just name the descriptors based on the actual command name.
> 
> However, I still think the invocations should probably still have the step name in them, so it might be confusing to have e.g. the descriptor called `dcm2bids_helper-3.1.0.json` and the invocation called `dcm2bids-3.1.0-prepare.json`?
> 
> Do you have any thoughts/preference?

I think mapping descriptors 1-1 on the tools CLI makes sense. It also detaches them from nipoppy terms which is consistent with what we discussed last time. 
As for the invocations, I think it's fine to name them differently since now we are in nipoppy scope where we try to abstract out / standardize run calls. From user perspective neither the `passengers` nor `crew members` will ever see descriptors, so it should be fine! :-)

nipoppy_cli/nipoppy/data/examples/sample_invocations/heudiconv-0.12.2-prepare.json
michellewang: In general I think it would be better to have it in invocation files (not the descriptor) for better visibility (same reasoning as your comment about the output spaces in fMRIPrep).

If people are looking at the invocations then we assume they sort of know what they are doing, so if they want to change the default invocation we provide then I say it's up to them. 

I can add `"converter": "dcm2niix"` to the invocation for the `convert` step (again, similar to `output_spaces` in fMRIPrep). But if we set it directly in the descriptor it would lose visibility, and I would rather take the chance that users mess up their invocations than users not knowing that we are changing the default behaviour of the software through the descriptor.

For `heuristic`, hardcoding it in the descriptor for `convert` is doable (it's what we had before this PR), but the whole point of this PR was that we wanted to move the Nipoppy-specific default values outside of the descriptor and inside the invocation...

michellewang:
> The default one, which i think is this: MNI152NLin2009cAsym.

Okay, I'll add that. In the tracker I have `*` instead of the specific `space` entities -- I assume that is still fine? This way even if users change the output spaces then the tracker would still work out-of-the-box.

PR #266, opened 2024-06-13
[ENH] add QSIPrep

michellewang:
Work in progress!

- [x] Descriptor file
- [ ] Default invocation file
- [ ] Default tracker config file
- [ ] Update sample global config files

### DIFF
--- a/nipoppy_cli/nipoppy/data/descriptors/qsiprep-0.21.4.json
+++ b/nipoppy_cli/nipoppy/data/descriptors/qsiprep-0.21.4.json
@@ -1,0 +1,734 @@
+{
+    "name": "qsiprep",
+    "description": "qsiprep",
+    "tool-version": "0.21.4",
+    "schema-version": "0.5",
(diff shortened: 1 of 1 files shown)


michellewang:
Added a basic invocation file which creates the following command (most of these are QSIPrep defaults)

```console
Generated Command:
[[NIPOPPY_CONTAINER_COMMAND]] [[NIPOPPY_FPATH_CONTAINER]] qsiprep \
    '[[NIPOPPY_DPATH_BIDS]]' '[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]' participant \
    --skip_bids_validation --participant_label '[[NIPOPPY_PARTICIPANT_ID]]' \
    --bids-database-dir '[[NIPOPPY_DPATH_PIPELINE_BIDS_DB]]' --recon-input-pipeline qsiprep \
    --omp-nthreads 4 -v -v --anat-modality T1w --b0-threshold 100 --dwi_denoise_window auto \
    --denoise-method dwidenoise --b1-biascorrect-stage final --distortion-group-merge none \
    --anatomical-template MNI152NLin2009cAsym --output-resolution 1.2 --b0-to-t1w-transform Rigid \
    --intramodal-template-transform BSplineSyN --b0-motion-corr-to iterative --hmc-transform Affine \
    --hmc_model eddy --shoreline_iters 2 --skull-strip-template OASIS \
    --fs-license-file '[[FREESURFER_LICENSE_FILE]]' --pepolar-method TOPUP \
    --fmap-no-demean -w '[[NIPOPPY_DPATH_PIPELINE_WORK]]'
```

**Notes**
- QSIPrep does not have a "session" argument, and with multi-session studies where sessions are processed separately (like we do) there might be overwriting of files in `sub-<PARTICIPANT_ID>/anat`, see https://github.com/PennLINC/qsiprep/issues/757. Not sure if that is a problem for us, but if it is then we might have to do an outer `ses-<SESSION_ID>` directory like we do for Freesurfer
- There is an argument `freesurfer_input`: we probably want to point this to the Freesurfer output from fMRIPrep to avoid running Freesurfer twice. To be tested I think

nipoppy/data/examples/sample_global_config-all_pipelines.json
```diff
@@ -225,6 +225,32 @@
...
+        {
+            "NAME": "qsiprep",
+            "VERSION": "0.24.0",
+            "CONTAINER_INFO": {
+                "FILE": "[[NIPOPPY_DPATH_CONTAINERS]]/[[PIPELINE_NAME]]_[[PIPELINE_VERSION]].sif",
+                "URI": "docker://pennbbl/[[PIPELINE_NAME]]:[[PIPELINE_VERSION]]"
+            },
+            "CONTAINER_CONFIG": {
+                "ENV_VARS": {
+                    "TEMPLATEFLOW_HOME": "[[TEMPLATEFLOW_HOME]]"
```
bcmcpher, line 238: I think this is safe to keep, but I don't think they use templateflow?

nipoppy/data/examples/sample_global_config-latest_pipelines.json
```diff
@@ -163,6 +163,32 @@
...
+            },
+            "CONTAINER_CONFIG": {
+                "ENV_VARS": {
+                    "TEMPLATEFLOW_HOME": "[[TEMPLATEFLOW_HOME]]"
+                },
+                "ARGS": [
+                    "--bind",
+                    "[[FREESURFER_LICENSE_FILE]]",
+                    "--bind",
+                    "[[TEMPLATEFLOW_HOME]]"
```
bcmcpher, line 182: There current docs don't use TemplateFlow, but they do use freesurfer this way, so I think it's safe to keep.

bcmcpher (commented):
Other than TemplateFlow being kept in a few spots when it currently isn't used by QSIPrep, it keeps the symmetry with fMRIPrep and is a likely future addition.

Assuming the last of our tests pass, I think this is ready once they have a full release.

michellewang:
I think they do use Templateflow? https://github.com/search?q=repo%3APennLINC%2Fqsiprep%20templateflow&type=code

But not 100% sure

bcmcpher (approved):
Verified TemplateFlow dependency is in code, not necessarily documentation.
Good to go.

PR #267, opened 2024-06-14
[ENH] Add descriptors and sample invocations for BIDScoin

### LINKED ISSUE #240: Add support for BIDSCOIN
Source code: https://github.com/Donders-Institute/bidscoin/tree/master

BIDSCOIN uses `argparse` so this should not be very difficult, ~though it requires PyQt6 for the GUI~ (edit: there are branches for PyQt5 versions which also seem to be used by the container).

Also has an container image available: https://bidscoin.readthedocs.io/en/stable/installation.html#building-the-container-image.
```
sudo apptainer build bidscoin.sif docker://marcelzwiers/bidscoin:<version>
```

michellewang:
- Closes #240

BIDScoin has three main CLI commands:
1. `bidsmapper`: make the bidsmap
2. `bidseditor`: edit the bidsmap
3. `bidscoiner`: to the conversion

1 and 2 above involve a GUI, which is technically possible even with a container according to the [docs](https://bidscoin.readthedocs.io/en/stable/installation.html#run-bidscoin-tools-in-the-container) (using `xhost`), though I wasn't able to make that work on Compute Canada. But in any case it is still possible to modify the descriptor and use a baremetal install


nipoppy_cli/nipoppy/data/descriptors/bidscoiner-4.3.2.json
```diff
@@ -0,0 +1,79 @@
+{
+    "name": "bidscoiner",
+    "description": "bidscoiner",
```
nikhil153, line 3: Should this (and editor, mapper below) have more detail to explain the purpose?

nikhil153 (approved):

nipoppy_cli/nipoppy/data/descriptors/bidscoiner-4.3.2.json
michellewang: Updated

PR #273, opened 2024-07-05
Created user personas file for the docs

nikhil153:
We expect three types of user groups (i.e. personas) that would use nipoppy in various ways. These user personas are described in this doc.


michellewang:
@nikhil153 I think we can close this PR since now it has nothing in it. If you put the personas file in a new branch you can open a new PR from there

PR #283, opened 2024-07-12
[FIX] Fix `nipoppy run` creating empty PyBIDS database on second (and later) loop iteration

### LINKED ISSUE #282: BIDS database is empty for all but the first participant if multiple participants are run
Need to start PyBIDS ignore list from scratch, not update existing one

michellewang:
- Closes #282

### DIFF
--- a/nipoppy_cli/nipoppy/workflows/pipeline.py
+++ b/nipoppy_cli/nipoppy/workflows/pipeline.py
@@ -270,20 +270,22 @@ class BasePipelineWorkflow(BaseWorkflow, ABC):
         """Set up the BIDS database."""
         dpath_bids_db: Path = Path(dpath_bids_db)
 
+        pybids_ignore_patterns = self.pybids_ignore_patterns.copy()
+
         if participant_id is not None:
             add_pybids_ignore_patterns(
-                current=self.pybids_ignore_patterns,
+                current=pybids_ignore_patterns,
                 new=f"^(?!/{BIDS_SUBJECT_PREFIX}({participant_id}))",
             )
         if session_id is not None:
             add_pybids_ignore_patterns(
-                current=self.pybids_ignore_patterns,
+                current=pybids_ignore_patterns,
                 new=f".*?/{BIDS_SESSION_PREFIX}(?!{session_id})",
             )
 
         self.logger.info(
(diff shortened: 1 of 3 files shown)


nipoppy_cli/nipoppy/workflows/pipeline.py
```diff
@@ -270,20 +270,22 @@ def set_up_bids_db(
         """Set up the BIDS database."""
         dpath_bids_db: Path = Path(dpath_bids_db)
 
+        pybids_ignore_patterns = self.pybids_ignore_patterns.copy()
+
         if participant_id is not None:
             add_pybids_ignore_patterns(
-                current=self.pybids_ignore_patterns,
+                current=pybids_ignore_patterns,
```
nikhil153, line 277: so this function updates things in place, without returning anything?

nikhil153: If i understand this correctly, it keeps growing the ignore list by adding (supposedly) completed participants?

nipoppy_cli/nipoppy/workflows/pipeline.py
```diff
@@ -270,20 +270,22 @@ def set_up_bids_db(
...
                 new=f"^(?!/{BIDS_SUBJECT_PREFIX}({participant_id}))",
             )
         if session_id is not None:
             add_pybids_ignore_patterns(
-                current=self.pybids_ignore_patterns,
+                current=pybids_ignore_patterns,
                 new=f".*?/{BIDS_SESSION_PREFIX}(?!{session_id})",
```
nikhil153, line 283: Just out of curiosity, how come the regex patterns for `participant` and `session` are different?

nikhil153 (approved):

nipoppy_cli/nipoppy/workflows/pipeline.py
michellewang: Participant is at the beginning of the string. Also the prefixes are different

nipoppy_cli/nipoppy/workflows/pipeline.py
michellewang: That's what it used to do (which caused the bug), but it doesn't do that anymore because the list is copied

PR #288, opened 2024-07-15
[FIX] Remove working directory in `PipelineRunner` instead of `BasePipelineWorkflow`

### LINKED ISSUE #287: `nipoppy track` removes working directory

michellewang:
- Closes #287

Previously the `run_cleanup` method in `BasePipelineWorkflow` (inherited by *both* `PipelineRunner` and `PipelineTracker`) removed the pipeline's working directory, which is a mistake because the tracker shouldn't care about that directory. So now only `PipelineRunner` does this cleanup.

### DIFF
--- a/nipoppy_cli/nipoppy/tabular/bagel.py
+++ b/nipoppy_cli/nipoppy/tabular/bagel.py
@@ -50,6 +50,10 @@ class BagelModel(BaseTabularModel):
     pipeline_complete: str = Field(
         description="The status of the pipeline run for this participant-visit pair"
     )
+    # this is needed for now for the dashboard for work, but should be removed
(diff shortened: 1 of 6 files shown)


nipoppy_cli/nipoppy/tabular/bagel.py
```diff
@@ -50,6 +50,10 @@ class BagelModel(BaseTabularModel):
     pipeline_complete: str = Field(
         description="The status of the pipeline run for this participant-visit pair"
     )
+    # this is needed for now for the dashboard for work, but should be removed
```
surchs:
```suggestion
    # this is needed for now for the dashboard to work, but should be removed
```

surchs (approved):
:cook:

PR #300, opened 2024-07-22
[FIX] Improve pipeline config substitutions

### LINKED ISSUE #298: [BUG] Omitting pipeline version does not work
Need to troubleshoot why, probably something to do with the substitutions in the global config file, since the correct version is already grabbed (log file shows version number), but things like the paths to the invocation/descriptor file have `None` instead of the version number

michellewang:
- Closes #298

Originally, the pipeline configuration dictionaries in the global config used [`BaseWorkflow.process_template_json`](https://github.com/neurodatascience/nipoppy/blob/72e841a7ae7f459e076dfb6fc256f185094f2235/nipoppy_cli/nipoppy/workflows/pipeline.py#L217) to apply pipeline name/version substitutions based on the workflow attributes (e.g. passed from the CLI), which was hacky since it meant that all instances of `[[NIPOPPY_PIPELINE_NAME]]` and `[[NIPOPPY_PIPELINE_VERSION]]` were replaced by the same string. Now that `--pipeline-version` is an optional argument,  `[[NIPOPPY_PIPELINE_VERSION]]` is sometimes replaced by `None`, which is undesirable. 

Now the substitution is done at the `BidsPipelineConfig`/`ProcPipelineConfig` level (on instantiation, through Pydantic validation), with new substitution strings `[[PIPELINE_NAME]]`, `[[PIPELINE_VERSION]]` and `[[STEP_NAME]]`. The new substitution method is based on the pipeline fields, so the configs for other pipelines are also processed correctly. 

Note that this is a breaking change which is not ideal since we have users now, but essentially users will have to do find-replace in their global config files for:
- `[[NIPOPPY_PIPELINE_NAME]]` -> `[[PIPELINE_NAME]]`
- `[[NIPOPPY_PIPELINE_VERSION]` -> `[[PIPELINE_VERSION]]`


nipoppy_cli/tests/test_workflow_init.py
```diff
@@ -33,23 +28,30 @@ def test_run(dpath_root: Path):
...
-        assert Path(
-            dpath_root, ATTR_TO_DPATH_MAP["dpath_tracker_configs"], fpath.name
-        ).exists()
+    # check that pipeline config files have been copied
+    for pipeline_configs in (
+        workflow.config.BIDS_PIPELINES,
+        workflow.config.PROC_PIPELINES,
+    ):
+        for pipeline_config in pipeline_configs:
+            assert (
```
mathdugre, line 37: Just a nitpick, but this part could be modularized to make the test less verbose.

e.g.

```python3
def exist_or_none(o: object, s: str) -> bool:
  if attr := getattr(o, s, None):
    return attr.exist()
  return True
```

mathdugre (approved):

PR #341, opened 2024-08-27
[ENH] Add `"ANALYSIS_LEVEL"` field to pipeline step config to allow group-level processing

### LINKED ISSUE #257: Allow pipelines that do group processing
Currently the processing/BIDS pipeline runners loop over available participant-session pairs, and the only way to run group-level processing/pipelines is to specify some arbitrary participant and session (which is a hack). 

We can add a flag in the custom Boutiques configuration in descriptor files so that the pipeline developer can disable that looping (i.e., if that flag is on, then `get_participants_sessions_to_run` just returns `[(None, None)]`).

michellewang:
- Closes #257

- New configuration field for pipeline steps: `"ANALYSIS_LEVEL"`, which can be one of:
  - `"participant_session"`(default): same behaviour as before, looping over every participant-session pair (given some constraints)
  - `"participant"`: only looping over participants
  - `"session"`: only looping over sessions
  - `"group"`: only 1 iteration
- Note: This introduces the possibility of the Boutiques invocation having `None` instead of a participant ID (for example) if the user did not specify the invocation correctly
- Note 2: Currently the trackers are still at the pipeline level, not the pipeline _step_ level because that would make things break with the dashboard, though I think tracking at the step level makes sense. We should probably have a longer discussion on the derivatives bagel schema (and phases/stages things) @alyssadai @nikhil153 (see also #258) 
- Updated the default BIDSCOIN pipeline step configs
- Some test cleanup

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/workflows/pipeline.py
+++ b/nipoppy/workflows/pipeline.py
@@ -413,7 +413,7 @@ class BasePipelineWorkflow(BaseWorkflow, ABC):
         """Log a summary message."""
         if self.n_total == 0:
             self.logger.warning(
-                "No participant-session pairs to run. Make sure there are no mistakes "
+                "No participants/sessions to run. Make sure there are no mistakes "
                 "in the input arguments, the dataset's manifest or config file, "
                 f"and/or check the doughnut file at {self.layout.fpath_doughnut}"
             )
@@ -430,12 +430,18 @@ class BasePipelineWorkflow(BaseWorkflow, ABC):
             else:
                 color = LogColor.PARTIAL_SUCCESS
 
-            self.logger.info(
-                (
-                    f"[{color}]{prefix} for {self.n_success} out of "
-                    f"{self.n_total} participant-session pairs{suffix}[/]"
+            if (
+                self.pipeline_config.get_analysis_level(self.pipeline_step)
+                == AnalysisLevelType.group
+            ):
+                message_body = "on the entire study"
+            else:
+                message_body = (
+                    f"for {self.n_success} out of "
+                    f"{self.n_total} participants/sessions"
                 )
-            )
+
+            self.logger.info(f"[{color}]{prefix} {message_body}{suffix}[/]")
 
         return super().run_cleanup()
 
--- a/nipoppy/workflows/pipeline.py
+++ b/nipoppy/workflows/pipeline.py
@@ -8,7 +8,7 @@ import re
 from abc import ABC, abstractmethod
 from functools import cached_property
 from pathlib import Path
-from typing import Optional
+from typing import Iterable, Optional, Tuple
 
 import bids
 from pydantic import ValidationError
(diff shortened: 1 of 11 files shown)


mathdugre:
@michellewang , is this ready for review? Last time we talked, you mentioned some minor changes were required before review.

michellewang:
@mathdugre yes, made those changes yesterday so we are ready to go!

nipoppy/workflows/pipeline.py
```diff
@@ -39,6 +40,36 @@
...
+        sessions = []
+        for _, session in participants_sessions:
+            if session not in sessions:
+                sessions.append(session)
+        participants_session_updated = [(None, session) for session in sessions]
+
+    else:
+        participants_session_updated = participants_sessions
+
+    return participants_session_updated
```
mathdugre, line 70: I find it easier to read when the return statement is within the condition (see below). It avoids having to read the rest of the code to see if something modify the variable. Although, I don't know what the best practice is.

```python3
def apply_analysis_level(
    participants_sessions: Iterable[str, str],
    analysis_level: AnalysisLevelType,
) -> Tuple[str, str]:
  """Filter participant-session pairs to run based on the analysis level."""
  if analysis_level == AnalysisLevelType.group:
       return [(None, None)]

    elif analysis_level == AnalysisLevelType.participant:
        participants = []
        for participant, _ in participants_sessions:
            if participant not in participants:
                participants.append(participant)
        return [(participant, None) for participant in participants]

    elif analysis_level == AnalysisLevelType.session:
        sessions = []
        for _, session in participants_sessions:
            if session not in sessions:
                sessions.append(session)
        return [(None, session) for session in sessions]

    else:
        return participants_sessions

```

nipoppy/workflows/pipeline.py
```diff
@@ -373,7 +413,7 @@ def run_cleanup(self):
         """Log a summary message."""
         if self.n_total == 0:
             self.logger.warning(
-                "No participant-session pairs to run. Make sure there are no mistakes "
+                "No participants/sessions to run. Make sure there are no mistakes "
```
mathdugre, line 416: Nitpick, but I would put "or" instead of "/"
https://www.plainlanguage.gov/guidelines/conversational/dont-use-slashes/

tests/test_workflow_pipeline.py
```diff
@@ -593,42 +607,64 @@ def test_run_main_catch_errors(workflow: PipelineWorkflow):
 
 
 @pytest.mark.parametrize(
-    "n_success,n_total,expected_message",
+    "n_success,n_total,analysis_level,expected_message",
     [
-        (0, 0, "No participant-session pairs to run"),
+        (0, 0, "participant_session", "No participants/sessions to run"),
```
mathdugre, line 612: Same comment as before for the "/". Applies to all changes within this test

mathdugre (approved):
nitpicks only. Feel free to ignore them

PR #345, opened 2024-08-30
[BUG] keep working directory on failed run / flag to keep pipeline working directory on success

### LINKED ISSUE #297: Pipeline runner removes work directory regardless of job success / fail
It should keep working directory on pipeline failure to easily resume - currently it always deletes.

Ideally we can have it remove on success but have a flag to keep working directory regardless (for dev / testing new tools).

Snippet I think is the problem is here.
https://github.com/neurodatascience/nipoppy/blob/main/nipoppy_cli/nipoppy/workflows/runner.py#L209-L214

bcmcpher:
- Closes #297

- Modified `.run_cleanup()` method to not remove working files in failed processes to facilitate resuming  / restarting.
- Added additional flag to `PipelineRunner()` to optionally keep working files even on success.

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For bug fixes:


nipoppy/cli/parser.py
```diff
@@ -148,6 +148,17 @@ def add_arg_verbosity(parser: _ActionsContainer) -> _ActionsContainer:
     return parser
 
 
+def add_arg_keepworkdir(parser: _ActionsContainer) -> _ActionsContainer:
+    """Add a --keep_workdir argument to the parser."""
+    parser.add_argument(
+        "--keep_workdir",
```
michellewang, line 154: Use kebab-style to be consistent with other CLI args
```suggestion
        "--keep-workdir",
```

Also should update docstring after this change is made

nipoppy/workflows/runner.py
```diff
@@ -208,7 +210,17 @@ def run_single(self, participant_id: str, session_id: str):
 
     def run_cleanup(self):
         """Run pipeline runner cleanup."""
-        for dpath in [self.dpath_pipeline_bids_db, self.dpath_pipeline_work]:
-            if dpath.exists():
-                self.rm(dpath)
+        if self.n_success == self.n_total:
+            self.logger.info("All pipeline segments completed successfully.")
```
michellewang, line 214: Don't think this is needed since there is already a feedback message printed (from [here](https://github.com/nipoppy/nipoppy/blob/bc0f04ae5d7878c105d0228443d192acfd8b18a6/nipoppy/workflows/pipeline.py#L393-L398))

nipoppy/workflows/runner.py
```diff
@@ -208,7 +210,17 @@ def run_single(self, participant_id: str, session_id: str):
...
+            if not self.keep_workdir:
+                self.logger.info("Removing working / intermediary files.")
```
michellewang, line 216: Log message probably not needed since `BaseWorkflow.rm()` already prints a log message (see [here](https://github.com/nipoppy/nipoppy/blob/bc0f04ae5d7878c105d0228443d192acfd8b18a6/nipoppy/workflows/base.py#L264))

tests/test_workflow_runner.py
```diff
@@ -83,21 +83,23 @@ def test_run_setup(config: Config, tmp_path: Path):
     assert runner.dpath_pipeline_work.exists()
 
 
-def test_run_cleanup(tmp_path: Path):
+@pytest.mark.parametrize("keep_workdir", [True, False])
+def test_run_cleanup(tmp_path: Path, keep_workdir):
```
michellewang, line 87: This looks good but we should also include another test for different values of `n_success` and `n_total` (this is also what CodeCov is complaining about. See e.g. [this test](https://github.com/nipoppy/nipoppy/blob/main/tests/test_workflow_pipeline.py#L616) which sets `n_success` and `n_total` to different values with different expected results for each pair

nipoppy/cli/parser.py
```diff
@@ -148,6 +148,17 @@ def add_arg_verbosity(parser: _ActionsContainer) -> _ActionsContainer:
...
+        type=str,
+        required=False,
+        help="Keep pipeline working directory upon success (default: False).",
```
michellewang, line 157: Instead of `(default: False)`, I would put something like `(default: working directory deleted unless a run failed)".

Also ideally mention that this applies to the pyBIDS database too (if it was created)

michellewang (changes requested):
This looks good overall. The main thing would be to add a test that checks if the directories are kept when there are failures (even if `keep_workdir` is `False`). Otherwise minor comments about log message being redundant with other parts of the code.

michellewang (approved):
🧑‍🍳

michellewang (approved):

PR #358, opened 2024-09-16
[ENH] Allow to init a Nipoppy dataset with a specific BIDS dataset

Remi-Gau:
- Closes #

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/cli/parser.py
+++ b/nipoppy/cli/parser.py
@@ -148,6 +148,17 @@ def add_arg_verbosity(parser: _ActionsContainer) -> _ActionsContainer:
     return parser
 
 
+def add_arg_bids_source(parser: _ActionsContainer) -> _ActionsContainer:
+    """Add argument to init a layout with a BIDS datalad dataset."""
+    parser.add_argument(
+        "--bids-source",
+        type=str,
+        required=False,
+        help=("URL or path to a BIDS datalad dataset to initialize the layout with."),
+    )
+    return parser
+
+
 def add_subparser_init(
     subparsers: _SubParsersAction,
     formatter_class: type[HelpFormatter] = HelpFormatter,
@@ -162,6 +173,9 @@ def add_subparser_init(
         add_help=False,
     )
     parser = add_arg_dataset_root(parser)
+
+    parser = add_arg_bids_source(parser)
+
     return parser
 
 
--- a/nipoppy/workflows/dataset_init.py
+++ b/nipoppy/workflows/dataset_init.py
@@ -60,11 +60,13 @@ class InitWorkflow(BaseWorkflow):
         for dpath in self.layout.dpaths:
 
             # If a bids_source is passed it means datalad is installed.
+            print(self.bids_source)
+            print(dpath.stem)
             if self.bids_source is not None and dpath.stem == "bids":
                 self.logger.info(
                     f"Installing datalad BIDS raw dataset from {self.bids_source}."
                 )
-                copy_tree(self.bids_source, dpath)
+                copy_tree(self.bids_source, str(dpath))
 
             else:
                 self.mkdir(dpath)
--- a/nipoppy/workflows/dataset_init.py
+++ b/nipoppy/workflows/dataset_init.py
@@ -1,6 +1,7 @@
 """Workflow for init command."""
 
 import logging
+from distutils.dir_util import copy_tree
 from pathlib import Path
 from typing import Optional
 
@@ -63,18 +64,8 @@ class InitWorkflow(BaseWorkflow):
                 self.logger.info(
                     f"Installing datalad BIDS raw dataset from {self.bids_source}."
                 )
-                from datalad import api
+                copy_tree(self.bids_source, dpath)
 
-                dataset = None
-                if self.use_dalatad:
-                    dataset = self.dpath_root
-
-                api.install(
-                    dataset=dataset,
-                    path=dpath,
-                    source=self.bids_source,
-                    result_renderer="disabled",
-                )
             else:
                 self.mkdir(dpath)
 
--- a/nipoppy/workflows/dataset_init.py
+++ b/nipoppy/workflows/dataset_init.py
@@ -4,6 +4,8 @@ import logging
 from pathlib import Path
 from typing import Optional
 
+import pandas as pd
+
 from nipoppy.env import LogColor, StrOrPathLike
 from nipoppy.utils import (
     DPATH_DESCRIPTORS,
@@ -24,6 +26,7 @@ class InitWorkflow(BaseWorkflow):
     def __init__(
         self,
         dpath_root: Path,
+        bids_source=None,
         fpath_layout: Optional[StrOrPathLike] = None,
         logger: Optional[logging.Logger] = None,
         dry_run: bool = False,
@@ -37,20 +40,49 @@ class InitWorkflow(BaseWorkflow):
             dry_run=dry_run,
         )
         self.fname_readme = "README.md"
+        self.bids_source = bids_source
 
     def run_main(self):
-        """Create dataset directory structure."""
+        """Create dataset directory structure.
+
+        Create directories and add a readme in each.
+        Copy boutiques descriptors and invocations.
+        Copy default config files.
+
+        If the BIDS source dataset is requested, it is installed with datalad.
+        """
         # dataset must not already exist
         if self.dpath_root.exists():
             raise FileExistsError("Dataset directory already exists")
 
         # create directories
         for dpath in self.layout.dpaths:
-            self.mkdir(dpath)
+
+            # If a bids_source is passed it means datalad is installed.
+            if self.bids_source is not None and dpath.stem == "bids":
+                self.logger.info(
+                    f"Installing datalad BIDS raw dataset from {self.bids_source}."
+                )
+                from datalad import api
+
+                dataset = None
+                if self.use_dalatad:
+                    dataset = self.dpath_root
+
+                api.install(
+                    dataset=dataset,
+                    path=dpath,
+                    source=self.bids_source,
+                    result_renderer="disabled",
+                )
+            else:
+                self.mkdir(dpath)
 
         for dpath, description in self.layout.dpath_descriptions:
             fpath_readme = dpath / self.fname_readme
-            if description is not None and not self.dry_run:
+            if (description is not None and not self.dry_run) and not (
+                self.bids_source is not None and dpath.stem == "bids"
+            ):
                 fpath_readme.write_text(f"{description}\n")
 
         # copy descriptor files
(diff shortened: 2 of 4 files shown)


nipoppy/workflows/dataset_init.py
```diff
@@ -37,20 +41,38 @@ def __init__(
...
         # create directories
         for dpath in self.layout.dpaths:
-            self.mkdir(dpath)
+
+            # If a bids_source is passed it means datalad is installed.
+            if self.bids_source is not None and dpath.stem == "bids":
+                self.logger.info(
+                    f"Installing datalad BIDS raw dataset from {self.bids_source}."
+                )
+                copytree(self.bids_source, str(dpath))
```
Remi-Gau, line 67: note that no BIDS validation is done before copying the data

Remi-Gau: also this copies the data instead of symlink or moving it: to be discussed

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+                    x.name
+                    for x in (self.layout.dpath_bids / ppt).iterdir()
+                    if x.is_dir() and x.name.startswith("ses-")
+                ]
+            )
+            if not session_ids:
+                self.logger.warning(
+                    f"Skipping subject '{ppt}': could not find a session level folder."
+                )
+                continue
```
Remi-Gau, line 149: subjects without a session level are skipped from the manifest for now

nipoppy/workflows/dataset_init.py
```diff
@@ -37,20 +41,38 @@ def __init__(
...
+                copytree(self.bids_source, str(dpath))
+            else:
+                self.mkdir(dpath)
 
         for dpath, description in self.layout.dpath_descriptions:
             fpath_readme = dpath / self.fname_readme
-            if description is not None and not self.dry_run:
+            if (description is not None and not self.dry_run) and not (
+                self.bids_source is not None and dpath.stem == "bids"
+            ):
```
Remi-Gau, line 75: the BIDS dataset most likely already contains a README

michellewang: Maybe we could still add the Nipoppy README if `dpath / "README"` doesn't exist? Not sure how strict BIDS is with the name/extension of the README file though so it might not be a very simple check

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
 
+        if self.bids_source is not None:
+            self._init_manifest_from_bids_dataset()
+
         # inform user to edit the sample files
         self.logger.warning(
             f"Sample config and manifest files copied to {self.layout.fpath_config}"
             f" and {self.layout.fpath_manifest} respectively. They should be edited"
             " to match your dataset"
         )
```
michellewang, line 118: I think it would make sense to do something like:

```python
if self.bids_source is not None:
    self._init_manifest_from_bids_dataset()
else:
    self.copy(FPATH_SAMPLE_MANIFEST, self.layout.fpath_manifest, log_level=logging.DEBUG)
```

to avoid the unnecessary copying of the sample manifest that immediately gets overwritten.

Also we might want to split the warning onto one for the global config and one for the manifest so that if `--bids-source` is specified there is only a warning about the global config? Not sure

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+        """Assume a BIDS dataset with session level folders.
+
+        No BIDS validation is done.
+        """
+        df = {"participant_id": [], "visit_id": [], "session_id": [], "datatype": []}
+        participant_ids = sorted(
+            [
+                x.name
+                for x in (self.layout.dpath_bids).iterdir()
+                if x.is_dir() and x.name.startswith("sub-")
```
michellewang, line 130: Could use `BIDS_SUBJECT_PREFIX` in `nipoppy.env` here instead of `"sub-"`

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+            ]
+        )
+
+        self.logger.info("Creating a manifest.csv from the BIDS dataset content.")
```
michellewang, line 134:
```suggestion
        self.logger.info("Creating a manifest file from the BIDS dataset content.")
```
since we will change the manifest and other files to be TSV sometime in the future

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+
+        for ppt in participant_ids:
+
+            session_ids = sorted(
+                [
+                    x.name
+                    for x in (self.layout.dpath_bids / ppt).iterdir()
+                    if x.is_dir() and x.name.startswith("ses-")
```
michellewang, line 142: Could use `BIDS_SESSION_PREFIX` in `nipoppy.env` here instead of `"ses-"`

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
             f" and {self.layout.fpath_manifest} respectively. They should be edited"
             " to match your dataset"
         )
 
+    def _init_manifest_from_bids_dataset(self) -> None:
+        """Assume a BIDS dataset with session level folders.
+
+        No BIDS validation is done.
+        """
+        df = {"participant_id": [], "visit_id": [], "session_id": [], "datatype": []}
```
michellewang, line 125: With `from nipoppy.tabular.manifest import Manifest`:
```suggestion
        df = {
            Manifest.col_participant_id: [],
            Manifest.col_visit_id: [],
            Manifest.col_session_id: [],
            Manifest.col_datatype: [],
        }
```

to avoid hardcoding of column names

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+                        for x in (self.layout.dpath_bids / ppt / ses).iterdir()
+                        if x.is_dir()
+                    ]
+                )
+
+                df["participant_id"].append(ppt.replace("sub-", ""))
+                df["session_id"].append(ses.replace("ses-", ""))
+                df["datatype"].append("[" + "'" + "' ,'".join(datatypes) + "'" + "]")
+
+        df["visit_id"] = df["session_id"]
```
michellewang, line 164:
```suggestion
                df[Manifest.col_participant_id].append(check_participant_id(ppt))
                df[Manifest.col_session_id].append(check_session_id(ses))
                df[Manifest.col_datatype].append(datatypes)

        df[Manifest.col_visit_id] = df[Manifest.col_session_id]
```

Not 100% but I don't think `datatypes` needs to be converted to a string like that

nipoppy/workflows/dataset_init.py
michellewang: Also I think it is a little confusing to name this `df` since it is not yet a dataframe

nipoppy/workflows/dataset_init.py
```diff
@@ -85,13 +107,65 @@ def run_main(self):
...
+
+        df = pd.DataFrame(df)
+        df.to_csv(self.layout.fpath_manifest, index=False)
```
michellewang, line 167:
```suggestion
        manifest = Manifest(df).validate()
        self.save_tabular_file(manifest, workflow.layout.fpath_manifest)
```

`BaseWorkflow.save_tabular_file()` will create a file with a timestamp in the name, then make a symlink to `manifest.csv`. Which can be useful in non-Datalad datasets to keep older versions of the manifest file

tests/test_workflow_init.py
```diff
@@ -72,3 +73,28 @@ def test_run_cleanup(tmp_path: Path, caplog: pytest.LogCaptureFixture):
     workflow = InitWorkflow(dpath_root=tmp_path)
     workflow.run_cleanup()
     assert f"Successfully initialized a dataset at {workflow.dpath_root}" in caplog.text
+
+
+def test_init_bids(tmp_path):
```
michellewang, line 78: Maybe add a check that the manifest has been created and is valid?

E.g. after the workflow is run
```
from nipoppy.tabular.manifest import Manifest
assert isinstance(workflow.manifest, Manifest)
```

Ideally also check that the manifest rows match what is expected

michellewang (commented):
Thanks a lot @Remi-Gau!

Mostly just have some comments pointing out utilities/features in other parts of the codebase that could be used in some places

tests/test_workflow_init.py
michellewang: And to increase test coverage we would need a test checking that the warning is raised in a BIDS dataset without session-level folders

nipoppy/cli/parser.py
```diff
@@ -148,6 +148,17 @@ def add_arg_verbosity(parser: _ActionsContainer) -> _ActionsContainer:
     return parser
 
 
+def add_arg_bids_source(parser: _ActionsContainer) -> _ActionsContainer:
+    """Add argument to init a layout with a BIDS datalad dataset."""
+    parser.add_argument(
+        "--bids-source",
+        type=str,
+        required=False,
+        help=("URL or path to a BIDS datalad dataset to initialize the layout with."),
```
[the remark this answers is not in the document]
Remi-Gau: yup

Remi-Gau, line 157:
```suggestion
        help=("Path to a BIDS dataset to initialize the layout with."),
```

nipoppy/workflows/dataset_init.py
Remi-Gau: If we are adding a README to a BIDS dataset then I'd rather go with the default one from BIDS: https://github.com/bids-standard/bids-starter-kit/blob/main/templates/README.MD

nipoppy/workflows/dataset_init.py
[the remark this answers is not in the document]
Remi-Gau: done

Remi-Gau:
Addressed most of your points but I am still unsure about what's the most desirable default behavior: copy, move, copy with symlink?

nipoppy/workflows/dataset_init.py
michellewang: Having it copy by default makes sense since that is the safest option, but it would be good to have the option to create a symlink or move it just in case the BIDS dataset is huge

nipoppy/workflows/dataset_init.py
michellewang: I am okay with copying the default BIDS README to the `bids` subdirectory in all Nipoppy datasets

nipoppy/workflows/dataset_init.py
Remi-Gau: yeah: maybe let's keep it like this for now and maybe we can think of adding another CLI argument to say:
 
--mode: copy
--mode: move
--mode: link
--mode: datalad

nipoppy/workflows/dataset_init.py
Remi-Gau: do you prefer to have it in the repo or to fetch it from wherever it lives?

michellewang: I would fetch it from wherever it lives, and we could have a test to check that the remote file does exists

michellewang: We should also make sure that `nipoppy init` won't crash if the default BIDS README cannot be fetched for whatever reason

Remi-Gau: yeah that may also add "request" as a dependency, unless we want to stick to urllib

michellewang: I'm okay with a `requests` dependency, especially since we will (someday) move the pipeline files (descriptors, invocations, etc.) somewhere else and may want a relatively easy way for users to add a pipeline to their dataset

nipoppy/workflows/dataset_init.py
```diff
@@ -92,6 +119,83 @@ def run_main(self):
...
+                gh_org = "bids-standard"
+                gh_repo = "bids-starter-kit"
+                commit = "f2328c58238bdf2088bc587b0eb4198131d8ffe2"
+                path = "templates/README.MD"
+                url = (
+                    "https://raw.githubusercontent.com/"
+                    f"{gh_org}/{gh_repo}/{commit}/{path}"
+                )
+                response = requests.get(url)
+                fpath_readme.write_text(response.content.decode("utf-8"))
```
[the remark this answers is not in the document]
Remi-Gau: the bids starter repo will be integrated into the bids website so that template file us likely to move so I preferred to pin a specific commit to avoid problems.

nipoppy/workflows/dataset_init.py
Remi-Gau: have not forgotten about this: just forgot to push
will do as soon as I can restart my computer (currently without a charger for it)

michellewang (approved):
Looks good @Remi-Gau, thanks a lot! I will go ahead and merge this

PR #384, opened 2024-10-02
[WIP][ENH] pull the pipeline's container if it is not found

Remi-Gau:
- Closes #

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


nipoppy/workflows/pipeline.py
```diff
@@ -163,12 +164,86 @@ def fpath_container(self) -> Path:
...
+            docker_hub_repo = "nipreps"
+
+        command = (
+            f"{self.CONTAINER_CONFIG.COMMAND} build {self.fpath_container} "
+            f"docker://{docker_hub_repo}/{self.pipeline_name}:{self.pipeline_version}"
+        )
+
+        self.log_command(command)
+
+        def process_output(
```
Remi-Gau, line 203: copy pasted from the base workflow

may need to extract it into a separate method to be able to run it

nipoppy/workflows/pipeline.py
```diff
@@ -163,12 +164,86 @@ def fpath_container(self) -> Path:
...
+        )
+
+        if self.pipeline_name in ["mriqc", "fmriprep"]:
+            docker_hub_repo = "nipreps"
+        elif self.pipeline_name in ["dcm2bids"]:
+            docker_hub_repo = "nipreps"
+        elif self.pipeline_name in ["bidscoin"]:
+            docker_hub_repo = "nipreps"
+        elif self.pipeline_name in ["heudiconv"]:
+            docker_hub_repo = "nipreps"
```
michellewang, line 194: Wouldn't it be better to use the existing value in the global config file for this?
```python
self.pipeline_config.CONTAINER_INFO.URI
```

That way, this would work even for pipelines that do not come with Nipoppy by default.

Remi-Gau: Yeah I was planning to refactor things once I got the tests to pass.
But right now I am stuck on the red part of the "red - green - refactor" of test driven development.

nipoppy/workflows/pipeline.py
michellewang: Should this not just be a call to `self.run_command(command)`? Which will log the command and run it in a subprocess

https://github.com/nipoppy/nipoppy/blob/379ace12e40369af0655670a63553cad942a2d50/nipoppy/workflows/base.py#L95-L182

mathdugre:
Not sure if this can help, but boutiques as a cli to pull the container from a descriptor: [`bosh execute prepare`](https://boutiques.github.io/doc/_execute.html#execute-prepare)

Remi-Gau:
> Not sure if this can help, but boutiques as a cli to pull the container from a descriptor

not sure either: in case the user wants to use datalad then I'd want to do a datalad install of the image but in other cases why not.

need to check

michellewang:
> boutiques as a cli to pull the container from a descriptor: [bosh execute prepare](https://boutiques.github.io/doc/_execute.html#execute-prepare)

This wouldn't help with the current descriptors because they don't use Boutiques' built-in support for containers (Nipoppy injects the Singularity/Apptainer command directly into the descriptor's "command-line" field)

PR #420, opened 2024-11-01
[ENH] Allow BIDS datasets without session level

### LINKED ISSUE #171: Allow BIDS datasets without session level?
BIDS datasets do not require a session level (i.e. something like `bids/sub-1/anat` is a valid BIDS dataset), but Nipoppy is currently only compatible with datasets that have session directories. The runscripts and trackers could be modified so that the `session` argument/flag is optional, and handle both cases with/without sessions. The manifest/doughnut/bagels might also need to be a little different (?)

This would only apply to cases where users bring their own BIDS datasets, since if they start from DICOMs with Nipoppy then the BIDS dataset will always have session directories.

mtorabi59:
- Closes #171 

- datasets with no session definition will be allowed
- the manifest file will have a fake session name for participants with no session level when starting from a BIDS dataset
- the doughnut file won't exclude participants with no sessions

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/env.py
+++ b/nipoppy/env.py
@@ -9,7 +9,7 @@ StrOrPathLike = TypeVar("StrOrPathLike", str, os.PathLike)
 # BIDS
 BIDS_SUBJECT_PREFIX = "sub-"
 BIDS_SESSION_PREFIX = "ses-"
-FAKE_SESSION_ID = "IsFake"
+FAKE_SESSION_ID = "unnamed"
 
 # default config
 DEFAULT_PIPELINE_STEP_NAME = "default"
--- a/nipoppy/env.py
+++ b/nipoppy/env.py
@@ -9,7 +9,7 @@ StrOrPathLike = TypeVar("StrOrPathLike", str, os.PathLike)
 # BIDS
 BIDS_SUBJECT_PREFIX = "sub-"
 BIDS_SESSION_PREFIX = "ses-"
-FAKE_SESSION = "ses-IsFake"
+FAKE_SESSION_ID = "IsFake"
 
 # default config
 DEFAULT_PIPELINE_STEP_NAME = "default"
--- a/nipoppy/env.py
+++ b/nipoppy/env.py
@@ -9,6 +9,7 @@ StrOrPathLike = TypeVar("StrOrPathLike", str, os.PathLike)
 # BIDS
 BIDS_SUBJECT_PREFIX = "sub-"
 BIDS_SESSION_PREFIX = "ses-"
+FAKE_SESSION = "ses-IsFake"
 
 # default config
 DEFAULT_PIPELINE_STEP_NAME = "default"
--- a/nipoppy/tabular/doughnut.py
+++ b/nipoppy/tabular/doughnut.py
@@ -9,7 +9,7 @@ from typing import Optional
 from pydantic import Field
 from typing_extensions import Self
 
-from nipoppy.env import StrOrPathLike, FAKE_SESSION
+from nipoppy.env import FAKE_SESSION, StrOrPathLike
 from nipoppy.logger import get_logger
 from nipoppy.tabular.dicom_dir_map import DicomDirMap
 from nipoppy.tabular.manifest import Manifest, ManifestModel
--- a/nipoppy/tabular/doughnut.py
+++ b/nipoppy/tabular/doughnut.py
@@ -9,7 +9,7 @@ from typing import Optional
 from pydantic import Field
 from typing_extensions import Self
 
-from nipoppy.env import FAKE_SESSION, StrOrPathLike
+from nipoppy.env import FAKE_SESSION_ID, StrOrPathLike
 from nipoppy.logger import get_logger
 from nipoppy.tabular.dicom_dir_map import DicomDirMap
 from nipoppy.tabular.manifest import Manifest, ManifestModel
@@ -206,18 +206,16 @@ def generate_doughnut(
                 dpath=dpath_organized,
                 dname_subdirectory=Path(bids_participant_id, bids_session_id),
             )
-            if session_id == FAKE_SESSION:
+            if session_id == FAKE_SESSION_ID:
                 # if the session is fake, we don't expect BIDS data
                 # to have bids_session_id in the path
-                status_bidsified = check_status(
-                    dpath=dpath_bidsified,
-                    dname_subdirectory=Path(bids_participant_id),
-                )
+                dname_subdirectory = Path(bids_participant_id)
             else:
-                status_bidsified = check_status(
-                    dpath=dpath_bidsified,
-                    dname_subdirectory=Path(bids_participant_id, bids_session_id),
-                )
+                dname_subdirectory = Path(bids_participant_id, bids_session_id)
+            status_bidsified = check_status(
+                dpath=dpath_bidsified,
+                dname_subdirectory=dname_subdirectory,
+            )
 
         doughnut_records.append(
             {
--- a/nipoppy/tabular/doughnut.py
+++ b/nipoppy/tabular/doughnut.py
@@ -9,7 +9,7 @@ from typing import Optional
 from pydantic import Field
 from typing_extensions import Self
 
-from nipoppy.env import StrOrPathLike
+from nipoppy.env import StrOrPathLike, FAKE_SESSION
 from nipoppy.logger import get_logger
 from nipoppy.tabular.dicom_dir_map import DicomDirMap
 from nipoppy.tabular.manifest import Manifest, ManifestModel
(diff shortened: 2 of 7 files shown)


nipoppy/env.py
```diff
@@ -9,6 +9,7 @@
 # BIDS
 BIDS_SUBJECT_PREFIX = "sub-"
 BIDS_SESSION_PREFIX = "ses-"
+FAKE_SESSION = "ses-IsFake"
```
michellewang, line 12: @nikhil153 @alyssadai @surchs I think we decided to call made-up sessions `ses-unnamed` -- do we still want to go with that name or use something else?

nipoppy/tabular/doughnut.py
```diff
@@ -206,10 +206,18 @@ def check_status(
...
+                # to have bids_session_id in the path
+                status_bidsified = check_status(
+                    dpath=dpath_bidsified,
+                    dname_subdirectory=Path(bids_participant_id),
+                )
+            else:
+                status_bidsified = check_status(
+                    dpath=dpath_bidsified,
+                    dname_subdirectory=Path(bids_participant_id, bids_session_id),
+                )
```
michellewang, line 220: To minimize redundancy you could have:

```python
if session_id == FAKE_SESSION:
    dname_subdirectory = bids_participant_id
else:
    dname_subdirectory = Path(bids_participant_id, bids_session_id)
status_bidsified = check_status(
    dpath=dpath_bidsified,
    dname_subdirectory=dname_subdirectory,
)
```

nipoppy/workflows/dataset_init.py
```diff
@@ -154,20 +155,37 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
             )
-            if not session_ids:
-                self.logger.warning(
-                    f"Skipping subject '{ppt}': could not find a session level folder."
-                )
-                continue
+            if len(session_ids) == 0:
+                # if there are no session folders
+                # we will add a fake session for this participant
+                session_ids = [FAKE_SESSION]
```
michellewang, line 161: It would be good to write a warning to the log, something like `f"Could not find session-level folder(s) for participant {ppt}, using {FAKE_SESSION} in the manifest`"

nipoppy/env.py
michellewang, line 12: For consistency with the rest of the codebase and how we differentiate between session IDs with and without the BIDS prefix, I think this should be either `FAKE_SESSION_ID = "IsFake"` (or whatever name we decide to name this) or `FAKE_BIDS_SESSION_ID = "ses-IsFake"`. Would prefer `FAKE_SESSION_ID = "IsFake"` since most of the code doesn't use the BIDS prefix

nipoppy/workflows/dataset_init.py
```diff
@@ -154,20 +155,37 @@ def _init_manifest_from_bids_dataset(self) -> None:
                     if x.is_dir() and x.name.startswith(BIDS_SESSION_PREFIX)
                 ]
             )
-            if not session_ids:
-                self.logger.warning(
-                    f"Skipping subject '{ppt}': could not find a session level folder."
-                )
-                continue
+            if len(session_ids) == 0:
```
michellewang, line 158: I know you're not the one who came up with these variable names, but for clarity I think we should do some renaming:
- `session_ids` -> `bids_session_ids`
- `ses` -> `bids_session_id`
- `participant_ids` -> `bids_participant_ids`
- `ppt` -> `bids_participant_id`

Just to make it clear that these all have the BIDS prefixes to avoid confusion in the future

nipoppy/workflows/pipeline.py
```diff
@@ -315,7 +316,7 @@ def set_up_bids_db(
                 current=pybids_ignore_patterns,
                 new=f"^(?!/{BIDS_SUBJECT_PREFIX}({participant_id}))",
             )
-        if session_id is not None:
+        if (session_id is not None) and (session_id != FAKE_SESSION):
```
michellewang, line 319: This is not going to work because right now `session_id` doesn't have the BIDS prefix but `FAKE_SESSION` does. We can either remove the prefix from `FAKE_SESSION` (as I suggested previously) or use `nipoppy.utils.check_session_id(FAKE_SESSION)` to remove the prefix

nipoppy/workflows/dataset_init.py
mtorabi59: I agree!

nipoppy/workflows/dataset_init.py
```diff
@@ -146,49 +146,54 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
+                self.logger.warning(
+                    "Could not find session-level folder(s) for participant "
+                    f"{bids_participant_id}, using {FAKE_SESSION_ID} in the manifest"
+                )
+                bids_session_ids = [f"{BIDS_SESSION_PREFIX}{FAKE_SESSION_ID}"]
 
-            for ses in session_ids:
-                if ses == FAKE_SESSION:
+            for bids_session_id in bids_session_ids:
+                if bids_session_id == f"{BIDS_SESSION_PREFIX}{FAKE_SESSION_ID}":
```
michellewang, line 168: Can you `nipoppy.utils.participant_id_to_bids_participant_id` here:

```python
from nipoppy.utils import participant_id_to_bids_participant_id
    if bids_session_id == f"{participant_id_to_bids_participant_id(FAKE_SESSION_ID)}":
```

nipoppy/workflows/dataset_init.py
```diff
@@ -146,49 +146,54 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
                 ]
             )
-            if len(session_ids) == 0:
+            if len(bids_session_ids) == 0:
                 # if there are no session folders
                 # we will add a fake session for this participant
-                session_ids = [FAKE_SESSION]
+                self.logger.warning(
+                    "Could not find session-level folder(s) for participant "
+                    f"{bids_participant_id}, using {FAKE_SESSION_ID} in the manifest"
```
michellewang, line 163:
```suggestion
                    f"{bids_participant_id}, using session {FAKE_SESSION_ID} in the manifest"
```

Since `FAKE_SESSION_ID` has no BIDS prefix now

nipoppy/workflows/dataset_init.py
```diff
@@ -146,49 +146,54 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
                             if x.is_dir()
                         ]
                     )
 
-                # if there are no datatypes, raise warning and skip
-                if len(datatypes) == 0:
-                    self.logger.warning(
-                        f"Participant {ppt}, session {ses} has no datatypes. Skipping."
-                    )
-                    continue
```
michellewang: I forgot what we said about this -- did we still want a warning if there is no datatype?

tests/test_workflow_pipeline.py
```diff
@@ -507,6 +514,34 @@ def test_set_up_bids_db_ignore_patterns(workflow: PipelineWorkflow, tmp_path: Pa
...
+    """
+    dpath_pybids_db = tmp_path / "bids_db"
+    participant_id = "01"
+    session_id = FAKE_SESSION_ID
+
+    fids.create_fake_bids_dataset(
+        output_dir=workflow.layout.dpath_bids,
+        subjects=participant_id,
+        sessions=session_id,
+    )
```
michellewang, line 534: Wouldn't this create a BIDS dataset **with** `ses-{FAKE_SESSION_ID}` folders? Shouldn't it be this instead:

```python
    fids.create_fake_bids_dataset(
        output_dir=workflow.layout.dpath_bids,
        subjects=participant_id,
    )
```

michellewang: It's probably also better to put some data (e.g. anat) in the BIDS dataset, then check that the BIDS layout object returned by `set_up_bids_db` is not empty, e.g. 
```python
    bids_layout = workflow.set_up_bids_db(
        dpath_pybids_db=dpath_pybids_db,
        participant_id=participant_id,
        session_id=session_id,
    )
    assert len(bids_layout.get(extension=".nii.gz")) > 0
```

nipoppy/env.py
```diff
@@ -9,7 +9,7 @@
 # BIDS
 BIDS_SUBJECT_PREFIX = "sub-"
 BIDS_SESSION_PREFIX = "ses-"
-FAKE_SESSION = "ses-IsFake"
+FAKE_SESSION_ID = "IsFake"
```
michellewang, line 12:
```suggestion
FAKE_SESSION_ID = "unnamed"
```

I checked with the Neurobagel team and @nikhil153 and they prefer `unnamed` over `IsFake`

michellewang (commented):
Looks really good @mtorabi59! Just some minor comments remaining, should be quick to address :)

michellewang (approved):

PR #425, opened 2024-11-06
[MAINT] Fix `PerformanceWarning` in `BaseTabular` class

### LINKED ISSUE #424: Improve performance of `BaseTabular.add_or_update_records`
Currently we get a performance warning sometimes:
```
nipoppy/nipoppy/tabular/base.py:202: PerformanceWarning: indexing past lexsort depth may impact performance.
    self.loc[tuple(record[col] for col in self.index_cols), col] = record[
```

This is because the index isn't sorted before the `.loc[]` operation. Can be fixed but requires another fix in the `__init__` so that existing columns are not overwritten.

Also: the `reset_index` part should be in a `finally` clause since the df is modified inplace and we don't want to keep the modified index if the function crashes. Same thing for `Doughnut.set_status`.

michellewang:
- Closes #424

- Sort the index before using `.loc`
- Various fixes:
  - Do not set columns in `__init__` if there are already columns
  - Make `add_or_update_records` work even if `index_cols` has length 1
  - Add tests for `add_or_update_records`
  - Put `reset_index` in `finally` clause
     - Do this for `Doughnut.set_status` too
  - Make `mypy` happy

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


mathdugre (approved):
🧑‍🍳

PR #444, opened 2024-11-28
[MAINT] Refactor CLI to use Click

mathdugre:
- Use Click instead of Argparse to simplify the code for creating sub-command or sub-sub-command

## Checklist (for reviewers)

- [ ] ~~PR links to GitHub issue with mention `Closes #XXXX`~~
- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


mathdugre:
@michellewang , let me know if you have ideas to fix the codecov. Otherwise, I'll look more into it on my side.

tests/test_cli.py
```diff
@@ -1,152 +1,218 @@
...
+                )
+            )
+            == 1
+        )
+
+        # Dataset was not initialized properly
+        assert excinfo.value.startswtih(
+            "Dataset does not follow expected directory structure:"
+        )
+        assert result.exit_code == 1
```
michellewang, line 78: I don't think these lines are being run at all because the exception is thrown (e.g. there is a typo in `startswtih` but the test still passes). So we would need to de-indent some of this. But that still doesn't fix the coverage things though I have an idea for that, which would also address another comment I had about exceptions being written to the logger.

In `cli.py` we could have something like that for the subcommands:
```python
from nipoppy.env import ReturnCode

def doughnut(**params):
    """Command: nipoppy doughnut."""
    from nipoppy.workflows.doughnut import DoughnutWorkflow

    workflow = DoughnutWorkflow(**params)
    try:
        workflow.run()
    except Exception:
        workflow.logger.exception("Error while running nipoppy")
        if workflow.return_code == ReturnCode.SUCCESS:
            workflow.return_code = 2  # we could make a ReturnCode.OTHER_FAILURE for that

    sys.exit(workflow.return_code)
```

Then in `test_cli.py` the test could be like:
```python
def test_cli_doughnut(tmp_path: Path):
    dpath_root = tmp_path / "my_dataset"

    result = runner.invoke(cli, ["doughnut", "--dataset", str(dpath_root)])

    # check that a logfile was created
    assert (
        len(list((dpath_root / ATTR_TO_DPATH_MAP["dpath_logs"]).glob("doughnut/*.log")))
        == 1
    )
    assert result.exit_code == 2  # or ReturnCode.OTHER_FAILURE if we decide to do that
```

There is still a remaining thing that I think is suboptimal compared to what I had before: the logger won't catch errors that occur during workflow instantiation. For example, if I run `nipoppy init --dataset . -vvvv` (too many verbosity counts) then I get an ugly traceback with all the `click`-related frames. Could be sort-of "fixed" by not doing anything other than variable assignment in `__init__()`. But that is probably why I was passing an already-instantiated logger to the workflow classes instead of letting them make their own

tests/test_cli.py
```diff
@@ -1,152 +1,218 @@
...
+def test_cli_init_dir_exists(tmp_path: Path):
+    dpath_root = tmp_path / "my_dataset"
+    dpath_root.joinpath("dir_exists").mkdir(parents=True, exist_ok=True)
+
+    with pytest.raises(FileExistsError):
+        runner.invoke(
+            cli,
+            ["init", "--dataset", str(dpath_root)],
+            catch_exceptions=False,
+        )
```
michellewang, line 49: I think this should be in `test_workflow_init.py` instead of the CLI tests

nipoppy/cli.py
```diff
@@ -0,0 +1,213 @@
...
+        ),
+    )(func)
+    func = click.option(
+        "--verbose",
+        "-v",
+        count=True,
+        default=DEFAULT_VERBOSITY,
+        help="Verbosity level, Verbosity level, from 0 (least verbose) "
+        f"to 3 (most verbose). Default: {DEFAULT_VERBOSITY}.",
+    )(func)
```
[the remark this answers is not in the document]
mathdugre: I modified the help text to use the same as fMRIPrep verbose option. When no verbosity is passed, it will now default to 0.

tests/test_cli.py
```diff
@@ -1,152 +1,218 @@
 """Tests for the CLI."""
 
 from pathlib import Path
+from unittest import mock
```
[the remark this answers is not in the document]
mathdugre: I'm not quite sure of the difference either. We can look at it in another issue.

tests/test_workflow_runner.py
```diff
@@ -95,7 +95,7 @@ def test_run_cleanup(tmp_path: Path, keep_workdir):
     dpaths = [runner.dpath_pipeline_bids_db, runner.dpath_pipeline_work]
     for dpath in dpaths:
         dpath.mkdir(parents=True)
-    runner.run_cleanup()
+        runner.run_cleanup()
```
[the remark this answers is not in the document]
mathdugre: Indeed, good catch !

nipoppy/workflows/base.py
```diff
@@ -192,6 +201,10 @@ def save_tabular_file(self, tabular: BaseTabular, fpath: Path):
 
     def run_setup(self):
         """Run the setup part of the workflow."""
+        add_logfile(self.logger, self.generate_fpath_log())
```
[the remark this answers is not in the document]
mathdugre: Add a `skip_logging` to the base class. For now, it's only skipped for the `nipoppy init` command.

pyproject.toml
```diff
@@ -1,29 +1,30 @@
...
     "pandas",
     "pybids!=0.18.0",
     "pydantic",
-    'requests',
     "pydicom!=3.0.0",
     "rich",
-    "rich_argparse",
+    "rich-click",
     "typing-extensions",
+    'requests',
```
michellewang, line 27: `requests` is not needed (yet)?

tests/test_cli.py
```diff
@@ -2,30 +2,50 @@
...
+def test_cli_init_dir_exists(tmp_path: Path):
+    dpath_root = tmp_path / "my_dataset"
+    dpath_root.joinpath("dir_exists").mkdir(parents=True, exist_ok=True)
+
+    result = runner.invoke(
+        cli,
+        ["init", "--dataset", str(dpath_root)],
+    )
+
+    assert result.exit_code == ReturnCode.UNKOWN_FAILURE
```
michellewang, line 48: Is this not already tested in `test_workflow_init.py`?

michellewang (commented):
Made a few additional comments, some of which might be outdated though

docs/source/user_guide/tracking.md
```diff
@@ -41,15 +41,15 @@ Here is example of tracker configuration file (default for MRIQC 23.1.0):
 Importantly, pipeline completion status is **not** inferred from exit codes as trackers are run independently of the pipeline runners. Moreover, the default tracker configuration files are somewhat minimal and do not check all possible output files generated these pipelines.
 
 ```{tip}
-- The paths are expected to be relative to the {{dpath_derivatives}}`/<PIPELINE_NAME>/<PIPELINE_VERSION>/output` directory.
+- The paths are expected to be relative to the {{dpath_pipeline_output}}`/<PIPELINE_NAME>/<PIPELINE_VERSION>/output` directory.
```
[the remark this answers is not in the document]
michellewang: So this is outdated, and I can't make a comment in the `rst` file but there is the same change that should be made

nipoppy/cli.py
```diff
@@ -1,22 +1,50 @@
...
+
+
+def global_options(func):
+    """Define global options for the CLI."""
+    func = click.option(
+        "--verbose",
+        "-v",
+        count=True,
+        help="Increases log verbosity for each occurrence, debug level is -vvv",
```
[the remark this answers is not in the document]
mathdugre: As of right now, the default would print errors. 
The `-v` would print warning, `-vv` info, and `-vvv` debug.


Should we set the default to `-v` then, to print the warning?

nipoppy/cli.py
```diff
@@ -188,22 +216,39 @@ def bidsify(**params):
...
+
+@cli.command()
+@dataset_option
+@pipeline_options
+@click.option(
+    "--simulate",
+    is_flag=True,
+    help="Simulate the pipeline run without executing the generated command-line.",
+)
+@global_options
```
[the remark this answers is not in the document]
mathdugre: Would it be for the different pipeline "runners"?

nipoppy/cli.py
```diff
@@ -0,0 +1,254 @@
...
+    help=(
+        "Keep pipeline working directory upon success "
+        "(default: working directory deleted unless a run failed)"
+    ),
+)
+@click.option(
+    "--simulate",
+    is_flag=True,
+    help="Simulate the pipeline run without executing the generated command-line.",
+)
```
michellewang, line 218: Just noticed that we are also missing the `--tar` flag here: https://github.com/nipoppy/nipoppy/blob/75581ba6a547811737eabcc5d585acfa6ab2213d/nipoppy/cli/parser.py#L310-L318

pyproject.toml
mathdugre: The diff was from pre-commit. We use `requests` [here](https://github.com/nipoppy/nipoppy/blob/75581ba6a547811737eabcc5d585acfa6ab2213d/nipoppy/workflows/dataset_init.py#L124) apparently

tests/test_cli.py
mathdugre: I don't see a test for it right now. Could you point to the current one?

pyproject.toml
michellewang: Oh right, my bad. 

Why is it single quotes though? 😢

tests/test_cli.py
michellewang: I think here:
https://github.com/nipoppy/nipoppy/blob/75581ba6a547811737eabcc5d585acfa6ab2213d/tests/test_workflow_init.py#L56-L60

pyproject.toml
mathdugre: Not sure, I'll put in back in double quotes.

tests/test_cli.py
mathdugre: Ah I see. I'll remove the other one

nipoppy/cli.py
```diff
@@ -0,0 +1,259 @@
...
+    with handle_exception(PipelineTracker(**params)) as workflow:
+        workflow.run()
+
+
+@cli.command()
+@dataset_option
+@runners_options
+@global_options
+def extract(**params):
+    """Add subparser for extract command."""
```
michellewang, line 255:
```suggestion
    """Run an extraction pipeline."""
```
or something equally/more informative

mathdugre:
After discussing with @michellewang , we want to maintain the current way of passing `<dataset_path>` to the CLI, while also having the option to pass `--dataset <dataset_path>`. The former should be marked as deprecated.

nipoppy/cli.py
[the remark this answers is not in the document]
michellewang: I think since we are making this PR non-breaking we should revert to previous verbosity behaviour. Then in a future PR (or series of PRs) we can re-address default verbosity level and also make logs less verbose in general

michellewang:
@mathdugre I merged #462 which adds another CLI arg so there is a new merge conflict :/

michellewang (approved):
:cook: looks great @mathdugre!

PR #462, opened 2025-01-09
[ENH] Add flag for writing HPC participant-session list

### LINKED ISSUE #435: [ENH] Generate list of participants to be submitted for HPC job

### New feature

A back-up / manual option for submitting available (and incomplete processing) participants to HPC job. 
This will involve changing the [BasePipelineWorkflow](https://github.com/nipoppy/nipoppy/blob/main/nipoppy/workflows/pipeline.py)
- [ ] Checking bagel for identifying participants that are yet to be processed per session per pipeline
- [ ] Write separate lists of these participants on disk

### Unclear documentation

_No response_

michellewang:
- Closes #435

- Add flag `--write-list` to `run`, `bidsify`, and `extract` subcommands, so that users can run e.g. `nipoppy run ... --write-list my_list.tsv` to write the list of participants/sessions to run

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


nipoppy/cli/parser.py
```diff
@@ -148,6 +148,16 @@ def add_args_runner(parser: _ActionsContainer) -> _ActionsContainer:
         action="store_true",
         help="Simulate the pipeline run without executing the generated command-line.",
     )
+    parser.add_argument(
+        "--write-list",
```
nikhil153, line 152: do you think `write_job_list` or `write_job_table` (since we have two columns) is a better flag? Just a thought, no strong preference on my side.

nipoppy/workflows/pipeline.py
```diff
@@ -412,21 +415,29 @@ def run_main(self):
...
-            try:
-                self.run_single(participant_id, session_id)
-                self.n_success += 1
-            except Exception as exception:
-                self.return_code = ReturnCode.PARTIAL_SUCCESS
-                self.logger.error(
-                    f"Error running {self.pipeline_name} {self.pipeline_version}"
-                    f" on participant {participant_id}, session {session_id}"
-                    f": {exception}"
+        # TODO mutually exclusive with HPC option
```
nikhil153, line 418: Not sure if this needs to be mutually exclusive. It won't hurt to have this list as back-up, in case something goes wrong with HPC submission. 
If you agree, then we can also make this a default behavior, so there is no extra flag for CLI.

nipoppy/cli/parser.py
michellewang: I wouldn't call it `job-list` because I think it's confusing to call it a "job" if no job is being submitted. My initial thought was `--write-participant-session-list` but that is really long to type...

nipoppy/workflows/pipeline.py
michellewang: If something goes wrong with HPC submission I think it would make more sense for the user to try to debug/troubleshoot the problem. I am thinking this `--write-list` is more for users who have an HPC but not Slurm or SGE, or who don't have an HPC at all and want to parallelize on their own.

We don't have to make this mutually exclusive with the `--hpc` option (still, to be discussed I think) but I don't think this should be the default behaviour (also, if it is default behaviour we would have to auto-generate the path to write)

nikhil153: >If something goes wrong with HPC submission I think it would make more sense for the user to try to debug/troubleshoot the problem. 
There several things that can go wrong with nipoppy - hpc interface... I doubt users will spend time debugging all of those, especially if they have a existing way of submitting jobs that they are comfortable with. 
I don't see the downside of generating these lists by default. We can dump them in the same directory we are putting hpc configs.

nikhil153 (approved):
To me writing the list to disk by default still makes more sense, but happy to wait until actual users ask for it.

PR #466, opened 2025-01-10
[ENH] Make trackers handle tarball in pipeline output

### LINKED ISSUE #465: [ENH] Make trackers able to check paths inside tarballs

### New feature

Following #430, we now have pipeline outputs that are tarred if the exit code was 0 but not tarred if not. Currently the trackers cannot handle both cases.

### Unclear documentation

_No response_

michellewang:
- Closes #465

- If the tracker config file specifies a `PARTICIPANT_SESSION_DIR` value, and there is a tarball there (with `.tar` extension), then the paths inside that tarball are collected. If a tracker path cannot be found in the regular filesystem, then the tarball files are checked. Only if a path cannot be found in both cases would the tracker return `FAIL` status
   - This does slow down things quite a bit if the output has tarballs
- Note that the current tarball file-search implementation is more permissive than regular filesystem globbing... For example if the tracker config has path pattern `*file.txt`, if we check untarred files `dirA/file.txt` will not match, but if `dirA` here is tarred then this will match. Python does have a [built-in glob-like match](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.full_match) for strings (not checking the filesystem), but that is only introduced in Python 3.13. The solution would be to manually implement this glob-like matching functions (using `fnmatch` but taking OS filesystem separators into account), not sure if it would be worth it though...
- Tested locally and seems to work!

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:


tests/test_workflow_tracker.py
```diff
@@ -115,19 +117,64 @@ def test_run_setup_existing_bad_bagel(
 @pytest.mark.parametrize(
     "relative_paths,expected_status",
     [
-        (["01_ses-1.txt", "file.txt"], Bagel.status_success),
-        (["01_ses-1.txt", "file.txt", "missing.txt"], Bagel.status_fail),
+        (["dirA/01_ses-1.txt", "dirA/dirB/file.txt"], Bagel.status_success),
+        (["**/*.txt"], Bagel.status_success),
+        (["*file.txt"], Bagel.status_fail),
```
michellewang, line 122: Add test cases with wildcards/glob patterns

bcmcpher (approved):
It would be good to add a default archive path as an example in the default configs.

I was able to pull this down and also got it working. I agree there are more robust checks that would be beneficial here, but I think user feedback can better guide the priority on them.

PR #503, opened 2025-02-14
[ENH] Remove pipeline configs from global config

michellewang:
- Closes none but relates to #316 

Changes proposed in this pull request:
- Move pipeline configs out of global config and into the `<DATASET_ROOT>/pipelines/{bids,proc,extraction}/config.json`

## Checklist (for reviewers)
_This section is for the PR reviewer_

- [x] PR has an interpretable title with a prefix (e.g. `[BUG]`, `[DOC]`, `[ENH]`, `[MAINT]`)\
Refer to [NumPy Development Guide](https://numpy.org/doc/stable/dev/development_workflow.html#writing-the-commit-message) for a full list
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:
- [ ] There is at least one test that would fail under the original bug conditions

Issue #446, opened 2024-11-28
[MAINT] Make a Nipoppy bot


michellewang:
Neurobagel now has a nice bot (and [notes](https://neurobagel.github.io/dev-docs/sections/Neurobagel%20PATs/#authenticating-using-the-neurobagel-bot-app) on it), we should have that too for project automations/CI workflows

PR #506, opened 2025-02-20
[ENH] api to download and upload pipelines from zenodo.

mathdugre:
- Closes #

- 

## Checklist (for reviewers)

- [x] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


mathdugre:
We will need to create a `ZENODO_TOKEN` secret for the test to pass. It won't be used for now, but will be required once we have the upload functions.

nipoppy/zenodo.py
```diff
@@ -0,0 +1,70 @@
...
+        response = httpx.get(
+            f"{self.api_endpoint}/records/{zenodo_id}/files",
+            headers=self.headers,
+        )
+        if response.status_code != 200:
+            raise Exception(response.json())
+
+        # Exclude "md5:" prefix
+        files = {
+            entry["key"]: entry["checksum"][4:] for entry in response.json()["entries"]
```
michellewang, line 54: I think using `removeprefix()` is cleaner:
```suggestion
            entry["key"]: entry["checksum"].removeprefix("md5:") for entry in response.json()["entries"]
```

nipoppy/zenodo.py
```diff
@@ -0,0 +1,70 @@
...
+        InvalidChecksumError
+            Checksum mismatch between the downloaded file and the expected checksum.
+        """
+        output_dir.mkdir(parents=True, exist_ok=True)
+        response = httpx.get(
+            f"{self.api_endpoint}/records/{zenodo_id}/files",
+            headers=self.headers,
+        )
+        if response.status_code != 200:
+            raise Exception(response.json())
```
michellewang, line 50: We should probably have a better exception and error message

nipoppy/zenodo.py
```diff
@@ -0,0 +1,70 @@
...
+        files = {
+            entry["key"]: entry["checksum"][4:] for entry in response.json()["entries"]
+        }
+        for filename, checksum in files.items():
+            response = httpx.get(
+                f"{self.api_endpoint}/records/{zenodo_id}/files/{filename}/content",  # noqa E501
+                headers=self.headers,
+            )
+            if response.status_code != 200:
+                raise Exception(response.json())
```
michellewang, line 62: Same comment as above re: exception type and error message.

Would it be worth having a helper function for the HTTP requests that also checks the status code?

tests/test_zenodo.py
```diff
@@ -0,0 +1,38 @@
...
+
+    zenodo.download_record_files(zenodo_id, tmp_path)
+
+    assert len(list(tmp_path.iterdir())) == 2
+
+    with tmp_path.joinpath("zenodo-test.txt").open() as f:
+        assert f.read() == "test 123"
+
+    with tmp_path.joinpath("test2.txt").open() as f:
+        assert f.read() == "test2"
```
michellewang, line 38: I think with `Path` we can do something like this instead:
```suggestion
    assert tmp_path.joinpath("zenodo-test.txt").read_text() == "test 123"
    assert tmp_path.joinpath("test2.txt").read_text() == "test2"
```

We could also use the `/` operator instead of `joinpath` to reduce line length

michellewang (changes requested):
We might want to include the Zenodo test record files in this repo so that when the sandbox gets wiped we can easily recreate the record

We should also do the GitHub secret upload for the test to work. Maybe this would be a good time to create the Nipoppy bot account #446...

nipoppy/cli.py
```diff
@@ -343,3 +344,54 @@ def status(**params):
...
+@cli.group(cls=OrderedGroup)
+def pipeline():
+    """Local pipeline management."""
+    pass
+
+
+def zenodo_options(func):
+    """Define global options for the CLI."""
+    func = click.option(
+        "--zenodo-token",
```
mathdugre, line 358: Not good, this will leak the password

nipoppy/cli.py
```diff
@@ -343,3 +344,54 @@ def status(**params):
...
+        envvar="ZENODO_TOKEN",
+        type=str,
+        required=False,
+        help="Zenodo access token.",
+    )(func)
+    func = click.option(
+        "--zenodo-api",
+        envvar="ZENODO_API",
+        default="https://sandbox.zenodo.org/api",
```
michellewang, line 367: I assume we would want to change the default to be non-sandbox Zenodo eventually, right?

nipoppy/cli.py
michellewang: Do you mean because it will be in the command-line history?

Also, this is the same as what Boutiques does, no?

nipoppy/cli.py
```diff
@@ -343,3 +344,54 @@ def status(**params):
...
+
+@pipeline.command("add")
+@click.argument(
+    "zenodo_id",
+    type=str,
+)
+@dataset_option
+@zenodo_options
+def pipeline_add(**params):
+    """Add a new pipeline to the configuration."""
```
michellewang, line 383:
```suggestion
    """Add a new pipeline."""
```
Since we are not updating the global config (yet, to discuss)

nipoppy/cli.py
```diff
@@ -343,3 +344,54 @@ def status(**params):
...
+        access_token=params.get("zenodo_token"),
+    )
+
+    layout = DatasetLayout(dpath_root=params.get("dpath_root"))
+
+    zenodo_id = params.pop("zenodo_id")
+    download_dir = "zenodo." + zenodo_id.removeprefix("zenodo.")
+    zenodo.download_record_files(
+        zenodo_id, layout.dpath_pipelines.joinpath(download_dir)
+    )
```
michellewang, line 397: I think we should make a workflow class for this like the other CLI commands, thoughts?

nipoppy/zenodo.py
```diff
@@ -41,30 +52,38 @@ def download_record_files(self, zenodo_id: str, output_dir: Path):
...
+                    f"Failed to download file for zenodo.{zenodo_id}: {filename}"
+                )
 
             # Checksum verification before writing to disk
             content_md5 = hashlib.md5(response.content).hexdigest()
             if content_md5 != checksum:
-                raise InvalidChecksumError(f"Checksum mismatch: {filename}")
+                raise InvalidChecksumError(
+                    "Checksum mismatch: "
+                    f"{filename} has invalid checksum ({content_md5})"
```
michellewang, line 85:
```suggestion
                    f"{filename} has invalid checksum {content_md5}, expected {checksum}"
```

nipoppy/zenodo.py
```diff
@@ -41,30 +52,38 @@ def download_record_files(self, zenodo_id: str, output_dir: Path):
...
+        zenodo_id = zenodo_id.removeprefix("zenodo.")
         output_dir.mkdir(parents=True, exist_ok=True)
+
         response = httpx.get(
             f"{self.api_endpoint}/records/{zenodo_id}/files",
             headers=self.headers,
         )
         if response.status_code != 200:
-            raise Exception(response.json())
+            raise ZenodoAPIError(f"Failed to get files for record: {zenodo_id}")
```
michellewang, line 63: Would it add useful information if we print out `response.json()` like what you had before?

nipoppy/zenodo.py
```diff
@@ -41,30 +52,38 @@ def download_record_files(self, zenodo_id: str, output_dir: Path):
...
         }
         for filename, checksum in files.items():
             response = httpx.get(
                 f"{self.api_endpoint}/records/{zenodo_id}/files/{filename}/content",  # noqa E501
                 headers=self.headers,
             )
             if response.status_code != 200:
-                raise Exception(response.json())
+                raise ZenodoAPIError(
+                    f"Failed to download file for zenodo.{zenodo_id}: {filename}"
```
michellewang, line 77: Same comment as above re: `response.json()`

nipoppy/zenodo.py
```diff
@@ -41,30 +52,38 @@ def download_record_files(self, zenodo_id: str, output_dir: Path):
...
             content_md5 = hashlib.md5(response.content).hexdigest()
             if content_md5 != checksum:
-                raise InvalidChecksumError(f"Checksum mismatch: {filename}")
+                raise InvalidChecksumError(
+                    "Checksum mismatch: "
+                    f"{filename} has invalid checksum ({content_md5})"
+                )
 
             with output_dir.joinpath(filename).open("wb") as f:
                 f.write(response.content)
```
michellewang, line 89: Not 100% this will work but we could try this here:
```suggestion
            output_dir.joinpath(filename).write_bytes(response.content)
```

nipoppy/zenodo.py
```diff
@@ -12,19 +13,29 @@ class InvalidChecksumError(Exception):
...
-        self.headers = {
-            "Authorization": f"Bearer {self.access_token}",
-        }
+
+        if self.access_token is not None:
+            self.headers = {
+                "Authorization": f"Bearer {self.access_token}",
+            }
+        else:
+            self.headers = {}
```
michellewang, line 38: We can download public records without an access token?

tests/test_zenodo.py
```diff
@@ -21,18 +20,23 @@ def zenodo_id():
...
+    assert tmp_path.joinpath("test2.txt").read_text() == "test2"
 
-    with tmp_path.joinpath("test2.txt").open() as f:
-        assert f.read() == "test2"
+
+def test_download_invalid_record(tmp_path: Path):
+    """Test for downloading an invalid pipeline from Zenodo."""
+    zenodo = ZenodoAPI(api_endpoint=API_ENDPOINT)
+
+    with pytest.raises(ZenodoAPIError):
```
michellewang, line 41: We could make this check more explicit with `match=`

tests/test_zenodo.py
```diff
@@ -21,18 +20,23 @@ def zenodo_id():
     return "170588"
```
michellewang, line 20: Should we put the two test files somewhere (in this repo or other) for easier regeneration of the sandbox record?

michellewang (changes requested):

michellewang:
@mathdugre I changed the dev branch and now the diff is big, but I think that will be fixed once we merge #503

nipoppy/zenodo.py
mathdugre: Yes, only restricted records require an access token

nipoppy/cli.py
mathdugre: I'll have to double check what boutiques does.

Yes, the issue is with having it in the command history and as plain text on cli (e.g. during demo or tutorial).
Using env var or prompt once and save to a config file would be better

nipoppy/cli.py
mathdugre: Would the workflow class be for the whole API? Or one for each function --- download, upload, etc.?

michellewang: Haven't thought about it deeply but so far we have had one for each function. But we should discuss if/how we would want to deviate from the current way of doing things for the `pipeline` subcommands

mathdugre:
We probably want to discuss the exact format we expect to or the files on zenodo. For example, do we want zip files,? Otherwise we would lose the file hierarchy

michellewang:
I was thinking and there could be times where some of the pipeline files are dataset-specific (e.g. PyBIDS ignore patterns list, and HPC config file in the future). For these files do we want to upload them as-is to Zenodo or maybe use default files?

nipoppy/data/examples/sample_pipelines/proc/fmriprep-24.1.1/zenodo.json
```diff
@@ -0,0 +1,13 @@
+{
+    "title": "Upload test",
+    "description": "This is a test upload",
```
michellewang, line 3: For `title` and `description` we can infer them from the pipeline config information

nipoppy/zenodo.py
```diff
@@ -0,0 +1,253 @@
...
+    pass
+
+
+class ZenodoAPIError(Exception):
+    """Error when interacting with Zenodo API."""
+
+    pass
+
+
+class ZenodoAPI:
```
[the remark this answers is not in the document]
mathdugre: I would keep it as is since API is uppercase by itself.

nipoppy/cli.py
```diff
@@ -343,3 +344,81 @@ def status(**params):
...
+        required=False,
+        help="Zenodo access token.",
+    )(func)
+    func = click.option(
+        "--zenodo-api",
+        envvar="ZENODO_API",
+        default="https://zenodo.org/api",
+        type=str,
+        required=False,
+        help="Zenodo access token.",
```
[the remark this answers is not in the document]
mathdugre: Oops, good catch!

nipoppy/cli.py
```diff
@@ -343,3 +344,81 @@ def status(**params):
...
+    )(func)
```
mathdugre, line 371: @michellewang, do you think it would be better to have a `--sandbox` flag instead? It might be more intuitive for the users.

mathdugre:
> I was thinking and there could be times where some of the pipeline files are dataset-specific (e.g. PyBIDS ignore patterns list, and HPC config file in the future). For these files do we want to upload them as-is to Zenodo or maybe use default files?

I think having some default files, which remain reusable by other dataset, would be good. Similar to the invocation files

nipoppy/zenodo.py
```diff
@@ -0,0 +1,253 @@
...
+                )
+
+            # Checksum verification before writing to disk
+            content_md5 = hashlib.md5(response.content).hexdigest()
+            if content_md5 != checksum:
+                raise InvalidChecksumError(
+                    "Checksum mismatch: " f"{file} has invalid checksum ({content_md5})"
+                )
+
+            output_dir.joinpath(file).write_bytes(response.content)
```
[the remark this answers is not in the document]
mathdugre: Not sure I understand this comment. Could you clarify want you mean?

nipoppy/zenodo.py
```diff
@@ -0,0 +1,253 @@
...
+        return response.json()["links"]["self_doi"]
+
+    def _get_pipeline_metadata(self, pipeline_dir: Path) -> dict:
+        # TODO Get metadata from the pipeline configuration file.
+        metadata = {
+            "metadata": {
+                "publication_date": get_today(),
+                "publisher": "Nipoppy",
+                "resource_type": {"id": "software"},
+                "keywords": ["Nipoppy"],
```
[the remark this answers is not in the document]
mathdugre: I agree that it's not necessary

nipoppy/zenodo.py
```diff
@@ -0,0 +1,253 @@
...
+            }
+        }
+
+        pipeline_metadata = json.loads(pipeline_dir.joinpath("zenodo.json").read_text())
```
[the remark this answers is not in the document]
mathdugre: One the pipeline validation is merged, this will be checked before doing anything else for the upload.

michellewang:
> > I was thinking and there could be times where some of the pipeline files are dataset-specific (e.g. PyBIDS ignore patterns list, and HPC config file in the future). For these files do we want to upload them as-is to Zenodo or maybe use default files?
>
> I think having some default files, which remain reusable by other dataset, would be good. Similar to the invocation files

Yes, though what I am thinking of is how to enforce this. For example do we want to prevent invocation files that have absolute paths in them? Similar for HPC config (Slurm `--account` etc.). But that might be overkill (and difficult to do)

nipoppy/data/examples/sample_pipelines/proc/fmriprep-24.1.1/zenodo.json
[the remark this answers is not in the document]
mathdugre: When creating a record or updating an existing one, we need to pass the full metadata. This allows the pipeline developer to include the required metadata which we cannot infer.

nipoppy/data/examples/sample_pipelines/proc/fmriprep-24.1.1/zenodo.json
mathdugre: Could you point me to the location of the description we could reuse? I'm not sure to know which one you are refering to.

michellewang: I was thinking just something generic like:
```json
    "title": f"{pipeline_config.NAME} {pipeline_config.VERSION}",
    "description": f"Nipoppy configuration files for {pipeline_config.NAME} {pipeline_config.VERSION} pipeline"
```

nipoppy/data/examples/sample_pipelines/processing/fmriprep-24.1.1/zenodo.json
mathdugre, line 1: Check file exist before upload. Validation doesn't do it

mathdugre: We can use defaults
- Title: from the config "{PIPELINE_NAME}-{PIPELINE_VERSION}".
- Description: from descriptor description
- Creators: fetch creator's info from zenodo, since they already need an account with their token

nipoppy/cli.py
```diff
@@ -352,6 +355,77 @@ def pipeline():
...
+)
+@click.option(
+    "--force",
+    "-f",
+    is_flag=True,
+    help="Force download and overwrite existing files.",
+)
+@dataset_option
+@zenodo_options
+def pipeline_download(**params):
```
michellewang, line 390: Once we have `nipoppy pipeline install`, do we still want to keep `nipoppy pipeline download` or would we merge the Zenodo downloading part into the install workflow? I can't think of a good scenario where users would want to download without installing

nipoppy/workflows/zenodo.py
```diff
@@ -0,0 +1,155 @@
...
+        super().__init__(
+            dpath_root=Path(self.dpath_pipeline).parent.parent,
+            name="pipeline_upload",
+            fpath_layout=fpath_layout,
+            verbose=verbose,
+            dry_run=dry_run,
+        )
+
+    def _get_pipeline_metadata(
+        self, zenodo_metadata: Path, pipeline_config: Path
```
michellewang, line 39: Can we make it more explicit in the variable name that `zenodo_metadata` is a path to a file? 😅

nipoppy/workflows/zenodo.py
```diff
@@ -0,0 +1,155 @@
...
+            dpath_root=dpath_root,
+            name="pipeline_download",
+            fpath_layout=fpath_layout,
+            verbose=verbose,
+            dry_run=dry_run,
+        )
+
+    def run_main(self):
+        """Run the main workflow."""
+        output_dir = self.layout.dpath_pipelines / self.record_id
```
michellewang, line 112: I wonder if it would be better to download in a temporary directory (i.e. [`tempfile.TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)) instead of here. But this is minor I think so if we want this then it can be done in a follow-up PR

nipoppy/cli.py
mathdugre: I think this can be remove once both command exists.

Shoudl we keep it for this PR and remove it later on tho?

nipoppy/zenodo_api.py
```diff
@@ -0,0 +1,250 @@
...
+    def _create_new_version(self, record_id: str, metadata: dict) -> str:
+        response = httpx.post(
+            f"{self.api_endpoint}/records/{record_id}/versions",
+            headers=self.headers,
+        )
+        if response.status_code != 201:
+            raise ZenodoAPIError(
+                f"Failed to create a new version for zenodo.{record_id}:"
+                + response.json()
+            )
```
michellewang, line 121: Invalid use of `+` operator, this is what we get when test coverage is low 😅 
```suggestion
            raise ZenodoAPIError(
                f"Failed to create a new version for zenodo.{record_id}:"
                f" {response.json()}"
            )
```

michellewang (approved):
Tested this locally, here are some final notes (can be fixed in later PR(s)):

1. `nipoppy pipeline upload` takes `--zenodo-id` option whereas `download` has `RECORD_ID` arg, which is a bit inconsistent
2. We should document which permissions are required for the API tokens. Also, maybe for `--sandbox` we should say that it requires a separate token
3. Is there any scenario where a token is _not_ needed for upload? Currently `--zenodo-token` is optional but maybe we should make it mandatory
4. I think we should remove the `zenodo.json` in the fMRIPrep 24.1.1 folder since now we have default metadata
5. We should allow Zenodo IDs to have the `zenodo.` prefix when updating existing records (works for `download` but not `upload` currently)
6. The pipeline folders are not being put in the correct directory (should be e.g. `pipelines/processing`, not `pipelines`), but this can be fixed when merging with `pipeline install`
7. When I use the generic Zenodo ID (i.e. the one that resolves to the latest version) nothing works (I get 404 errors). Not sure if there is anything we can do about that but it would be nice if we allowed that ID as well

But otherwise this worked for me except for one bug in error handling. Once that is fixed I am okay with merging!

michellewang:
By the way, in the [Zenodo docs](https://developers.zenodo.org/#quickstart-upload) they talk about this new file upload API, which is supposedly more performant (and allows larger files though that is not really relevant for us)

michellewang:
@mathdugre I fixed some merge conflicts, then had to change some things related to the sample pipeline test data: mismatch between Zenodo and what's in `dev-catalog`, since `dev-catalog` now has pipeline variables. Since soon we won't have `sample_pipelines` anymore I made a copy of `fmriprep-24.1.1`, removed `PIPELINE_VARIABLES` and put it in `tests/data`. Will merge since I would like to start working on merging `install`/`download`, but we can revisit my changes later if you want

PR #546, opened 2025-03-14
[BUG] Add validation for non-alphanumeric characters in `participant_id` and `session_id` in manifest

### LINKED ISSUE #545: [BUG] Manifest validation isn't checking for non-alphanumerical characters

### Expected Behavior

`participant_id` and `session_id` cannot have non-alphanumerical characters.

### Current Behavior

Currently we only check that the `sub-` and `ses-` prefixes aren't included

### Error message

_No response_

### Environment

- OS:
- Python version:

### Version

- version:

### How to reproduce

_No response_

### Anything else?

_No response_

michellewang:
- Closes #545

- Throw error if `participant_id` or `session_id` has non-alphanumeric characters
- `visit_id` can still have non-alphanumeric characters

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/utils.py
+++ b/nipoppy/utils.py
@@ -81,7 +81,7 @@ def check_participant_id(participant_id: Optional[str], raise_error=False):
     if participant_id.startswith(BIDS_SUBJECT_PREFIX):
         if raise_error:
             raise ValueError(
-                f'Participant ID should not start with "{BIDS_SUBJECT_PREFIX}"'
+                f'Invalid participant ID: should not start with "{BIDS_SUBJECT_PREFIX}"'
                 f", got {participant_id}"
             )
         else:
@@ -115,7 +115,7 @@ def check_session_id(session_id: Optional[str], raise_error=False):
     if session_id.startswith(BIDS_SESSION_PREFIX):
         if raise_error:
             raise ValueError(
-                f'Session ID should not start with "{BIDS_SESSION_PREFIX}"'
+                f'Invalid session ID: should not start with "{BIDS_SESSION_PREFIX}"'
                 f", got {session_id}"
             )
         else:
--- a/nipoppy/utils.py
+++ b/nipoppy/utils.py
@@ -57,14 +57,20 @@ def session_id_to_bids_session_id(session_id: Optional[str]) -> str:
 
 
 def check_participant_id(participant_id: Optional[str], raise_error=False):
-    """Make sure a participant ID does not have the BIDS prefix.
+    """Make sure a participant ID is valid.
+
+    Specifically:
+    - Check that it does not have the 'sub-' prefix
+    - Check that it only has alphanumeric characters
 
     Parameters
     ----------
     participant_id : Optional[str]
         The participant ID to check. If None, returns None.
-    strict : bool, optional
-        Whether to raise an error if the participant ID has the prefix, by default False
+    raise_error : bool, optional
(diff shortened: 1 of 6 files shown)


nipoppy/utils.py
```diff
@@ -81,24 +87,36 @@ def check_participant_id(participant_id: Optional[str], raise_error=False):
     if participant_id.startswith(BIDS_SUBJECT_PREFIX):
         if raise_error:
             raise ValueError(
-                f'Participant ID should not start with "{BIDS_SUBJECT_PREFIX}"'
+                f'Invalid participant ID: should not start with "{BIDS_SUBJECT_PREFIX}"'
```
nikhil153, line 90: Wouldn't this be caught by non-alphanumeric char test because of `-` ? |
I guess we should still check if `sub` or `ses` are part of participant IDs and force them them to remove it. 
I am also wondering if we should warn against IDs that begin with `0` or `o`

nipoppy/utils.py
```diff
@@ -115,11 +133,18 @@ def check_session_id(session_id: Optional[str], raise_error=False):
     if session_id.startswith(BIDS_SESSION_PREFIX):
         if raise_error:
             raise ValueError(
-                f'Session ID should not start with "{BIDS_SESSION_PREFIX}"'
+                f'Invalid session ID: should not start with "{BIDS_SESSION_PREFIX}"'
```
nikhil153, line 136: Same comments as `participant_id`

nikhil153 (commented):
Couple of comments to see if we should check for additional things to help better naming practices for IDs.

nipoppy/utils.py
michellewang: > Wouldn't this be caught by non-alphanumeric char test because of - ?

Yes, but the reason I still kept it is because this helper function is used in both manifest validation but also to fix CLI inputs (i.e. if users write `--participant-id sub-123` that is fine, we still accept it).

I don't think we should warn against IDs that begin with 0, and if some users name their participant IDs with `sub` inside then that's not ideal but technically still valid, so I don't think we should force them to remove it

nikhil153: I guess, I am not sure in the case where people name their participant IDs as `sub1234`. It would be super confusing / annoying to have BIDS IDs as `sub-sub1234`. 
Similar problem for sessions... 
The `o` and `0` would be just a warning so they don't do that in future or decide to change on their own volition. Not forcing to change.

michellewang: I don't get why having e.g. `001` as `participant_id` is not good

> It would be super confusing / annoying to have BIDS IDs as sub-sub1234

I agree, but maybe there's a reason why some users would want things this way (for example if in tabular files they have `sub1234`). Otherwise I would think that, once they see `sub-sub1234` in the BIDS data, if they are really annoyed then they will redo the BIDSification, which would not take that long

nikhil153: `001` is fine I guess. 
`sub-sub123` is something that we should warn against as best practices.. of course they can see it after BIDSification, but it's going to be better to know about it before that especially since that's the most annoying step which people would not want to repeat... 
I agree we shouldn't error out.. just give it as a warning.

nikhil153 (approved):
I still think we should warn against `sub-sub123`. But up to you.

PR #558, opened 2025-03-19
[ENH] Add `nipoppy pipeline validate` command

michellewang:
- Blocked by #556 
- Closes #557
- Closes #480

Changes proposed in this pull request:
- So I ended up doing this one first since the validation functionalities will be needed later anyway and `nipoppy pipeline install` has a bit of a blocker since it needs the pipeline variables to be ready (need  #548 to be merged ideally)

## Checklist (for reviewers)
_This section is for the PR reviewer_

- [x] PR has an interpretable title with a prefix (e.g. `[BUG]`, `[DOC]`, `[ENH]`, `[MAINT]`)\
Refer to [NumPy Development Guide](https://numpy.org/doc/stable/dev/development_workflow.html#writing-the-commit-message) for a full list
- [x] PR links to GitHub issue with mention `Closes #XXXX`
- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:
- [ ] There is at least one test that would fail under the original bug conditions

PR #560, opened 2025-03-20
[ENH] Force pipeline step config files to only have relative paths

michellewang:
- Closes none

- Force these paths to be relative in pipeline step configs:
  - `INVOCATION_FILE`
  - `DESCRIPTOR_FILE`
  - `TRACKER_CONFIG_FILE`
  - `PYBIDS_IGNORE_FILE`
- Use the new "pipeline type" awareness to refactor some config-file-related things

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


nipoppy/config/pipeline_step.py
```diff
@@ -70,10 +72,42 @@ def validate_before(cls, data: Any) -> Any:
...
+        Specifically:
+
+        - Fields in _path_fields must not be absolute
+        - DESCRIPTOR_FILE and INVOCATION_FILE must both be defined or both be None
+        """
+        for field in self._path_fields:
+            if (path := getattr(self, field)) is not None:
+                # check that path is relative
+                if Path(path).is_absolute():
+                    raise ValueError(f"{field} must be a relative path, got {path}")
```
mathdugre, line 89: We might want to be stricter and ensure the path is in the pipeline's directory.

tests/test_config_pipeline_step.py
```diff
@@ -126,11 +128,51 @@ def test_descriptor_invocation_fields(descriptor_file, invocation_file, expect_e
...
+            ProcPipelineStepConfig,
+            True,
+        ),
+    ],
+)
+def test_absolute_paths(data, pipeline_class, expect_error):
+    with (
+        pytest.raises(ValidationError, match=".* must be a relative path, got")
+        if expect_error
+        else nullcontext()
```
mathdugre, line 171: This is probably not needed since all parametrization have expect_error=True. Similarly, we could remove the expect_error in the parametrization

nipoppy/config/pipeline_step.py
michellewang: Agreed, but we can't do that directly here since we don't know where the file is. But this is something I had in the `nipoppy pipeline validate` PR (#558) though it will have to be updated after this PR is merged

tests/test_config_pipeline_step.py
michellewang: Good catch! 😅

michellewang: Ok so I think I just forgot to have a positive test case, added now

tests/test_workflow_pipeline.py
```diff
@@ -465,6 +465,34 @@ def test_get_pipeline_config(
     )
 
 
+def test_get_pipeline_config_invalid(workflow: PipelineWorkflow):
+    pipeline_name = "new_pipeline"
+    pipeline_version = "1.0.0"
+    dpath_pipeline_bundle = (
+        workflow.layout.dpath_pipelines / "proc" / f"{pipeline_name}-{pipeline_version}"
```
mathdugre, line 472: should be `processing` instead of `proc`

mathdugre (changes requested):

mathdugre (approved):

PR #582, opened 2025-03-31
[FIX] Make runners always bind dataset root directory in containers

### LINKED ISSUE #577: [BUG] `sourcedata/imaging/pre_reorg` is not mounted in containers in `nipoppy bidsify`

### Expected Behavior

This causes problems because `nipoppy reorg` creates symlinks pointing to `sourcedata/imaging/pre_reorg` which are broken inside the container.

We might want to bind the whole dataset root directory too if that does not break anything

### Current Behavior

_No response_

### Error message

_No response_

### Environment

- OS:
- Python version:

### Version

- version:

### How to reproduce

_No response_

### Anything else?

_No response_

michellewang:
- Closes #577

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/workflows/runner.py
+++ b/nipoppy/workflows/runner.py
@@ -70,6 +70,9 @@ class PipelineRunner(BasePipelineWorkflow):
         if bind_paths is None:
             bind_paths = []
 
+        # always bind the dataset's root directory
+        bind_paths = [self.layout.dpath_root] + bind_paths
+
         # get and process container config
         container_config = self.pipeline_step_config.get_container_config()
         container_config = ContainerConfig(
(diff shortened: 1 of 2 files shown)


nipoppy/workflows/runner.py
```diff
@@ -70,6 +70,9 @@ def process_container_config(
         if bind_paths is None:
             bind_paths = []
 
+        # always bind the dataset's root directory
+        bind_paths = [self.layout.dpath_root] + bind_paths
```
nikhil153, line 74: Remind me, why do we need to explicitly bind anything inside `dpath_root` after this? Can't we use relative paths from here?

michellewang: It's because if e.g. the BIDS directory is a symlink to somewhere else, then only binding the root path is not enough (the symlink will be broken inside the container). So it's safer to bind the subdirectories because Nipoppy will resolve the symlink when creating the container command

PR #598, opened 2025-04-15
[FIX] Fix Zenodo keywords not being set correctly

### LINKED ISSUE #597: [BUG] `nipoppy pipeline upload` does not seem to add keywords to Zenodo record

### Expected Behavior

Keywords are in the record, for example like in Boutiques records:

Image

### Current Behavior

Keywords are not in the record: https://sandbox.zenodo.org/records/197584

### Error message

_No response_

### Environment

- OS:
- Python version:

### Version

- version:

### How to reproduce

_No response_

### Anything else?

_No response_

michellewang:
- Closes #597

- Also set the default creator name to "Nipoppy user" instead of "Test, Nipoppy"

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--598.org.readthedocs.build/en/598/


nipoppy/workflows/zenodo.py
```diff
@@ -72,12 +71,14 @@ def _get_pipeline_metadata(
...
-            )
-        )
+        for keyword in [
+            "Nipoppy",
+            f"pipeline_type:{pipeline_config.PIPELINE_TYPE.value}",
+        ]:
+            if (keyword_dict := {"subject": keyword}) not in metadata["metadata"][
+                "subjects"
+            ]:
+                metadata["metadata"]["subjects"].append(keyword_dict)
```
michellewang, line 81: It would be good to have a test that the uploaded records has the expected keywords, but this is not that easy with the way the Zenodo API class is written currently since the draft/version creation methods return the Zenodo ID only, not the full response JSON

michellewang:
Idea: it might also be useful to have `pipeline_name:ABC` and `pipeline_version:XYZ` as keywords. This would allow more filtering in the Zenodo search interface, and also could allow `nipoppy pipeline search` search to display the actual name/version of the pipeline in addition to the Zenodo record title. Thoughts?

mathdugre (approved):

PR #605, opened 2025-04-21
[FIX] Restore substitutions behaviour for backward-compatibility

michellewang:
- Closes none

- Currently the user-defined substitutions only apply within the global config file, which causes problems for users who use the migration script to update their dataset to 0.4, since their pipeline configs don't have pipeline variables (which is fine) and rely on the substitutions being propagated to the config files (`config.json`, invocation/descriptor/etc.)
- So this PR restores the old behaviour where user-defined substitutions (`SUBSTITUTIONS` field in global config file) get propagated to individual pipeline files. These substitutions take precedence over the new pipeline variables for now, which should be fine since we expect existing users to only have substitutions and new users to only have pipeline variables (except for the container store path, and eventually probably the HPC things). In the future we will probably want to handle cases where a substitution is the same as an existing pipeline variable

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--605.org.readthedocs.build/en/605/


julia-pfarr (approved):
I tested the changes and the error does not occur anymore

PR #617, opened 2025-05-01
Allow nipoppy commands to run anywhere within a nipoppy project

### LINKED ISSUE #583: [ENH] Automatically detect if current working directory is inside a Nipoppy dataset?

### New feature

Right now if we run a command inside a subdirectory of the Nipoppy root dataset, we get an error message complaining that a lot of paths are not found. It might be nice to have more `git`-like behaviour where parent directories are searched until a valid Nipoppy dataset is found

### Unclear documentation

_No response_

mathdugre:
- Closes #583 

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--617.org.readthedocs.build/en/617/


nipoppy/utils.py
```diff
@@ -399,3 +404,26 @@ def apply_substitutions_to_json(
...
+
+    This is done by checking if the `.nipoppy` directory exists in the
+    current directory or any of its parents.
+    If the directory is found, it returns the path to the `.nipoppy`
+    directory. If not, it returns False.
+
+    Parameters
+    ----------
+    cwd : _type_, optional
+        _description_, by default Path.cwd()
```
michellewang, line 420: This is causing docs to fail
```suggestion
    cwd : nipoppy.env.StrOrPathLike, optional
        Path to directory, by default Path.cwd()
```

nipoppy/utils.py
```diff
@@ -399,3 +404,26 @@ def apply_substitutions_to_json(
...
+    """
+    current = Path(cwd).resolve()
+    while True:
+        candidate = current / NIPOPPY_DIR_NAME
+        if candidate.is_dir():
+            return current  # Found
+        if current.parent == current:
+            return False  # Reached root, not found
```
michellewang, line 428: This is better than checking for `current == Path('/')`?

nipoppy/workflows/base.py
```diff
@@ -297,14 +297,20 @@ def __init__(
         _validate_layout : bool, optional
             If True, validate the layout during setup, by default True
         """
-        super().__init__(name=name, verbose=verbose, dry_run=dry_run)
-
-        self.dpath_root = Path(dpath_root)
+        self.dpath_root = is_nipoppy_project(dpath_root)
+        # `.nipoppy` is not created by default in version 0.3.4 and below
+        if self.dpath_root is False:
+            self.dpath_root = Path(dpath_root)
```
michellewang, line 303: I think we can create the `.nipoppy` in `run_setup()` after the validation is done in Line 340 -- this way for users who created the dataset with an older version of Nipoppy, after the first time they run with this code they will get this feature as well

nipoppy/layout.py
```diff
@@ -53,6 +53,10 @@ class LayoutConfig(BaseModel):
 
     model_config = ConfigDict(extra="forbid")
 
+    dpath_nipoppy: OptionalDpathInfo = Field(
+        default=OptionalDpathInfo(path=Path(NIPOPPY_DIR_NAME), _is_directory=True),
+        description="Directory Nippopy specific files",
+    )
```
michellewang, line 59: This technically allows users to customize what they want this directory to be called, which we probably don't want. So we can remove this from `LayoutConfig` and instead have it only in `DatasetLayout`

nipoppy/layout.py
```diff
@@ -186,6 +190,7 @@ def __init__(
         self.config = config
 
         # directories (for type hinting)
+        self.dpath_nipoppy: Path
```
michellewang, line 193:
```suggestion
        self.dpath_nipoppy: Path = self.dpath_root / NIPOPPY_DIR_NAME
```

nipoppy/workflows/base.py
michellewang, line 303: Can this be a one-liner instead? Or is that not a good idea (is `pathlib.Path` always truthy?)
```suggestion
        # `.nipoppy` is not created by default in version 0.3.4 and below
        self.dpath_root = is_nipoppy_project(dpath_root) or Path(dpath_root)
```

tests/test_workflow_status.py
```diff
@@ -395,3 +395,23 @@ def test_run(dpath_root: Path, processing_status_table: ProcessingStatusTable):
...
+    dpath_root: Path, processing_status_table: ProcessingStatusTable
+):
+    derivatives = dpath_root.joinpath("derivatives")
+    derivatives.mkdir(parents=True, exist_ok=True)
+
+    workflow = StatusWorkflow(dpath_root=derivatives)
+    workflow.config = get_config()
+    workflow.manifest = make_manifest(n_participants=10)[0]
+    workflow.curation_status_table = CurationStatusTable()  # Checks for empty table
+    workflow.processing_status_table = processing_status_table
```
michellewang, line 414: Would it be a good idea to use a fixture for the workflow since it is repeated between this test and the previous one? For example I think something like this should work:

```python
@pytest.fixture(params=[".", "derivatives"])
def workflow(request: pytest.FixtureRequest, tmp_path: Path) -> StatusWorkflow:
    workflow = StatusWorkflow(dpath_root=tmp_path / request.param)
    workflow.config = get_config()
    workflow.manifest = make_manifest(n_participants=10)[0]
    workflow.curation_status_table = CurationStatusTable()  # Checks for empty table
    workflow.processing_status_table = processing_status_table
    return workflow
```

michellewang (changes requested):
Works well locally, this will be very nice!

Minor comments, and a suggestion to create the `.nipoppy` directory if it doesn't already exist and we checked that the directory passed by the user is a valid Nipoppy dataset

michellewang (approved):
🧑‍🍳

PR #634, opened 2025-05-12
[ENH] Automatically download container when running `nipoppy pipeline install`

### LINKED ISSUE #606: [ENH] Download pipeline container (if needed/possible) when running `nipoppy pipeline install`?

### New feature

Similar to the command printed out in `nipoppy run` but run it by default in `nipoppy pipeline install`. Since the user is supposed to have Internet already given that it is needed to download things from Zenodo 

### Unclear documentation

_No response_

michellewang:
- Closes #606

- If container URI is in pipeline config file, pull the container (ask user to confirm first)
- Do not allow container URI to be specified if file path is not specified (in container info config)
- Some aesthetic changes so that `rich` console outputs (i.e. things we print that do not use the logger) are aligned/indented the same way as the log messages

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:


nipoppy/workflows/pipeline_store/install.py
```diff
@@ -98,6 +105,58 @@ def _update_config_and_save(self, pipeline_config: BasePipelineConfig) -> Config
...
+                            pipeline_config.CONTAINER_INFO.URI,
+                        ]
+                    )
+                    status.update(f"Container downloaded to {fpath_container}")
+            except subprocess.CalledProcessError as exception:
+                self.logger.error(
+                    f"Failed to download container {pipeline_config.CONTAINER_INFO.URI}"
+                    f" to {fpath_container}: {exception}"
+                )
+                self.return_code = ReturnCode.UNKNOWN_FAILURE
```
mathdugre, line 158: Do we still consider the pipeline installation as successful if the container cannot download?
If not, we might want to raise SystemExit instead.

tests/test_workflow_pipeline_store_install.py
```diff
@@ -155,6 +167,137 @@ def test_update_config_and_save_no_other_change(
...
+
+    mocked_confirm_ask = mocker.patch(
+        "nipoppy.workflows.pipeline_store.install.Confirm.ask", return_value=False
+    )
+
+    mocked_run_command = mocker.patch.object(workflow, "run_command")
+
+    workflow._download_container(pipeline_config)
+    mocked_confirm_ask.assert_called_once()
+    mocked_run_command.assert_not_called()
```
mathdugre, line 223: 🍒 This could be a single function with pytest.mark.parametrize

tests/test_workflow_pipeline_store_install.py
```diff
@@ -155,6 +167,137 @@ def test_update_config_and_save_no_other_change(
...
+):
+    fpath_container = (
+        workflow.layout.dpath_containers / pipeline_config.CONTAINER_INFO.FILE.name
+    )
+    fpath_container.parent.mkdir(parents=True, exist_ok=True)
+    fpath_container.touch()
+    mocked = mocker.patch.object(workflow, "run_command")
+
+    workflow._download_container(pipeline_config)
+    mocked.assert_not_called()
```
mathdugre, line 298: 🍒 This is repeated in most test. It could be refactored in a function

michellewang: Usually the assert is a bit different between the functions, so I think we can leave things as they are now

mathdugre (approved):
🧑‍🍳

Issue #520, opened 2025-02-26
[ENH] init: add --force option to allow initializing in a non-empty folder


yarikoptic:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### New feature

ATM, if I decided to initiate in e.g.within empty datalad dataset, nipoppy would forbid

```
❯ nipoppy init --dataset $PWD-try1
...
         FileExistsError: Dataset directory is non-empty: /home/yoh/proj/nipoppy-try1                                                                                               

❯ ls -l $PWD-try1
total 0

❯ ls -la $PWD-try1
total 4
drwxrwxr-x 1 yoh yoh   52 Feb 26 10:47 ./
drwxr-xr-x 1 yoh yoh 9732 Feb 26 10:47 ../
drwxrwxr-x 1 yoh yoh   40 Feb 26 10:47 .datalad/
drwxrwxr-x 1 yoh yoh  190 Feb 26 10:47 .git/
-rw-rw-r-- 1 yoh yoh  115 Feb 26 10:47 .gitattributes

```

It would be nice to be able to allow it to proceed in such cases. Then we could potentially use it with `datalad run` to make a git commit record on initialization using nipoppy.

alternatives
- make it ignore "hidden" files (all files and folders starting with period) 
- or specifically ignore git and datalad files

mathdugre:
Hi @asmacdo, that would be great!

You can find the current implementation for handling initialization in a non-empty directory here:
https://github.com/nipoppy/nipoppy/blob/ea67a3df3dfe96703ef276eeccf64d653911f0ef/nipoppy/workflows/dataset_init.py#L65-L77

PR #682, opened 2025-07-10
[ENH]: allow forced dataset init on non-empty dirs

asmacdo:
This PR adds `--force` flag option to the init command. 

This is useful for dirs that are already under version control and have .git or .datalad dirs.

Fixes https://github.com/nipoppy/nipoppy/issues/520

### TODO fix Collision behavior (should be discussed before merge)

What should the expected behavior be when there are existing files that will collide?
[In this comment](https://github.com/nipoppy/nipoppy/issues/520#issuecomment-3058127939) and in the helptext I added, we would clobber those files, which seems reasonable to me. 

This seems to be the case for all files except for the bids source.  When I run the init --force command twice, I would expect to clobber all files, but instead there is a problem with the bids source files.  
`FileExistsError: [Errno 17] File exists: '/home/austin/devel/sandbox/ds004101' -> '/tmp/nipoppy-force-test/bids'` 

I get this same error in symlink mode and copy mode, move mode hits a permissions error due to the source data being a datalad dataset. 

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--682.org.readthedocs.build/en/682/

### DIFF
--- a/nipoppy/cli.py
+++ b/nipoppy/cli.py
@@ -320,6 +320,14 @@ def cli():
     help=("Path to a BIDS dataset to initialize the layout with."),
 )
 @click.option(
+    "--force",
+    is_flag=True,
+    help=(
+        "Create a nipoppy dataset even if there are already files present."
+        " (May clobber existing files.)"
+    ),
+)
+@click.option(
     "--mode",
     type=click.Choice(["copy", "move", "symlink"]),
     default="symlink",
--- a/nipoppy/cli.py
+++ b/nipoppy/cli.py
@@ -321,6 +321,7 @@ def cli():
 )
 @click.option(
     "--force",
+    "-f",
     is_flag=True,
     help=(
         "Create a nipoppy dataset even if there are already files present."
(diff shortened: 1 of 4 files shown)


asmacdo:
I think it could be merged as-is, just wanted to make sure the clobber edge case on the bids source dir is alright.

mathdugre:
That looks good as-is. Thank you for the contribution 🎉

mathdugre (approved):

michellewang:
@asmacdo thanks for the PR!

I discussed the edge case with @mathdugre briefly and we think it would make sense to clobber the BIDS directory too when both `--bids-source` and `--force` are given, so probably worth it to fix that before merging. 

I think something like this should work:
- If `bids_source` and `force` and BIDS directory exists (and `bids_source` is not the BIDS directory inside the Nipoppy dataset), delete it (might have to handle symlink vs not symlink case)
- Then handle rest (copy/move/symlink modes) as usual

https://github.com/nipoppy/nipoppy/blob/89322e59f12cd0c3235b24b85f9841ed3f76df2b/nipoppy/workflows/dataset_init.py#L89-L100

If you see a better way, feel free to do that instead!

asmacdo:
Thanks @mathdugre, updated!

PR #719, opened 2025-08-08
[ENH] Expose more data in template HPC job script

### LINKED ISSUE #717: [ENH] Refactor template HPC job script

### New feature

Currently in the template we print an array of all the commands to run, then index them with the job array ID. That can be annoying because the participant and session IDs are not directly accessible inside the script.

Instead we could print an array of participants and sessions (or two arrays) and index those instead with the job array ID. This would expose the participant ID and the session ID which could be used later in the script.

We should also pass the pipeline name and version to the template for more flexibility in custom scripting.

### Unclear documentation

_No response_

michellewang:
- Closes #717

- Add:
   - Path to dataset root directory
   - Pipeline name, pipeline version, pipeline step
   - List of participant IDs
   - List of session IDs
- No breaking changes because the command list is still the same

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

### DIFF
--- a/nipoppy/data/hpc/job_script_template.sh
+++ b/nipoppy/data/hpc/job_script_template.sh
@@ -70,7 +70,14 @@
 {%- endif %}
 {% endif %}
 
-{#-
+# for custom scripting
+DPATH_ROOT="{{ NIPOPPY_DPATH_ROOT }}"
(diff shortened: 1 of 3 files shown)


mathdugre (approved):
LGTM 🧑‍🍳

PR #729, opened 2025-08-13
[MAINT] Parallelize tests with xdist

### LINKED ISSUE #723: [ENH] Parallelizing pytest tests by default for local and CI executions

### New feature

Parallelizing pytest tests by default for local and CI executions

### Unclear documentation

_No response_

mathdugre:
- Closes #723

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--729.org.readthedocs.build/en/729/


mathdugre:
Seem flaky because of the interactiong with pytest.LogCaptureFixture

michellewang:
@mathdugre adding the `clean_loggers` fixture somehow speeds up the tests even if `--dist loadfile` is specified. Locally all tests pass all the time but for some reason on GitHub actions some of them still fail :/

mathdugre:
Thank you for looking into it

michellewang:
I have an idea which might work if it’s true that it’s only the `run_command` tests that are flaky: we mark them as “no xdist” or something like that and run them separately in CI. I’ll try something later

mathdugre:
I'm not a fan of adding an extra pytest marker to prevent the flaky parallel tests from failing.
How much speedup do we expect on GitHub from this? @michellewang

mathdugre:
Let's do it then. We could try to figure out the root issue for the flaky parallel tests in a future issue.

michellewang:
Not sure if related but the [`pytest-xdist` docs](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers) say that their [`execnet` dependency](https://github.com/pytest-dev/execnet) uses stdout internally for some things, so that might be interfering with our failing tests. But the tests are only failing on GitHub, so there could be an environment-related factor too

mathdugre:
> Not sure if related but the [`pytest-xdist` docs](https://pytest-xdist.readthedocs.io/en/stable/known-limitations.html#output-stdout-and-stderr-from-workers) say that their [`execnet` dependency](https://github.com/pytest-dev/execnet) uses stdout internally for some things, so that might be interfering with our failing tests. But the tests are only failing on GitHub, so there could be an environment-related factor too

I think that's only for the `-s` flag from pytest, not with general logging.

PR #742, opened 2025-08-21
[ENH] Add `--use-subcohort` argument to `nipoppy process`/`bidsify`/`extract`

### LINKED ISSUE #736: [ENH] Add CLI option for user-provided list of participants/sessions to run

### New feature

We have `--write-list` for the runners which writes a TSV of participants and sessions that still need to be run (useful for local parallelization or HPC job submission without using the `--hpc` option). However, we currently don't have a way for users to specify a precise list of participants/sessions to run (for example a sub-cohort). We can add an option (say `--use-list`) which takes a TSV in the same format that `--write-list` outputs and filters down the original participant/session list (from diff between status files) with those user-provided participant/sessions. 

### Unclear documentation

_No response_

michellewang:
- Closes #736 

- `--use-list` takes as input a file in the same format that `--write-list` outputs
- Also implemented input validation for `--hpc` and `--write-list` which was previously a TODO comment

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--742.org.readthedocs.build/en/742/


nipoppy/workflows/pipeline.py
```diff
@@ -149,17 +149,25 @@ def __init__(
...
         participant_id: str = None,
         session_id: str = None,
+        use_list: Optional[StrOrPathLike] = None,
         hpc: Optional[str] = None,
         write_list: Optional[StrOrPathLike] = None,
         fpath_layout: Optional[StrOrPathLike] = None,
         verbose: bool = False,
         dry_run=False,
     ):
+        if hpc and (write_list is not None):
```
mathdugre, line 159: The syntax is a bit inconsistent.
```suggestion
        if hpc and write_list:
```

nipoppy/workflows/pipeline.py
```diff
@@ -149,17 +149,25 @@ def __init__(
...
+            raise ValueError(
```
mathdugre, line 160: Do you think an `AttributeError` could make more sense here?

tests/unit/workflows/test_pipeline.py
```diff
@@ -233,6 +234,26 @@ def test_init(args):
...
+                "dpath_root": "my_dataset",
+                "pipeline_name": "my_pipeline",
+                "hpc": "slurm",
+                "write_list": "list.tsv",
+            },
+            ValueError,
+            "HPC job submission and writing a list of participants and sessions are mutually exclusive.",  # noqa: E501
+        ),
+    ],
+)
```
mathdugre, line 251: It might be cleaner to have the values directly in the test, since we only have a single set of value tested.

tests/unit/test_cli.py
```diff
@@ -128,6 +128,18 @@ def test_cli_gui_visibility(monkeypatch, trogon_installed):
...
+                "bidsify",
+                "--dataset",
+                "[mocked_dir]",
+                "--pipeline",
+                "my_pipeline",
+                "--use-list",
+                "[mocked_file]",
+            ],
+            "nipoppy.workflows.bids_conversion.BidsConversionRunner",
+        ),
```
mathdugre, line 142: Not sure if it's necessary. We don't test individual flags for other sub-commands right now. This could be something we can consider in the future?

mathdugre (commented):
I'm not a fan of the `--use-list` name. It sounds quite vague.
Maybe`--subset`?, but it's not that much better. We might want to discuss more on the name to use.

michellewang:
> I'm not a fan of the --use-list name. It sounds quite vague.
> Maybe--subset?, but it's not that much better. We might want to discuss more on the name to use.

My reasoning was that we already have `--write-list` and they are related to each other, but I agree that `--use-list` is not a great name. We can think about it a bit more.

nipoppy/workflows/pipeline.py
michellewang: Not sure, I feel `AttributeError` is more for when you try to access an object's attribute but something goes wrong

michellewang:
Decision at docs meeting:
- `--write-subcohort` --> deprecate `--write-list`
- `--use-subcohort`

nipoppy/workflows/pipeline.py
mathdugre: Not sure which one is best, or the exact usage of `AttributeError`. Let's keep the `ValueError` then.

mathdugre (approved):
LGTM 🧑‍🍳

PR #773, opened 2025-09-18
[ENH] Add support for Docker

### LINKED ISSUE #700: [ENH] Docker support

### New feature

Following #686, we are now closer to having Docker support. The missing parts are:
- [x] Add `docker` as valid container command [here](https://github.com/nipoppy/nipoppy/blob/18d5bde00c0683497dabf9a226cd83edf7aae43c/nipoppy/env.py#L32-L36)
- [x] Update `prepare_container` and `check_container_args` to handle `docker` args (i.e. instead of using hardcoded Apptainer bind flags we should have some logic for using either Apptainer/Singularity or Docker flag/syntax)
    - [ ] Possibly also remove `--cleanenv` if it's there? Or replace by Docker equivalent if it exists. Or expect users to delete it themselves from global config but that's a bit annoying since by default we have `--cleanenv`. In general I don't think we want to maintain a mapping of Apptainer<->Docker args
- [x] Environment variables can be set in Apptainer/Singularity containers through `APPTAINERENV_{VAR_NAME}`/`SINGULARITYENV_{VAR_NAME}`. It doesn't seem like there's an equivalent for Docker, but both Apptainer/Singularity and Docker accept `--env` flags so maybe we should just move to that instead
- [x] Update container pull commands ([here](https://github.com/nipoppy/nipoppy/blob/18d5bde00c0683497dabf9a226cd83edf7aae43c/nipoppy/workflows/pipeline_store/install.py#L142-L149) and [here](https://github.com/nipoppy/nipoppy/blob/18d5bde00c0683497dabf9a226cd83edf7aae43c/nipoppy/workflows/pipeline.py#L265-L267)). Docker stores image files centrally (is that the only option?) so it doesn't really make sense to ask the user to pull these containers in the first place
- [x] Use `--force-docker` flag in runners ([here](https://github.com/nipoppy/nipoppy/blob/18d5bde00c0683497dabf9a226cd83edf7aae43c/nipoppy/workflows/runner.py#L150-L165))
- [ ] Double-check that docs do not say that Nipoppy is only compatible with Apptainer/Singularity or Linux

### Unclear documentation

_No response_

michellewang:
- Closes #700

- Refactor container-related functions from `nipoppy.config.container` to `ContainerHandler` classes
- Use `ContainerHandler` in runner classes as well as when checking if a container image has already been downloaded/generating the command to pull the container
- Had to change the way bind args was handled because Docker does not seem to support things like `--volume /some/path` (need `--volume /some/path:/somepath` instead)

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--773.org.readthedocs.build/en/773/


mathdugre (changes requested):

mathdugre (changes requested):

nipoppy/workflows/runner.py
```diff
@@ -60,21 +62,31 @@ def launch_boutiques_run(
...
             else:
                 bosh_exec_launch_args.extend(
                     [
-                        "--force-singularity",
                         "--no-automount",
-                        "--imagepath",
-                        str(self.fpath_container),
                         "--container-opts",
-                        shlex.join(container_config.ARGS),
+                        shlex.join(container_handler.args),
```
mathdugre:
```suggestion
f"--container-opts={shlex.join(container_handler.args)}",
```

mathdugre: The current Boutiques CLI parser has issues parsing `--container-opts` value that have a flag from the bosh cli (e.g., `-v`)

michellewang: Sorry, I thought I did it, maybe didn’t push

mathdugre (approved):

PR #816, opened 2025-11-06
[ENH] Add `Study` and `NipoppyDataRetriever` classes

### LINKED ISSUE #814: [ENH] Create `NipoppyStudy` class for representing a Nipoppy dataset

### New feature

It can contain the `DatasetLayout` object, the manifest, and the global config.

It can also have a query/get method for wrangling tabular files.

### Unclear documentation

_No response_

michellewang:
- Closes #814

- Extract `Study` class from `BaseDatasetWorkflow` in preparation for eventual deeper refactor
    - Contains `layout`, `manifest`, `config`, `curation_status_table` and `processing_status_table`
    - Does not have `dicom_dir_map` for now, I'm not sure how to handle this since the map is not necessarily a file on disk
- Create `NipoppyDataRetriever` class for getting tabular data from disk
    - ~Currently only supports tabular derivatives (output/IDP) data, but will add demographics data soon (need to discuss a bit more/check with Neurobagel team)~ (done)

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--816.org.readthedocs.build/en/816/

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,32 @@
+---
+name: Check documentation links
+
+on:
+  pull_request:
+    branches: ['main']
+  schedule:
+    - cron: '0 0 * * 0' # Weekly on Sunday at 00:00 UTC
+  workflow_dispatch:
+
+jobs:
+  linkcheck:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5
+
+    - uses: actions/setup-python@v6
+      with:
+        python-version: "3.11"
+
+    - name: Install package with doc dependencies
+      run: |
+        pip install -U pip
+        pip install .[doc]
+
+    - name: Generate JSON schemas for docs
+      run: |
+        python docs/scripts/pydantic_to_jsonschema.py
+
+    - name: Run linkcheck
+      run: |
+        python -m sphinx -b linkcheck -D linkcheck_timeout=30 docs/source/ docs/_build/linkcheck
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -13,10 +13,6 @@ build:
   jobs:
     pre_build:
       - python docs/scripts/pydantic_to_jsonschema.py
-    post_build:
-      # check for broken external links
-      # https://docs.readthedocs.com/platform/stable/build-customization.html#perform-a-check-for-broken-links
-      - python -m sphinx -b linkcheck -D linkcheck_timeout=30 docs/source/ $READTHEDOCS_OUTPUT/linkcheck
 
 python:
   install:


nipoppy/workflows/base.py
```diff
@@ -307,18 +303,21 @@ def __init__(
...
         self.fpath_layout = fpath_layout
         self._skip_logfile = _skip_logfile
         self._validate_layout = _validate_layout
 
-        self.layout = DatasetLayout(
+        self.study = Study(
             dpath_root=self.dpath_root,
-            fpath_config=self.fpath_layout,
+            fpath_layout=self.fpath_layout,
+            logger=self.logger,
```
mathdugre: The workflow should accept a `study: Study` params direclty. This would avoid passing the other params (`dpath_root`, `fpath_config`, `fpath_layout`) separately in every workflows

nipoppy/workflows/base.py
mathdugre: We probably should move this to the new Study class

nipoppy/data_api.py
```diff
@@ -0,0 +1,122 @@
...
+                    f", got invalid specification {derivatives_spec}"
+                )
+
+    def _get_derivatives_table(
+        self,
+        pipeline_name: str,
+        pipeline_version: str,
+        filepath_pattern: str,
+    ) -> pd.DataFrame:
+        candidate_paths = list(
```
mathdugre: Maybe rename to `candidate_path` (no s). This would make it more explicit that we only expect one. Otherwise, the pop at line 84 might not be obvious at quick glance.

nipoppy/data_api.py
```diff
@@ -0,0 +1,122 @@
...
+            a MultiIndex of participant IDs and session IDs.
+        """
+        # input validation
+        if derivatives is None:
+            derivatives = []
+        # NOTE this will make more sense once demographic data is supported
+        if not derivatives:
+            raise ValueError(
+                "At least one derivative/demographic specification must be defined."
+            )
```
mathdugre: Isn't setting the derivatives to `[]` always failing the condition that follows?

nipoppy/data_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+        dfs = []
+        for derivatives_spec in derivatives:
+            dfs.append(self._get_derivatives_table(*derivatives_spec))
+
+        df = pd.concat(dfs, axis="columns", join="outer")
+
+        # only use participants/sessions from the manifest
+        df = df.loc[df.index.isin(self.study.manifest.get_participants_sessions()), :]
```
mathdugre: We could return this directly instead of updating df and returning it right after

tests/unit/test_study.py
```diff
@@ -0,0 +1,100 @@
...
+    return Study(dpath_root=dpath_root)
+
+
+def test_init_default(tmp_path: Path):
+    custom_logger = get_logger(name="nipoppy.custom_logger", verbose=True)
+    study = Study(dpath_root=tmp_path, logger=custom_logger)
+    assert study.dpath_root == tmp_path
+    assert study.fpath_layout is None
+    assert study.logger == custom_logger
+    assert study.verbose is False
```
mathdugre: Same as above

tests/unit/test_study.py
```diff
@@ -0,0 +1,100 @@
...
+    fname_layout,
+    tmp_path: Path,
+    mocker: pytest_mock.MockFixture,
+):
+    fpath_layout = tmp_path / fname_layout
+
+    mocked_layout = mocker.patch("nipoppy.study.DatasetLayout")
+    study.fpath_layout = fpath_layout
+
+    _ = study.layout
```
mathdugre: Why the `_ =` ... ? Can't we skip the assignment? Otherwise, this should be documented.

tests/unit/test_data_api.py
```diff
@@ -0,0 +1,176 @@
...
+
+@pytest.fixture
+def api(tmp_path: Path) -> NipoppyDataAPI:
+    dpath_root = tmp_path / "my_study"
+    return NipoppyDataAPI(dpath_root=dpath_root)
+
+
+def test_init(api: NipoppyDataAPI):
+    assert api.dpath_root == api.dpath_root
+    assert isinstance(api.study, Study)
```
mathdugre: I don't think this is needed

nipoppy/study.py
```diff
@@ -0,0 +1,117 @@
...
+            raise FileNotFoundError(f"Manifest file not found: {fpath_manifest}")
+
+    @cached_property
+    def curation_status_table(self) -> CurationStatusTable:
+        """The curation status table."""
+        fpath_table = self.layout.fpath_curation_status
+        try:
+            return CurationStatusTable.load(fpath_table)
+        except FileNotFoundError:
+            raise FileNotFoundError(f"Curation status file not found: {fpath_table}")
```
mathdugre: Same as above

nipoppy/study.py
```diff
@@ -0,0 +1,117 @@
...
+        return config
+
+    @cached_property
+    def manifest(self) -> Manifest:
+        """The manifest table."""
+        fpath_manifest = Path(self.layout.fpath_manifest)
+        try:
+            return Manifest.load(fpath_manifest)
+        except FileNotFoundError:
+            raise FileNotFoundError(f"Manifest file not found: {fpath_manifest}")
```
mathdugre: Same as above

mathdugre (changes requested):

nipoppy/data_api.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def _load_tsv(self, fpath: StrOrPathLike, index_cols: List[str]) -> pd.DataFrame:
+        return pd.read_csv(
+            fpath,
+            sep="\t",
+            index_col=index_cols,
+            dtype={col: str for col in index_cols},
+        )
+
+    def _check_derivatives_arg(
```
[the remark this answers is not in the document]
michellewang: Good catch, I accidentally deleted the call to this function...

nipoppy/study.py
```diff
@@ -0,0 +1,117 @@
...
+
+class Study(Base):
+    """Representation of a Nipoppy study."""
+
+    def __init__(
+        self,
+        dpath_root: StrOrPathLike,
+        fpath_layout: Optional[StrOrPathLike] = None,
+        logger: Optional[Logger] = None,
+        verbose: bool = False,
```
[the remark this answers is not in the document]
michellewang: If we have an open issue about it, I don't think we need a TODO in the code

tests/unit/test_study.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+@pytest.mark.parametrize("fname_layout", ["layout1.json", "layout2.json"])
+def test_layout(
+    study: Study,
+    fname_layout,
+    tmp_path: Path,
+    mocker: pytest_mock.MockFixture,
+):
+    fpath_layout = tmp_path / fname_layout
```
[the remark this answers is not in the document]
michellewang: No?

The `study` fixture uses the default layout (which is valid). Here we test custom layout files. If we add this parametrization to the `study` fixture, it will duplicate all the tests where the fixture is used. And we'll need to mock the layout creation since the file paths used here do not exist.

nipoppy/workflows/base.py
michellewang: Agreed but let's do this in a future PR

nipoppy/workflows/base.py
```diff
@@ -349,50 +348,13 @@ def run_setup(self):
...
-        fpath_manifest = Path(self.layout.fpath_manifest)
-        try:
-            return Manifest.load(fpath_manifest)
-        except FileNotFoundError:
-            raise FileNotFoundError(f"Manifest file not found: {fpath_manifest}")
+        """The manifest table."""
+        return self.study.manifest
 
     @cached_property
     def curation_status_table(self) -> CurationStatusTable:
```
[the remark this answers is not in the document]
michellewang: For this one I disagree since it is treated differently from `layout`, `config`, and `manifest`: if the study doesn't have the curation status table then we create it automatically. I don't think this logic should be in the `Study` class itself since (for now) it is just a view on the dataset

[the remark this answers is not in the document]
michellewang: Other difficulty with this one in particular is that we are using workflow methods to write the file to disk. So can't easily move to `Study` class without introducing circularity

nipoppy/data_api.py
```diff
@@ -0,0 +1,122 @@
...
+from typing import List, Optional, Tuple
+
+import pandas as pd
+
+from nipoppy.env import StrOrPathLike
+from nipoppy.study import Study
+from nipoppy.tabular.manifest import Manifest
+
+
+class NipoppyDataAPI:
```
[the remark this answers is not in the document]
mathdugre: @nikhil153 , do you have opinion on this?

nipoppy/workflows/base.py
```diff
@@ -442,7 +403,7 @@ def processing_status_table(self) -> ProcessingStatusTable:
         Otherwise, return an empty processing status table.
         """
         try:
-            return ProcessingStatusTable.load(self.layout.fpath_processing_status)
+            return self.study.processing_status_table
         except FileNotFoundError:
             return ProcessingStatusTable()
```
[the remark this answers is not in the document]
mathdugre: After discussion with @michellewang , we agree that the Study class should be used to retrieve information about the study (from disk), and eventually pipelines it has.
Possibly, we could improve the docstring of the class to clarify its purpose.

nipoppy/data_api.py
```diff
@@ -60,59 +81,76 @@ def _check_derivatives_arg(
...
                 f"Found more than one file matching {filepath_pattern} for pipeline "
-                f"{pipeline_name}, version {pipeline_version}: {candidate_paths}"
+                f"{pipeline_name}, version {pipeline_version}: {candidate_path}"
             )
-        return self._load_tsv(candidate_paths.pop(), index_cols=self.imaging_index_cols)
+        return self._load_tsv(
+            candidate_path.pop(), index_cols=self.index_cols_derivatives
+        )
 
     def get_tabular_data(
```
mathdugre: Would be great to have separate functions for `get_derivatives` and `get_phenotypes`. Those functions could be reuse here to merge both tables

mathdugre (commented):
Only a few nitpicks and questions

tests/unit/test_data_retriever.py
```diff
@@ -0,0 +1,374 @@
...
+@pytest.mark.parametrize(
+    "phenotypes,derivatives,get_phenotypes_called,get_derivatives_called",
+    [
+        (["nb:Age"], None, True, False),
+        (
+            None,
+            [("pipeline1", "v1.0", "pattern1")],
+            False,
+            True,
+        ),
```
mathdugre: I would prefer the params to have the same formatting (for readability), either both one-liners or mutli-lines

mathdugre (approved):
Minor comments, otherwise LGTM 🧑‍🍳

PR #822, opened 2025-11-13
[MAINT] Use global logger

### LINKED ISSUE #818: [MAINT] Refactor logging to reduce coupling
Currently, each workflow creates a logger which is then pass around functions it calls. This necessities multiple classes and functions to accept a logger as an argument. To reduce the coupling between the different components and the logging, we should restructure the logging to have a single logger which can be imported, and modified as needed, in different modules.

mathdugre:
- Closes #818

- 

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--822.org.readthedocs.build/en/822/

### DIFF
--- a/nipoppy/workflows/base.py
+++ b/nipoppy/workflows/base.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import json
 import logging
 import os
 import shlex
@@ -14,24 +13,19 @@ from pathlib import Path
 from typing import Optional, Sequence
 
 from nipoppy.base import Base
-from nipoppy.config.main import Config
 from nipoppy.env import EXT_LOG, StrOrPathLike
 from nipoppy.exceptions import FileOperationError, ReturnCode
 from nipoppy.layout import DatasetLayout
 from nipoppy.logger import get_logger
+from nipoppy.study import Study
 from nipoppy.tabular.base import BaseTabular
 from nipoppy.tabular.curation_status import (
     CurationStatusTable,
     generate_curation_status_table,
 )
 from nipoppy.tabular.dicom_dir_map import DicomDirMap
-from nipoppy.tabular.manifest import Manifest
 from nipoppy.tabular.processing_status import ProcessingStatusTable
-from nipoppy.utils.utils import (
-    add_path_timestamp,
-    is_nipoppy_project,
-    process_template_str,
-)
+from nipoppy.utils.utils import add_path_timestamp, is_nipoppy_project
 
 logger = get_logger()
 
@@ -180,7 +174,6 @@ class BaseWorkflow(Base, ABC):
         """Run the main part of the workflow."""
         pass
 
-    @abstractmethod
     def run_cleanup(self):
         """Run the cleanup part of the workflow."""
         pass
@@ -304,18 +297,20 @@ class BaseDatasetWorkflow(BaseWorkflow, ABC):
         _validate_layout : bool, optional
             If True, validate the layout during setup, by default True
         """
+        super().__init__(name=name, verbose=verbose, dry_run=dry_run)
+
         # `.nipoppy` is not created by default in version 0.3.4 and below
         self.dpath_root = is_nipoppy_project(dpath_root) or Path(dpath_root)
         self.fpath_layout = fpath_layout
         self._skip_logfile = _skip_logfile
         self._validate_layout = _validate_layout
 
-        self.layout = DatasetLayout(
-            dpath_root=self.dpath_root,
-            fpath_config=self.fpath_layout,
+        self.study = Study(
+            DatasetLayout(
+                dpath_root=self.dpath_root,
+                fpath_config=self.fpath_layout,
+            )
         )
-
-        super().__init__(name=name, verbose=verbose, dry_run=dry_run)
 
     def generate_fpath_log(
         self,
@@ -329,7 +324,7 @@ class BaseDatasetWorkflow(BaseWorkflow, ABC):
             dnames_parent = [dnames_parent]
         if fname_stem is None:
             fname_stem = self.name
-        dpath_log = self.layout.dpath_logs / self.name
+        dpath_log = self.study.layout.dpath_logs / self.name
         for dname in dnames_parent:
             dpath_log = dpath_log / dname
         return dpath_log / add_path_timestamp(f"{fname_stem}{EXT_LOG}")
@@ -337,61 +332,12 @@ class BaseDatasetWorkflow(BaseWorkflow, ABC):
     def run_setup(self):
         """Run the setup part of the workflow."""
         if self._validate_layout:
-            self.layout.validate()
+            self.study.layout.validate()
 
         if not self._skip_logfile:
             logger.add_file_handler(self.generate_fpath_log())
 
         super().run_setup()
-
-    @cached_property
-    def config(self) -> Config:
-        """
-        Load the configuration.
-
-        Raise error if not found.
-        """
-        fpath_config = self.layout.fpath_config
-        try:
-            # load and apply user-defined substitutions
-            logger.debug(f"Loading config from {fpath_config}")
-            config = Config.load(fpath_config)
-        except FileNotFoundError as e:
-            raise FileOperationError(
-                f"Config file not found: {self.layout.fpath_config}"
-            ) from e
-
-        # replace path placeholders in the config
-        # (except in the user-defined substitutions)
-        user_substitutions = config.SUBSTITUTIONS  # stash original substitutions
-        # this might modify the SUBSTITUTIONS field (which we don't want)
-        config = Config(
-            **json.loads(
-                process_template_str(
-                    config.model_dump_json(),
-                    objs=[self, self.layout],
-                )
-            )
-        )
-        # restore original substitutions
-        config.SUBSTITUTIONS = user_substitutions
-
-        return config
-
-    @cached_property
-    def manifest(self) -> Manifest:
-        """
-        Load the manifest.
-
-        Raise error if not found.
-        """
-        fpath_manifest = Path(self.layout.fpath_manifest)
-        try:
-            return Manifest.load(fpath_manifest)
-        except FileNotFoundError as e:
-            raise FileOperationError(
-                f"Manifest file not found: {fpath_manifest}"
-            ) from e
 
     @cached_property
     def curation_status_table(self) -> CurationStatusTable:
@@ -400,20 +346,20 @@ class BaseDatasetWorkflow(BaseWorkflow, ABC):
 
         Otherwise, generate a new one.
         """
-        fpath_table = Path(self.layout.fpath_curation_status)
+        fpath_table = Path(self.study.layout.fpath_curation_status)
         try:
-            return CurationStatusTable.load(fpath_table)
+            return self.study.curation_status_table
         except FileNotFoundError:
             logger.warning(
                 f"Curation status file not found: {fpath_table}"
                 ". Generating a new one on-the-fly"
             )
             table = generate_curation_status_table(
-                manifest=self.manifest,
+                manifest=self.study.manifest,
                 dicom_dir_map=self.dicom_dir_map,
-                dpath_downloaded=self.layout.dpath_pre_reorg,
-                dpath_organized=self.layout.dpath_post_reorg,
-                dpath_bidsified=self.layout.dpath_bids,
+                dpath_downloaded=self.study.layout.dpath_pre_reorg,
+                dpath_organized=self.study.layout.dpath_post_reorg,
+                dpath_bidsified=self.study.layout.dpath_bids,
                 empty=False,
             )
 
@@ -439,21 +385,21 @@ class BaseDatasetWorkflow(BaseWorkflow, ABC):
         Otherwise, return an empty processing status table.
         """
         try:
-            return ProcessingStatusTable.load(self.layout.fpath_processing_status)
+            return self.study.processing_status_table
         except FileNotFoundError:
             return ProcessingStatusTable()
 
     @cached_property
     def dicom_dir_map(self) -> DicomDirMap:
         """Get the DICOM directory mapping."""
-        fpath_dicom_dir_map = self.config.DICOM_DIR_MAP_FILE
+        fpath_dicom_dir_map = self.study.config.DICOM_DIR_MAP_FILE
         if fpath_dicom_dir_map is not None and not Path(fpath_dicom_dir_map).exists():
             raise FileOperationError(
                 f"DICOM directory map file not found: {fpath_dicom_dir_map}"
             )
 
         return DicomDirMap.load_or_generate(
-            manifest=self.manifest,
+            manifest=self.study.manifest,
             fpath_dicom_dir_map=fpath_dicom_dir_map,
-            participant_first=self.config.DICOM_DIR_PARTICIPANT_FIRST,
+            participant_first=self.study.config.DICOM_DIR_PARTICIPANT_FIRST,
         )
--- a/nipoppy/workflows/pipeline_store/list.py
+++ b/nipoppy/workflows/pipeline_store/list.py
@@ -39,11 +39,13 @@ class PipelineListWorkflow(BaseDatasetWorkflow):
         for pipeline_type in PipelineTypeEnum:
             pipeline_names_to_versions_map = defaultdict(list)
             dpath_pipeline_bundles = (
-                self.layout.dpath_pipelines
+                self.study.layout.dpath_pipelines
                 / DatasetLayout.pipeline_type_to_dname_map[pipeline_type]
             )
             for fpath_config in sorted(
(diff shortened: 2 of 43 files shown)


nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
+            self.removeHandler(self.stdout_handler)
+        # stdout: INFO and WARNING
+        # If verbosity is enabled, also display DEBUG
+        log_level = logging.DEBUG if verbosity else logging.INFO
+        self.stdout_handler = rich_handler(log_level, console=CONSOLE_STDOUT)
+        self.stdout_handler.addFilter(lambda record: record.levelno <= logging.WARNING)
+        self.addHandler(self.stdout_handler)
+        return self
+
+    def add_file_handler(self, file: Path) -> logging.Handler:
```
mathdugre: @michellewang , do you think we should impose a maximum of one file handler?
This would simplify the function's logic and allow to return `self` to chain commands from the class.

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
 from functools import partial
 from pathlib import Path
-from typing import Optional
 
 import rich_click as click
 from rich.logging import RichHandler
+from typing_extensions import Self
 
 from nipoppy.console import CONSOLE_STDERR, CONSOLE_STDOUT
-from nipoppy.env import IS_TESTING, LogColor, StrOrPathLike
```
michellewang: `IS_TESTING` is no longer used anywhere in the code base, so we can remove it from `nipoppy.env` too

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
+    def __init__(self, *args, **kwargs):
+        """Initialize the Nipoppy logger."""
+        super().__init__(*args, **kwargs)
+        # self.propagate = False
+        self.setLevel(logging.DEBUG)
+        # stderr: ERROR and CRITICAL
+        self.stderr_handler = rich_handler(logging.ERROR, console=CONSOLE_STDERR)
+        self.addHandler(self.stderr_handler)
+
+    def verbose(self, verbosity: bool) -> Self:
```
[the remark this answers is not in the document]
michellewang: I would prefer `set_verbose` since the arg is not a traditional verbosity level (it makes no sense to me to have `verbosity` be `True` or `False`, `verbosity` should be something like 1/2/3).

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
+    NAME: str = "nipoppy"
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the Nipoppy logger."""
+        super().__init__(*args, **kwargs)
+        # self.propagate = False
+        self.setLevel(logging.DEBUG)
+        # stderr: ERROR and CRITICAL
+        self.stderr_handler = rich_handler(logging.ERROR, console=CONSOLE_STDERR)
+        self.addHandler(self.stderr_handler)
```
michellewang: I find it not ideal to have the stderr and stdout handlers be defined in different places in the code.

At the minimum we should have a comment saying that the stdout handler is set in `verbose()`

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
+class NipoppyLogger(logging.Logger):
+    """Custom logger for Nipoppy."""
+
+    NAME: str = "nipoppy"
+
+    def __init__(self, *args, **kwargs):
+        """Initialize the Nipoppy logger."""
+        super().__init__(*args, **kwargs)
+        # self.propagate = False
+        self.setLevel(logging.DEBUG)
```
michellewang: We might want to explain that this is the level that will be used in the log files, and that stdout/stderr will have different log levels

nipoppy/logger.py
michellewang: I think that makes sense

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
+    def add_file_handler(self, file: Path) -> logging.Handler:
+        """Add a file handler to the logger."""
+        file.parent.mkdir(parents=True, exist_ok=True)
+        handler = logging.FileHandler(file)
+        handler.setFormatter(logging.Formatter(FILE_FORMAT, datefmt=DATE_FORMAT))
+        self.addHandler(handler)
+        self.info(f"Writing the log to {file}")
+        return handler
+
+    def capture_warnings(self, capture: bool = True) -> Self:
```
michellewang: I think this method name is confusing since it can turn the warnings capturing on or off instead of always to on like before. Rename to `set_capture_warnings`?

nipoppy/workflows/base.py
```diff
@@ -343,10 +340,19 @@ def run_setup(self):
...
+            self._log_handler = logger.add_file_handler(self.generate_fpath_log())
 
         super().run_setup()
 
+    def run_cleanup(self):
+        """Run the cleanup part of the workflow."""
+        # remove log file handler
+        if hasattr(self, "_log_handler"):
+            self._log_handler.close()
+            logger.removeHandler(self._log_handler)
```
michellewang: This is the part that will be in the `NipoppyLogger` class if we only allow a single log file, right?

The only possible drawback I can think of is: if for any reason a user will want to use their own log file, then they will be prevented to do so. A silly workaround would be to have a wrapper class `NipoppyFileHandler` which would not conflict with the built-in `FileHandler`

nipoppy/logger.py
michellewang: See my other comment in `BaseWorkflow`

nipoppy/workflows/pipeline_store/search.py
```diff
@@ -7,11 +7,14 @@
...
 from nipoppy.console import _INDENT, CONSOLE_STDOUT
-from nipoppy.env import ZENODO_COMMUNITY_ID, LogColor
+from nipoppy.env import ZENODO_COMMUNITY_ID
+from nipoppy.logger import LogColor, get_logger
 from nipoppy.utils.html import strip_html_tags
 from nipoppy.workflows.base import BaseWorkflow
 from nipoppy.zenodo_api import ZenodoAPI
 
+logger = get_logger()
+
```
michellewang: Nitpick: I would put this after the uppercase constant definitions

nipoppy/workflows/pipeline_store/search.py
```diff
@@ -55,8 +58,6 @@ def __init__(
         self.community = community
         self.size = size
 
-        self.zenodo_api.set_logger(self.logger)
-
```
michellewang: I don't see anywhere where the Nipoppy logger is being passed to the Zenodo API. Is that normal?

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -37,8 +40,6 @@ def __init__(
             dry_run=dry_run,
         )
 
-        self.zenodo_api.set_logger(self.logger)
-
```
michellewang: Same comment as for `PipelineSearchWorkflow`

tests/unit/test_logger.py
```diff
@@ -4,22 +4,33 @@
...
-def test_get_logger(verbose: bool, name: str):
-    logger = get_logger(verbose=verbose, name=name)
-    assert isinstance(logger, logging.Logger)
-    assert logger.name == name
+def test_get_logger(verbose: bool, logger, monkeypatch: pytest.MonkeyPatch):
+    logger.verbose(verbose)
+    assert logger.name == PROGRAM_NAME
+    assert logger.level == logging.DEBUG
+
+    # assert False, logger.handlers
```
michellewang:
```suggestion
```

tests/unit/test_logger.py
```diff
@@ -4,22 +4,33 @@
...
+    assert isinstance(logger.handlers[0], rich.logging.RichHandler)
+    assert logger.handlers[0].level == logging.ERROR
+
+    assert isinstance(logger.handlers[1], rich.logging.RichHandler)
+    if verbose:
+        assert logger.handlers[1].level == logging.DEBUG
+    else:
+        assert logger.handlers[1].level == logging.INFO
```
michellewang: We could add the stdout handler log level as part of the test parametrization to avoid reimplementing logic inside the test

tests/unit/test_logger.py
```diff
@@ -45,44 +56,19 @@ def test_get_logger_capsys(verbose: bool, capsys: pytest.CaptureFixture):
...
-    fpath_log = tmp_path / "test.log"
-    add_logfile(logger, fpath_log)
-    logger.info("Test")
-    assert fpath_log.exists()
-
-
-def test_add_logfile_mkdir(tmp_path: Path, caplog: pytest.LogCaptureFixture):
-    logger = logging.getLogger("test_add_logfile_mkdir")
+@pytest.mark.no_xdist
+def test_add_file_handler(tmp_path: Path, logger):
```
michellewang: Why remove the part checking that the parent directory of the logfile is created?

tests/unit/test_logger.py
michellewang: Missing test(s) for `capture_warnings`

michellewang (changes requested):
It might be useful to have a third verbosity level for only showing errors. Though that would need to be handled in a future PR because it would involve (breaking) changes in the CLI

tests/unit/workflows/test_base.py
```diff
@@ -28,12 +28,11 @@ def test_abstract_class():
 def test_init(workflow: BaseWorkflow):
     assert workflow.name == "my_workflow"
     assert workflow.return_code == 0
-    assert isinstance(workflow.logger, logging.Logger)
 
 
-@pytest.mark.no_xdist
 @pytest.mark.parametrize("command", ["echo x", "echo y"])
 @pytest.mark.parametrize("prefix_run", ["[RUN]", "<run>"])
+@pytest.mark.no_xdist
```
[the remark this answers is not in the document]
mathdugre: Mostly for cases where the parametrization is very long, making it hard to notice the no_xdist.

e.g.
https://github.com/mathdugre/nipoppy/blob/d72b237726df40cdfead201ff254c79a95a8aa0b/tests/unit/workflows/test_pipeline.py#L1279-L1310

nipoppy/logger.py
```diff
@@ -1,113 +1,130 @@
...
-    # If verbosity is enabled, also display DEBUG
-    verbosity = logging.DEBUG if verbose else logging.INFO
-    stdout_handler = rich_handler(verbosity, console=CONSOLE_STDOUT)
-    stdout_handler.addFilter(lambda record: record.levelno <= logging.WARNING)
-    logger.addHandler(stdout_handler)
-
-    # Add custom method for SUCCESS level
-    setattr(logger, "success", types.MethodType(success, logger))
+CONSOLE_FORMAT = "%(message)s"
+FILE_FORMAT = "%(asctime)s %(levelname)-7s %(message)s"
```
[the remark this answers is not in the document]
mathdugre: Which logging levels would you want?

tests/unit/workflows/test_pipeline.py
```diff
@@ -1647,16 +1657,20 @@ def write_error_file(*args, **kwargs):
...
     workflow: PipelineWorkflow,
     mocker: pytest_mock.MockFixture,
     caplog: pytest.LogCaptureFixture,
     job_id,
 ):
-    mocked = _set_up_hpc_for_testing(workflow, mocker)
-    mocked.return_value = job_id
+    import logging
+
+    with caplog.at_level(logging.INFO, logger="nipoppy"):
```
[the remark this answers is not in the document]
mathdugre: No longer needed. Good catch

nipoppy/logger.py
michellewang: Yes I think that's a good idea

nipoppy/workflows/pipeline_store/list.py
```diff
@@ -63,20 +66,16 @@ def _log_pipeline_info(
...
-            self.logger.warning(f"No available {pipeline_type.value} pipelines")
+            logger.warning(f"No available {pipeline_type.value} pipelines")
             return
 
-        self.logger.info(
-            f"[{LogColor.SUCCESS}]"
-            f"Available {pipeline_type.value} pipelines and versions"
-            "[/]"
-        )
+        logger.success(f"Available {pipeline_type.value} pipelines and versions")
```
michellewang: I don't think it makes sense to have the 🎉 emojis here, maybe we should just use `INFO` level

nipoppy/workflows/pipeline_store/search.py
michellewang: I'm not sure I understand

nipoppy/workflows/pipeline_store/list.py
mathdugre: Also, what do you think of the emojis for warnings and failures? Maybe we should stick to yellow and red without emojis?

I realized they don't look great on my terminal (when repeated)
Screenshot 2025-11-20 at 21 01 58

michellewang: Warning emojis also overlap for me, red Xs are fine. I would be okay with a single emoji, it does make them pop up more

nipoppy/workflows/base.py
```diff
@@ -343,10 +340,14 @@ def run_setup(self):
             self.layout.validate()
 
         if not self._skip_logfile:
-            add_logfile(self.logger, self.generate_fpath_log())
+            self._log_handler = logger.add_file_handler(self.generate_fpath_log())
```
michellewang: Do we still need to set this to `self._log_handler`?

nipoppy/workflows/base.py
```diff
@@ -343,10 +340,14 @@ def run_setup(self):
...
 
         super().run_setup()
 
+    def run_cleanup(self):
+        """Run the cleanup part of the workflow."""
+        return super().run_cleanup()
+
```
michellewang: I don't think this is needed here

nipoppy/zenodo_api.py
```diff
@@ -52,9 +47,18 @@ def set_authorization(self, access_token: str):
...
         self.headers.update({"Authorization": f"Bearer {access_token}"})
 
-    def set_logger(self, logger: logging.Logger):
+    @property
+    def logger(self) -> logging.Logger:
+        """Get the logger for the ZenodoAPI instance."""
+        return self._logger
+
+    @logger.setter
+    def logger(self, logger: logging.Logger):
```
michellewang: Would this be more accurate for the type annotation?

```suggestion
    def logger(self, logger: logging.Logger | None):
```

michellewang (changes requested):

nipoppy/workflows/runner.py
```diff
@@ -51,7 +54,7 @@ def launch_boutiques_run(
         # process the descriptor if it containers Nipoppy-specific placeholder
         # expressions (legacy behaviour)
         if TEMPLATE_REPLACE_PATTERN.search(self.descriptor["command-line"]):
-            self.logger.info("Processing the JSON descriptor")
+            logger.info("Processing the JSON descriptor")
```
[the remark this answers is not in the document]
mathdugre: Maybe we can decide on the different log levels in another PR?

nipoppy/pipeline_validation.py
```diff
@@ -228,34 +241,22 @@ def _log(msg: str) -> None:
...
-    log_level=logging.DEBUG,
 ) -> None:
     """Check that all files are within the bundle directory."""
     dpath_bundle: Path = Path(dpath_bundle).resolve()
-    if logger is not None:
-        logger.log(
-            level=log_level,
-            msg="Checking that all files are within the bundle directory",
-        )
+    logger.debug("Checking that all files are within the bundle directory")
```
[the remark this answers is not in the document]
mathdugre: While I see the value of having the file validated log to INFO, for this one—and the one below—I think they should remain as debug.

mathdugre:
I'm not sure how we can properly test the warning capture.

I've made this test
```python
def test_warning_capture(
    caplog: pytest.LogCaptureFixture,
    recwarn: pytest.WarningsRecorder,
):
    import warnings

    get_logger()

    warning_msg = "This is a test warning."
    warnings.warn(warning_msg)

    # No warning should be recorded
    assert len(recwarn) == 1

    # Warning should be captured as a log record
    assert len(caplog.records) == 1
    assert caplog.records[0].levelno == logging.WARNING
    assert warning_msg in caplog.records[0].message
```

It fails within the test suite, but pass if I create a new project with only the test and the custom logger.

Any ideas how to debug this further?

nipoppy/workflows/pipeline_store/search.py
```diff
@@ -54,6 +54,7 @@ def __init__(
             dry_run=dry_run,
         )
         self.zenodo_api = zenodo_api or ZenodoAPI()
+        self.zenodo_api.logger = logger  # use nipoppy logger configuration
```
michellewang: We need the same thing for `PipelineUploadWorkflow` and `PipelineInstallWorkflow`

PR #913, opened 2026-03-05
[MAINT] Extract services from workflow classes to remove coupling

### LINKED ISSUE #889: [MAINT] Refactor HPC CLI command generation
There's quite a lot of repetition between the `_generate_cli_command_for_hpc` function in the different Runner classes:
* https://github.com/mathdugre/nipoppy/blob/7c50192874875defdb56ddc27ed0777e0aa129c2/nipoppy/workflows/extractor.py#L120-L151
* https://github.com/mathdugre/nipoppy/blob/7c50192874875defdb56ddc27ed0777e0aa129c2/nipoppy/workflows/processing_runner.py#L144-L176
* https://github.com/mathdugre/nipoppy/blob/7c50192874875defdb56ddc27ed0777e0aa129c2/nipoppy/workflows/bids_conversion.py#L79-L109

We should refactor to avoid duplication.

### LINKED ISSUE #907: [MAINT] Remove redundant method `BaseWorkflow._save_tabular_file`
We could remove this function to directly call the `save_with_backup` method.

https://github.com/nipoppy/nipoppy/blob/c4580e572c258924ba0233c6d6d1ef108513dc38/nipoppy/workflows/base.py#L142-L148

mathdugre:
- Closes #889
- Closes #907

- Add a workflow context to facilitate passing argument across the workflow methods
- Extract service for
  - Handling boutiques exec and simulate
  - Handling HPC command gen and submission

## Checklist (for reviewers)

- [x] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--913.org.readthedocs.build/en/913/


nipoppy/workflows/services/__init__.py
```diff
@@ -0,0 +1 @@
+"""Workflow services for Nipoppy."""
```
michellewang: Would `utils` be better?

nipoppy/workflows/services/boutiques.py
```diff
@@ -0,0 +1,148 @@
...
+    ----------
+    context : WorkflowContext
+        The shared workflow context.
+    descriptor : dict
+        The Boutiques descriptor for the container.
+    """
+
+    def __init__(self, context: WorkflowContext, descriptor: dict):
+        self.context = context
+        self.descriptor = descriptor
```
michellewang: Why do we have `descriptor` as an init argument + `descriptor_str` as an argument to `run`? `self.descriptor` does not appear to be used

nipoppy/workflows/services/boutiques.py
```diff
@@ -0,0 +1,148 @@
...
+
+        try:
+            self._execute(command, run_command, dry_run)
+        except subprocess.CalledProcessError as exception:
+            error_message = self._get_error_message(exit_code=exception.returncode)
+            raise ExecutionError(
+                f"{error_message} (return code: {exception.returncode})"
+            )
+
+        if self.mode == "Simulating" and bosh_exec_launch_args:
```
michellewang: I think this kind of pattern is not best practice since we should not be using knowledge about a subclass in the parent class. Alternatives would be to use a hook or override this in the subclass, what do you think?

nipoppy/workflows/services/boutiques.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, context: WorkflowContext, descriptor: dict):
+        self.context = context
+        self.descriptor = descriptor
+
+    def run(
+        self,
+        invocation_str: str,
+        descriptor_str: str,
+        bosh_exec_launch_args: list[str] | None = None,
+        run_command: subprocess.run | None = None,
```
michellewang: The type for `run_command` should be `Callable` instead of `subprocess.run`, no?

nipoppy/workflows/services/boutiques.py
```diff
@@ -0,0 +1,149 @@
...
+from collections.abc import Callable
+
+from nipoppy.exceptions import ExecutionError
+from nipoppy.logger import get_logger
+from nipoppy.workflows.services.context import WorkflowContext
+
+logger = get_logger()
+
+
+class BoshRunner:
```
michellewang: Would it make more sense to call this `BoshLaunch` instead? For consistency with `BoshSimulate`

Possibly `BoshLaunchRunner` and `BoshSimulateRunner`. In which case we could rename `services` to `runners` (if and only if `processing_runner.py` etc. are renamed in the future)

nipoppy/workflows/runner.py
```diff
@@ -38,6 +39,44 @@ def __init__(
...
 
+    @cached_property
+    def container_runner(self) -> BoshRunner:
+        """Get the container runner service."""
+        if self.simulate:
+            return BoshSimulate(
+                context=self.workflow_context,
+                descriptor=self.descriptor,
+            )
+        return BoshRunner(
```
michellewang: I fell it would be better to have the `if`/`else` branching set the class, and then do `return bosh_runner_class(context=..., descriptor=...)`

michellewang (changes requested):
I also requested a review from Copilot, not sure if the comments are useful or not but sometimes it has caught things that I missed.

nipoppy/workflows/services/hpc.py
```diff
@@ -0,0 +1,259 @@
...
+        if len(job_args) == 0:
+            logger.warning("HPC configuration is empty")
+
+        template_ast = Environment().parse(FPATH_HPC_TEMPLATE.read_text())
+        template_vars = meta.find_undeclared_variables(template_ast)
+        missing_vars = set(job_args.keys()) - template_vars
+        if len(missing_vars) > 0:
+            logger.warning(
+                "Found variables in the HPC config that are not used in the template "
+                f"job script: {missing_vars}. Update the config or modify the template "
```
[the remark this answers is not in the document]
michellewang: @mathdugre I think that's a valid comment, `unused_vars` is a better name

tests/unit/workflows/services/test_boutiques.py
```diff
@@ -0,0 +1,65 @@
...
+                "id": "input_file",
+                "name": "Input File",
+                "type": "File",
+                "value-key": "[INPUT]",
+            }
+        ],
+    }
+
+
+def test_run_bosh_launch(bosh_descriptor, mocker):
```
michellewang: Add type annotations?

nipoppy/workflows/runner.py
```diff
@@ -38,6 +40,39 @@ def __init__(
         self.simulate = simulate
         self.keep_workdir = keep_workdir
 
+    @cached_property
+    def bosh_runner(self) -> Callable[..., int]:
+        """Get the container runner service."""
+        if self.simulate:
+            return run_bosh_simulate
+        else:
+            return run_bosh_launch
```
michellewang: This is technically no longer a "runner", rename `bosh_runner`?

michellewang (changes requested):
Main comments:
- Not sure having the Boutiques part being functions instead of classes is truly better
- Tests need deeper look

nipoppy/workflows/services/hpc.py
```diff
@@ -0,0 +1,291 @@
...
+            if option in default_options:
+                raise ValueError(
+                    f"Option {option} is already set by the default options and cannot "
+                    "be overridden via extra_options."
+                )
+            command.extend([option, str(value)])
+
+        extra_flags = extra_flags or []
+        for flag in extra_flags:
+            command.append(flag)
```
michellewang: Should we have a check that `extra_flags` does not contain something that is already in `command`?

nipoppy/workflows/pipeline.py
```diff
@@ -695,188 +690,70 @@ def apply_analysis_level(
 
     def run_main(self):
         """Run the pipeline."""
+        participants_sessions = self._prepare_participants_sessions()
+
+        if self.write_subcohort is not None:
+            self._handle_write_subcohort(participants_sessions)
```
michellewang: "Handle" is vague. Can rename argument to `output_subcohort_path`/`input_subcohort_path` or something else if you can think of a better name.

tests/unit/workflows/services/test_boutiques.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@pytest.mark.parametrize(
+    "bosh_func,error_msg",
+    [
+        (run_bosh_launch, "Pipeline execution failed"),
+        (run_bosh_simulate, "Pipeline simulation failed"),
+    ],
+)
+def test_bosh_launch_capture_error(
```
michellewang:
```suggestion
def test_run_bosh_func_capture_error(
```

michellewang (changes requested):

nipoppy/workflows/pipeline.py
```diff
@@ -695,188 +690,70 @@ def apply_analysis_level(
 
     def run_main(self):
         """Run the pipeline."""
+        participants_sessions = self._prepare_participants_sessions()
```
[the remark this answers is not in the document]
mathdugre: After discussion with @michellewang, I reworked the run_main method to do the load, filter, analysis_level, then handle the execution strategy seperately.

This change remove the method to rename.

nipoppy/workflows/runner.py
```diff
@@ -207,3 +257,14 @@ def process_container_config(
         )
 
         return container_command, container_handler
+
+    def run_main(self):
+        """Run the pipeline."""
```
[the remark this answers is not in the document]
michellewang: @mathdugre tests are failing because `override` is not available in Python 3.10

michellewang (changes requested):
Mostly minor comments, I think we are close :)

michellewang (changes requested):
We are missing code coverage in `generate_cli_command`

Image

michellewang (approved):
🧑‍🍳

PR #923, opened 2026-03-16
[ENH] Avoid DICOM filename collisions by prefixing filenames with the parent directory during reorg

### LINKED ISSUE #694: [ENH] Recognize DICOM files with identical filenames within different sequence directories

### New feature

Hello maintainers, thank you for building a tool that makes curating datasets less tedious! 

I recently started to integrate nipoppy for curation and processing of imaging data, and I came across`"Cannot move file because it already exists"` error with the `nipoppy reorg` command. The DICOM files I'm working with are named identically across different sequences. For example, `sub-A/ses-1` have the following (partial) directory tree:

```bash
sub-A/ses-1
├── S10010
│   ├── I10
│   └── I11
├── S1010
│   ├── I10
│   └── I11
├── S11010
│   ├── I10
│   └── I11
├── S12010
│   ├── I10
│   └── I11
├── S4010
│   ├── I10
│   ├── I100
│   ├── I1000
│   ├── I10000
│   ├── I100000
│   ├── I100010
│   ├── I100020
│   ├── I100030
│   ├── I100040
│   ├── 100050
│   ├── I100060

```

With help from @michellewang on discord, I was able to resolve it by renaming the files to be unique by prefix-ing the sequence directory name as follows:

```bash
sub-A/ses-1
├── S10010
│   ├── S10010_S10010_I10
│   └── S10010_S10010_I11
├── S1010
│   ├── S1010_I10
│   └── S1010_I11
├── S11010
│   ├── S11010_I10
│   └── S11010_I11
├── S12010
│   ├── S12010_I10
│   └── S12010_I11
├── S2010
│   ├── S2010_I10
│   └── S2010_I11
├── S3010
│   ├── S3010_I10
│   └── S3010_I11
├── S4010
│   ├── S4010_I10
│   ├── S4010_I100
│   ├── S4010_I1000
│   ├── S4010_I10000
│   ├── S4010_I100000
│   ├── S4010_I100010
│   ├── S4010_I100020
│   ├── S4010_I100030
│   ├── S4010_I100040
│   ├── S4010_I100050
│   ├── S4010_I100060
```
I'd be happy to make the necessary contributions to the codebase to include this renaming feature if y'all think its a good idea. I'm not sure how common this issue is but our scanners at NIMHANS seem to be set up this way. 

Thanks! 

### Unclear documentation

_No response_

mibur1:
This PR updates DICOM reorganization so files from different sequence directories no longer collide when they share the same filename.

- Closes #694

- Make `apply_fname_mapping()` prefix filenames with the parent directory during DICOM reorg to avoid collisions across series
- Narrow `fpath_source` from `StrOrPathLike` to `Path`
- Update the corresponding unit tests

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--923.org.readthedocs.build/en/923/


nipoppy/workflows/dicom_reorg.py
```diff
@@ -86,16 +86,14 @@ def get_fpaths_to_reorg(
...
-        Apply a mapping from the original (full) file path to destination file name.
+        Append the parent directory name to the DICOM files.
 
-        This method does not change the file name by default, but it can be overridden
-        if the file names need to be changed during reorganization (e.g. for easier
-        BIDS conversion).
+        This helps to avoid filename collisions across DICOM series.
         """
-        return Path(fpath_source).name
+        return f"{fpath_source.parent.name}_{fpath_source.name}"
```
michellewang: This solution works for some cases but could fail with more deeply nested directory structures:

```
pre_reorg/
└── 001/
    └── A/
        ├── protocol1/
        │   └── DICOM/
        │       └── 00000001.dcm
        └── protocol2/
            └── DICOM/
                └── 00000001.dcm
```

A solution would be to compute the hash of the full parent path, and prepend that instead:

```python
import hashlib

HASH_LENGTH = 7
hash = hashlib.md5(str(fpath_source.parent).encode("UTF-8")).hexdigest()
return f"{hash[:HASH_LENGTH]}_{fpath_source.name}"
```

This should reasonably avoid collisions while prevent the file names from getting too long. Files copied from the same folder will have the same hash. The caveat would be that we won't be able to tell which directory a file came from, but this is not a huge issue in my opinion given that the DICOM reorg step is for preparation for BIDSification. But let me know what you think.

michellewang (changes requested):
Thanks @mibur1!

I proposed an alternative solution (suggested by @mathdugre) that could prevent/reduce collisions with more deeply nested directories. Let me know if you have questions about this or any other of my comments.

michellewang (changes requested):
Thanks @mibur1! I think one more pass then we should be good to go :)

The pre-commit format changes are fine, it's because we have a line length limit of 88 characters

michellewang (approved):
Looks good, thanks @mibur1! 🎉

Issue #922, opened 2026-03-12
[MAINT] File organization and Public API


mathdugre:
We should make a better separation between the public and private API. This would make a distinction on:
- What we can add/remove args without breaking changes
- What function requires more documentation

We should structure the files so the import hierarchy makes sense. 

Organize the files more logically instead of having them at the root.

No longer have "utils" generic name files.
Similar for `base.py`, which is currently present three times in the codebase.

michellewang:
Closing because we have decided to address this in the future after discussing more in-depth on the whole code architecture/design. Maybe a milestone would be helpful

PR #953, opened 2026-04-01
[MAINT] Fix #888 Extract Tar handling to a separate component

### LINKED ISSUE #888: [MAINT] Decouple Tar functionality
The tar features are highly coupled with the `ProcessingRunner` class. We should extract those functions to facilitate the redesign of the workflow classes.

mathdugre:
- Closes #888

- Move tar logic from ProcessingRunner to a new module (`nipoppy.workflows.tar_handler`). Note: this would be moved to new location in https://github.com/nipoppy/nipoppy/issues/922

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--953.org.readthedocs.build/en/953/

### DIFF
--- a/nipoppy/workflows/processing_runner.py
+++ b/nipoppy/workflows/processing_runner.py
@@ -2,16 +2,15 @@
 
 from functools import cached_property
 from pathlib import Path
-from tarfile import is_tarfile
 from typing import Optional
 
 from nipoppy.config.tracker import TrackerConfig
-from nipoppy.env import EXT_TAR, PROGRAM_NAME, StrOrPathLike
-from nipoppy.exceptions import ConfigError, FileOperationError
+from nipoppy.env import PROGRAM_NAME, StrOrPathLike
+from nipoppy.exceptions import FileOperationError
 from nipoppy.logger import get_logger
 from nipoppy.utils import fileops
-from nipoppy.workflows.base import _run_command
 from nipoppy.workflows.runner import Runner
+from nipoppy.workflows.tar_handler import TarHandler
 
 logger = get_logger()
 
@@ -56,6 +55,7 @@ class ProcessingRunner(Runner):
             keep_workdir=keep_workdir,
         )
         self.tar = tar
+        self._tar_handler = TarHandler(dry_run=self.dry_run)
 
     @cached_property
     def dpaths_to_check(self) -> list[Path]:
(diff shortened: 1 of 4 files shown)


nipoppy/workflows/processing_runner.py
```diff
@@ -71,48 +71,22 @@ def _check_tar_conditions(self):
...
-
-        # make sure that the tarfile was created successfully before removing
-        # original directory
-        if fpath_tarred.exists() and is_tarfile(fpath_tarred):
-            fileops.rm(dpath, dry_run=self.dry_run)
-        else:
-            logger.error(f"Failed to tar {dpath} to {fpath_tarred}")
-
-        return fpath_tarred
+        self._tar_handler.dry_run = self.dry_run
```
michellewang, line 88: This is not needed
```suggestion
```

nipoppy/workflows/processing_runner.py
```diff
@@ -56,6 +55,7 @@ def __init__(
             keep_workdir=keep_workdir,
         )
         self.tar = tar
+        self._tar_handler = TarHandler(dry_run=self.dry_run)
```
michellewang, line 58: Do we prefer having `_tar_handler` as an attribute or a property? It should be property if we want to be strict with the "`__init__` does nothing but set attributes" convention

nipoppy/workflows/processing_runner.py
michellewang: And, since we remove it, I think we can remove the `tar_directory()` method entirely and directly call `_tar_handler.tar_directory()` instead

nipoppy/workflows/tar_handler.py
```diff
@@ -0,0 +1,69 @@
+"""TAR-related workflow helper."""
```
michellewang, line 1: Nitpick: don't write `TAR` in all caps

nipoppy/workflows/tar_handler.py
```diff
@@ -0,0 +1,69 @@
...
+from nipoppy.exceptions import ConfigError, FileOperationError
+from nipoppy.logger import get_logger
+from nipoppy.utils import fileops
+from nipoppy.workflows.base import _run_command
+
+logger = get_logger()
+
+
+class TarHandler:
+    """Handle TAR validation and directory archiving."""
```
michellewang:
```suggestion
    """Handle tar validation and directory archiving."""
```

nipoppy/workflows/processing_runner.py
```diff
@@ -71,48 +71,22 @@ def _check_tar_conditions(self):
...
-                "The PARTICIPANT_SESSION_DIR field in the tracker config must set "
-                "in the tracker config file at "
-                f"{self.pipeline_step_config.TRACKER_CONFIG_FILE}"
-            )
+        tracker_config_file = self.pipeline_step_config.TRACKER_CONFIG_FILE
+        participant_session_dir = (
+            self.tracker_config.PARTICIPANT_SESSION_DIR
+            if self.tar and tracker_config_file is not None
+            else None
+        )
```
michellewang, line 79: This is not good because we need to throw the error immediately if `self.pipeline_step_config.TRACKER_CONFIG_FILE` is `None`

Otherwise `self.tracker_config` will not work

So I think we need to discuss this further because I can't see a clean way to extract the validation code entirely

nipoppy/workflows/tar_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def __init__(self, dry_run: bool = False):
```
michellewang, line 19: I think this should take `tar_requested` as input as well. Then the validate/tarring methods can do nothing if `self.tar_requested is None`

nipoppy/workflows/tar_handler.py
```diff
@@ -0,0 +1,69 @@
...
+
+        _run_command(
+            f"tar {tar_flags} {fpath_tarred} -C {dpath.parent} {dpath.name}",
+            dry_run=self.dry_run,
+        )
+
+        if fpath_tarred.exists() and is_tarfile(fpath_tarred):
+            fileops.rm(dpath, dry_run=self.dry_run)
+        else:
+            logger.error(f"Failed to tar {dpath} to {fpath_tarred}")
```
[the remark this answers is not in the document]
michellewang: I think this is valid (though minor)

nipoppy/workflows/tar_handler.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, dry_run: bool = False):
+        self.dry_run = dry_run
+
+    def validate_preconditions(
+        self,
+        tar_requested: bool,
+        tracker_config_file: Optional[StrOrPathLike],
+        participant_session_dir: Optional[StrOrPathLike] = None,
+    ):
+        """Validate TAR-related configuration requirements."""
```
michellewang, line 28: All caps "TAR" feels aggressive

nipoppy/workflows/processing_runner.py
```diff
@@ -71,48 +71,22 @@ def _check_tar_conditions(self):
...
-            raise ConfigError(
-                "Tarring requested but no participant-session directory specified. "
-                "The PARTICIPANT_SESSION_DIR field in the tracker config must set "
-                "in the tracker config file at "
-                f"{self.pipeline_step_config.TRACKER_CONFIG_FILE}"
-            )
+        tracker_config_file = self.pipeline_step_config.TRACKER_CONFIG_FILE
+        participant_session_dir = (
+            self.tracker_config.PARTICIPANT_SESSION_DIR
+            if self.tar and tracker_config_file is not None
```
michellewang, line 77: The `if self.tar` is not needed

nipoppy/workflows/processing_runner.py
michellewang, line 68: If the point is to decouple, shouldn't we use `self._tar_handler.validate_preconditions()` directly instead of a wrapper method `_check_tar_conditions()`?

nipoppy/workflows/tar_handler.py
michellewang, line 28: Can we keep the same docstring as before? It's more specific than this new one

tests/unit/workflows/test_tar_handler.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_validate_preconditions_no_tracker_config(tar_handler: TarHandler):
+    with pytest.raises(
+        ConfigError,
+        match="Tarring requested but there is no tracker config file",
+    ):
+        tar_handler.validate_preconditions(
+            tar_requested=True,
+            tracker_config_file=None,
+            participant_session_dir="sub-01/ses-1",
```
michellewang, line 26: This is an impossible scenario because if `tracker_config_file` is `None`, then there is no way to get `participant_session_dir`

michellewang (changes requested):
I am not convinced that it makes sense for the `TarHandler` class to be handling config validation. Other comments are mostly minor, though I do think it makes sense for `TarHandler.__init__` to take in `tar_requested` as input

nipoppy/workflows/processing_runner.py
mathdugre: Sounds good, will switch it to property

nipoppy/workflows/processing_runner.py
mathdugre: Good point

mathdugre:
Closing in favor of the reimplementation in #991

PR #978, opened 2026-04-10
[ENH] Add OpenTelemetry usage telemetry to nipoppy CLI

blackpearl006:
## Summary

- Adds opt-out usage telemetry to the nipoppy CLI tracking command execution, completion status and geographic distribution of installations
- Implements a server-side observability stack (OTel Collector → Prometheus → Grafana) 

## What gets tracked

| Metric | Recorded when |
|--------|---------------|
| `nipoppy_commands_executed_total{command}` | Any command starts |
| `nipoppy_commands_completed_total{command, status, return_code}` | Any command exits |
| `nipoppy_location_by_country_installations_total{country}` | `nipoppy init` only |

Opt out at any time with `export OTEL_SDK_DISABLED=true`.

## To run

- [ ] `pip install -e .` installs all OTel dependencies automatically
- [ ] Run any nipoppy command with `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317` — counter should increment on each run
- [ ] `export OTEL_SDK_DISABLED=true` — nipoppy runs normally with no metrics sent
- [ ] Start server stack: `cd server && docker compose up -d`, open Grafana at http://localhost:3000

----
📚 Documentation preview 📚: https://nipoppy--978.org.readthedocs.build/en/978/


pyproject.toml
```diff
@@ -27,6 +27,12 @@ dependencies = [
     "rich",
     "rich-click",
     "typing-extensions",
+    # OpenTelemetry for metrics (optional - gracefully handled if missing)
+    "opentelemetry-api>=1.20.0",
+    "opentelemetry-sdk>=1.20.0",
+    "opentelemetry-exporter-otlp-proto-grpc>=1.20.0",
+    # GeoIP for location tracking (optional)
+    "requests",
```
michellewang, line 35: Is it possible to use `httpx` instead of `requests`?

nipoppy/telemetry/geo.py
```diff
@@ -0,0 +1,64 @@
...
+
+Performs a one-time GeoIP lookup during `nipoppy init` and records the
+country code as a telemetry metric.
+"""
+
+try:
+    import requests
+    _REQUESTS_AVAILABLE = True
+except ImportError:
+    _REQUESTS_AVAILABLE = False
```
michellewang, line 12: We can use `httpx` instead of `requests` here, and we don't need `_REQUESTS_AVAILABLE` as we always assume that all required dependencies are available

nipoppy/cli/__init__.py
```diff
@@ -26,7 +26,7 @@ def exception_handler(workflow):
         workflow.return_code = ReturnCode.INVALID_CONFIG
         logger.error(e)
     except SystemExit as e:
-        workflow.return_code = e.code or ReturnCode.UNKNOWN_FAILURE
+        workflow.return_code = e.code if e.code is not None else ReturnCode.UNKNOWN_FAILURE
```
michellewang, line 29: Could you explain why this change is needed?

nipoppy/cli/cli.py
```diff
@@ -30,6 +30,22 @@ def decorator(f):
...
+    )
+    _TELEMETRY_AVAILABLE = True
+except ImportError:
+    _TELEMETRY_AVAILABLE = False
+    def track_command(name):
+        return lambda f: f
+
+
+if _TELEMETRY_AVAILABLE:
+    initialize_telemetry(service_version=__version__)
```
michellewang, line 47: I think we can assume that telemetry dependencies are always available

nipoppy/telemetry/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""
+OpenTelemetry telemetry for Nipoppy CLI.
+
+Two independent features:
+1. Command tracking: which commands are used and how often (@track_command)
+2. Location tracking: geographic distribution of installations (record_location)
+"""
+
+from nipoppy.telemetry.decorators import track_command
```
michellewang, line 9: We prefer using absolute import paths inside the codebase itself. Relative paths are only used for the public API right now

So this `__init__.py` can be empty and the import paths from other files should be changed

nipoppy/cli/cli.py
```diff
@@ -137,13 +153,16 @@ def cli():
 )
 @global_options
 @layout_option
+@track_command("init")  # ← CLIENT-SIDE: Track init command
```
michellewang, line 156: The comment here is not that useful in my opinion, can be removed

nipoppy/cli/pipeline_catalog.py
```diff
@@ -16,6 +16,12 @@
 from nipoppy.env import PipelineTypeEnum
 from nipoppy.zenodo_api import ZenodoAPI
 
+try:
+    from nipoppy.telemetry import track_command
+except ImportError:
+    def track_command(name):
+        return lambda f: f
```
michellewang, line 23: Same comment as for `cli.py`: we can remove the `try`/`except` here

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+- Providing thread-safe access to metrics
+"""
+
+# Shutdown pattern reference:
+# https://oneuptime.com/blog/post/2026-02-06-otel-sdk-shutdown-python-atexit-sigterm/view
+import atexit
+import os
+import signal
+import sys
+from typing import Optional, Dict
```
michellewang, line 16: We learned recently that the best practice for typing is to use e.g. `str | None` instead of `Optional[str]`, and `dict` instead of `Dict`, so moving forward we'll enforce this in the codebase

nipoppy/telemetry/decorators.py
michellewang, line 1: To discuss: would it be useful/better to have a `TelemetryHandler` class?

So that if code is run without using the CLI (which we will allow in the future) we will still have telemetry

Also because it would be useful (in a future PR) to configure telemetry from the global config file, not just the OTEL environment variable

nipoppy/telemetry/decorators.py
```diff
@@ -0,0 +1,153 @@
...
+                return_code = e.code if isinstance(e.code, int) else ReturnCode.UNKNOWN_FAILURE
+                status = _return_code_to_status(return_code)
+                raise
+            except Exception:
+                # Fallback: exception escaped before exception_handler could convert it
+                status = "failure"
+                return_code = ReturnCode.UNKNOWN_FAILURE
+                raise
+            finally:
+                _record_completion_metric(command_name, status, return_code)
```
michellewang, line 71: `_record_completion_metric` could record the exit code directly instead of a string status, since the exit code is more precise

nipoppy/telemetry/decorators.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    if code == ReturnCode.SUCCESS:
+        return "success"
+    elif code in (ReturnCode.PARTIAL_SUCCESS, ReturnCode.NO_PARTICIPANTS_OR_SESSIONS_TO_RUN):
+        return "partial"
+    else:
+        return "failure"
+
+
+def _record_command_metric(command_name: str) -> None:
```
michellewang, line 97: Is it redundant to have both command start and command completion metrics?

nipoppy/telemetry/geo.py
```diff
@@ -0,0 +1,64 @@
+"""
+Geographic location tracking.
+
+Performs a one-time GeoIP lookup during `nipoppy init` and records the
```
michellewang, line 4: `nipoppy init` should not be referred here since the functions are more general

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    if _SHUTDOWN_CALLED:
+        return
+    _SHUTDOWN_CALLED = True
+    if _PROVIDER is not None:
+        _PROVIDER.shutdown()
+
+
+def initialize_telemetry(
+    service_name: str = "nipoppy",
+    service_version: str = "1.0.0",
```
michellewang, line 44: Is this meant to be the Nipoppy version? If yes this should be changed to dynamically get the version

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Returns:
+        True if initialized successfully, False otherwise
+    """
+    global _METER, _METRICS, _PROVIDER
+
+    if _PROVIDER is not None:
+        return True
+
+    if os.getenv("OTEL_SDK_DISABLED", "false").lower() == "true":
```
michellewang, line 69: To think about: what is the best way to disable telemetry in dev environments? Maybe a check if the version tag has `"dev"` in it, e.g. `0.4.3.dev46+g1b95898b1`

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+            else:
+                sys.exit(0)
+
+        signal.signal(signal.SIGTERM, _sigterm_handler)
+
+        return True
+
+    except Exception as e:
+        print(f"⚠ Warning: Telemetry initialization failed: {e}")
+        print("  Continuing without telemetry...")
```
michellewang, line 140: We should use the Nipoppy logger for that: `from nipoppy.logger import get_logger`

nipoppy/telemetry/decorators.py
```diff
@@ -0,0 +1,153 @@
...
+            except SystemExit as e:
+                # Extract return code from SystemExit (set by exception_handler)
+                return_code = e.code if isinstance(e.code, int) else ReturnCode.UNKNOWN_FAILURE
+                status = _return_code_to_status(return_code)
+                raise
+            except Exception:
+                # Fallback: exception escaped before exception_handler could convert it
+                status = "failure"
+                return_code = ReturnCode.UNKNOWN_FAILURE
+                raise
```
michellewang, line 69: I think this is causing commands to crash if something goes wrong with the telemetry

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    try:
+        if otlp_endpoint is None:
+            otlp_endpoint = os.getenv(
+                "OTEL_EXPORTER_OTLP_ENDPOINT",
+                "http://localhost:4317"
+            )
+
+        # Strip scheme prefix for gRPC; track whether to use TLS
+        insecure = True
+        if otlp_endpoint.startswith("https://"):
```
michellewang, line 81: Better to use `otlp_endpoint.removeprefix("https://")` instead of otlp_endpoint[len("https://"):]

nipoppy/telemetry/metrics.py
```diff
@@ -0,0 +1,198 @@
...
+        ),
+
+        # Track geographic distribution of installations
+        # Attributes: country (US, CA, IN, etc.)
+        "location_by_country": meter.create_counter(
+            name="location.by_country",
+            description="Number of installations per country",
+            unit="installations",
+        ),
+    }
```
michellewang, line 178: We might want to use a dataclass for this instead of a dict

nipoppy/telemetry/README.md
michellewang, line 1: The rest of the codebase doesn't have READMEs like this so I am not sure we should have one here

server/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+```bash
+ssh pavot
+cd ~/server
+docker compose down -v
+docker compose up -d
+docker compose ps               # confirm all three are Up
+```
+
+After a fresh start the dashboard panels will show "No data" until nipoppy commands are run.
```
michellewang, line 110: We definitely want the data to persist after the containers are shut down, not for everything to be wiped

michellewang (commented):
Thanks @blackpearl006, this will be a very useful addition to Nipoppy!

I have some initial comments and pointed out some "big picture" things that should be discussed on Friday.

There is an easy-to-resolve merge conflict, it would be good to address it and merge from `main` so that `pre-commit` can run.

One thing to keep in mind is that we will need tests for everything in the `nipoppy.telemetry` module, but that should wait until other decisions are finalized.

But overall this is great progress, looking forward to digging into it deeper!

nipoppy/telemetry/geo.py
michellewang, line 1: From some local testing, it looks like the country code is only being sent if `nipoppy init` ran successfully. It would be better to send all the time instead

michellewang:
Random thought: should we have a two telemetry servers: one for testing and one for production?

mathdugre:
> Random thought: should we have a two telemetry servers: one for testing and one for production?

I agree it could be useful to have an extra server for dev

nipoppy/cli/cli.py
```diff
@@ -137,13 +153,16 @@ def cli():
...
 def init(**params):
     """Initialize a new dataset."""
     from nipoppy.workflows.dataset_init import InitWorkflow
 
     params = dep_params(**params)
     with exception_handler(InitWorkflow(**params)) as workflow:
         workflow.run()
+        if _TELEMETRY_AVAILABLE:
+            record_location()
```
mathdugre, line 165: Should we send the geo code after each command execution.

nipoppy/cli/cli.py
mathdugre, line 156: We could use the `__name__`  attribute to get the function name directly. This would prevent typos or function mismatch if we rename them in the future.

nipoppy/config/main.py
```diff
@@ -293,3 +293,4 @@ def load(cls, path: StrOrPathLike, apply_substitutions=True) -> Self:
             config_dict[substitutions_key] = substitutions
         config = cls(**config_dict)
         return config
+
```
mathdugre, line 296: This could be removed

nipoppy/telemetry/decorators.py
mathdugre: note for discussion: We could also use the decorator in the future implementation of the dispatcher.
We should evaluate both solutions.

nipoppy/telemetry/geo.py
```diff
@@ -0,0 +1,64 @@
...
+
+    try:
+        # Step 1: Get public IP
+        ip_response = requests.get('https://api.ipify.org', timeout=5)
+        ip_response.raise_for_status()
+        public_ip = ip_response.text.strip()
+
+        # Step 2: Resolve country via db-ip.com
+        # Example response: {"countryCode": "IN", "countryName": "India", ...}
+        response = requests.get(f'https://api.db-ip.com/v2/free/{public_ip}', timeout=5)
```
mathdugre, line 33: What is the daily limit per user?

Note: we could also consider the other option of building a local DB; possibly in the `~/.nipoppy/` dir

nipoppy/telemetry/geo.py
```diff
@@ -0,0 +1,64 @@
...
+
+    try:
+        metrics = get_metrics()
+        if metrics and "location_by_country" in metrics:
+            country_code = get_user_country()
+            metrics["location_by_country"].add(
+                1,
+                attributes={"country": country_code}
+            )
+    except Exception:
```
mathdugre, line 63: We could log when an error occurs

nipoppy/telemetry/geo.py
```diff
@@ -0,0 +1,64 @@
...
+        data = response.json()
+
+        country_code = data.get('countryCode')
+        if country_code and isinstance(country_code, str) and len(country_code) == 2:
+            return country_code.upper()
+
+        return "UNKNOWN"
+
+    except Exception:
+        return "UNKNOWN"
```
mathdugre, line 44: We could log when an error occurs.

mathdugre:
We should use the [Numpy docstring style](https://numpydoc.readthedocs.io/en/latest/format.html#sections)

nipoppy/telemetry/metrics.py
mathdugre: We could also consider using [Pydantic](https://pydantic.dev/docs/validation/latest/get-started/)

nipoppy/telemetry/decorators.py
```diff
@@ -0,0 +1,153 @@
...
+        6. Re-raises SystemExit to preserve original behavior
+    """
+    def decorator(func: Callable) -> Callable:
+        @wraps(func)  # Preserves function metadata for Click
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            # === PRESENTATION MARKER: Record Command Invocation ===
+            _record_command_metric(command_name)
+
+            # === PRESENTATION MARKER: Execute and Track Completion ===
+            status = "success"
```
mathdugre, line 56: We can use `ReturnCode.SUCCESS.name`

nipoppy/telemetry/README.md
blackpearl006: I will delete this file.

mathdugre (commented):
Thank you, a few minors comments

nipoppy/cli/cli.py
blackpearl006: I'll clean all the in-line comments.

nipoppy/telemetry/metrics.py
blackpearl006: I have made a note of it

blackpearl006:
TODO: 
1. Add TelemetryHandler
2. Instead of telemetry call in cli.py use BaseWorkflow
3. Remove the telemetry call at the beginning of each command 
4. Check the telemetry storage and expiration

michellewang:
> TODO:
>
>    Add TelemetryHandler
>    Instead of telemetry call in cli.py use BaseWorkflow
>    Remove the telemetry call at the beginning of each command
>    Check the telemetry storage and expiration

Also: switch to HTTP instead of gRPC? @blackpearl006

mathdugre:
I was looking at the code from [squad](https://github.com/bradygaster/squad) and noticed they also use OTel.
Having a quick look at their implementation, they use an event bus with a Pub/Sub model. It could be interesting to consider a similar implementation in Nipoppy.

## Example components
- **Event**: Defines what an event looks like
```
from dataclasses import dataclass, field
from datetime import datetime, timezone
from typing import Any, Optional

@dataclass
class AppEvent:
    event_type: str
    payload: Any
    session_id: Optional[str] = None
    timestamp: datetime = field(default_factory=lambda: datetime.now(timezone.utc))

```

- **EventBus**:  bus for Pub/Sub communication
```
import logging
from typing import Callable, Dict, List, Set

logger = logging.getLogger(__name__)
EventHandler = Callable[[AppEvent], None]

class EventBus:
    def __init__(self):
        self._handlers: Dict[str, Set[EventHandler]] = {}
        self._wildcard_handlers: Set[EventHandler] = set()

    def subscribe(self, event_type: str, handler: EventHandler):
        if event_type not in self._handlers:
            self._handlers[event_type] = set()
        self._handlers[event_type].add(handler)

    def subscribe_all(self, handler: EventHandler):
        self._wildcard_handlers.add(handler)

    def emit(self, event: AppEvent):
        handlers = self._handlers.get(event.event_type, set()) | self._wildcard_handlers
        
        for handler in handlers:
            try:
                handler(event)
            except Exception as e:
                # Error Isolation: One bad subscriber shouldn't crash the app
                logger.error(f"EventBus handler failed for {event.event_type}: {e}", exc_info=True)

```

- **NoOps wrapper**:  Provide resiliency when telemetry is unavailable.

- **OTelBridge**: A bridge between the EventBus interface and the OTel API

## Exemple usage
```
# main.py
from event_bus import EventBus, AppEvent
from telemetry_bridge import bridge_event_bus_to_otel

# 1. Initialization
bus = EventBus()
bridge_event_bus_to_otel(bus)

# 2. Business Logic (Knows nothing about OpenTelemetry!)
def run_business_logic():
    bus.emit(AppEvent(
        event_type="session:created", 
        session_id="12345", 
        payload={"user": "admin"}
    ))
    
    # ... do work ...
    
    bus.emit(AppEvent(
        event_type="session:error", 
        session_id="12345", 
        payload={"error": "Database connection timeout"}
    ))

if __name__ == "__main__":
    run_business_logic()

```

michellewang:
@blackpearl006 let us know when this is ready for another round of reviews (no rush) :)

blackpearl006 (commented):
Tested end-to-end with pavot as the telemetry server and ducky as the client, and I'm getting a 502 Bad Gateway when sending to https://telemetry.nipoppy.org (empty POST).

Root cause: on pavot, the Caddy reverse proxy is configured for the old gRPC port ( saw it on /etc/caddy/Caddyfile; we need to switch it from 4317 to 4318) and reload caddy. 
@michellewang can you do that ? 

I"ll test the pipeline again once it"s done

blackpearl006:
Tested the pipeline with the required changes to caddyfile. This PR is ready for review

nipoppy/telemetry/decorators.py
blackpearl006: implemented TelemetryHandler in nipoppy/telemetry/handler.py for this PR

nipoppy/config/main.py
blackpearl006: DONE

nipoppy/telemetry/geo.py
blackpearl006: DONE

pyproject.toml
blackpearl006: i have made the switch

nipoppy/cli/__init__.py
blackpearl006: not required, i have reverted this change

nipoppy/cli/cli.py
blackpearl006: yes. the new `handler.py` does not have this try except block

nipoppy/telemetry/__init__.py
blackpearl006: internal import uses the absolute path. I have noted this design choice

nipoppy/cli/pipeline_catalog.py
blackpearl006: done

nipoppy/telemetry/metrics.py
blackpearl006: done

pyproject.toml
```diff
@@ -27,6 +27,11 @@ dependencies = [
     "rich",
     "rich-click",
     "typing-extensions",
+    # OpenTelemetry for the telemetry client
+    "opentelemetry-api>=1.20.0",
+    "opentelemetry-sdk>=1.20.0",
+    "opentelemetry-exporter-otlp-proto-http>=1.20.0",
```
michellewang: Please sort in alphabetical order with the rest of the dependencies. Also I feel the comment is unnecessary

michellewang: Why pin to `>=1.20.0`? We generally don't pin dependencies unless we know there is a feature we use that was only added in a specific version

tests/unit/test_telemetry.py
```diff
@@ -0,0 +1,107 @@
+"""Tests for the TelemetryHandler class.
+
+Telemetry must be non-confounding: if it is disabled, uninitialized, or
+fails internally, it must never raise and never disrupt the caller.
+"""
```
michellewang, line 4:
```suggestion
"""Tests for the TelemetryHandler class."""
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+
+logger = get_logger()
+
+
+def get_user_country() -> str:
+    """
+    Get the user's country code from their public IP address.
+
+    Returns a two-letter ISO country code (e.g. "US", "CA", "IN") or
+    "UNKNOWN" on any failure.
```
michellewang: Is there a reason why this should not return `None` instead of `"UNKNOWN"`?

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+    """
+    try:
+        ip_response = httpx.get("https://api.ipify.org", timeout=5)
```
michellewang: `timeout` can be added as an optional argument to `get_user_country`

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        data = response.json()
+
+        country_code = data.get("countryCode")
+        if country_code and isinstance(country_code, str) and len(country_code) == 2:
+            return country_code.upper()
+
+        return "UNKNOWN"
+
+    except Exception as e:
+        # Debug-level only: telemetry stays invisible in normal runs.
```
michellewang:
```suggestion
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        ip_response = httpx.get("https://api.ipify.org", timeout=5)
+        ip_response.raise_for_status()
+        public_ip = ip_response.text.strip()
+
+        response = httpx.get(f"https://api.db-ip.com/v2/free/{public_ip}", timeout=5)
+        response.raise_for_status()
+        data = response.json()
+
+        country_code = data.get("countryCode")
+        if country_code and isinstance(country_code, str) and len(country_code) == 2:
```
michellewang: I think checking that it is a string is enough. Also, we prefer explicit testing of `if country_code is not None` instead of `if country_code`
```suggestion
        if country_code is not None and isinstance(country_code, str):
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            return country_code.upper()
+
+        return "UNKNOWN"
+
+    except Exception as e:
+        # Debug-level only: telemetry stays invisible in normal runs.
+        logger.debug(f"GeoIP country lookup failed: {e}")
```
michellewang:
```suggestion
        logger.debug(f"Country lookup failed: {e}")
```

Since this is not using GeoIP

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        return "UNKNOWN"
+
+
+@dataclass
+class MetricInstruments:
+    """OpenTelemetry counter instruments used by Nipoppy telemetry."""
```
michellewang:
```suggestion
    """OpenTelemetry counter instruments."""
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+    Self-contained OpenTelemetry metrics handler.
+
+    Each instance owns its own MeterProvider, so it does not touch the global
+    OpenTelemetry state and multiple instances can coexist safely. All public
+    methods are fail-safe and never raise.
+    """
+
+    def __init__(
+        self,
+        service_name: str = "nipoppy",
```
michellewang: Service name should be obtained from `nipoppy.env.PROGRAM_NAME`

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        service_version: str = "1.0.0",
```
michellewang: Service version should have `None` as default value and be obtained by the base workflow before instantiating the handler

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        self._service_name = service_name
+        self._service_version = service_version
+        self._otlp_endpoint = otlp_endpoint
+        self._export_interval_millis = export_interval_millis
+        self._metric_reader = metric_reader
+
+        self._provider: MeterProvider | None = None
+        self._metrics: MetricInstruments | None = None
+        self._shutdown_called = False
```
michellewang: For consistency with the rest of the code base, these should be public (without leading underscore)

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,277 @@
...
+        Initialize the meter provider and metric instruments.
+
+        Safe to call multiple times (only initializes once). Returns False if
+        initialization is disabled or fails; never raises.
+        """
+        if self._provider is not None:
+            return True
+
+        if os.getenv("OTEL_SDK_DISABLED", "false").lower() == "true":
+            return False
```
michellewang, line 125: The OpenTelemetry SDK should handle this internally, no?

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,277 @@
...
+
+        try:
+            # Telemetry must never disrupt the CLI. The OTLP/HTTP exporter logs a
+            # full traceback when the collector is unreachable; silence OTel's own
+            # logging so export failures stay invisible to the user.
+            logging.getLogger("opentelemetry").setLevel(logging.CRITICAL)
```
michellewang, line 129: Technically this is not truly silencing the log

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            # Telemetry must never disrupt the CLI. The OTLP/HTTP exporter logs a
+            # full traceback when the collector is unreachable; silence OTel's own
+            # logging so export failures stay invisible to the user.
+            logging.getLogger("opentelemetry").setLevel(logging.CRITICAL)
+
+            resource = Resource(
+                attributes={
+                    SERVICE_NAME: self._service_name,
+                    SERVICE_VERSION: self._service_version,
+                    "deployment.environment": os.getenv("ENVIRONMENT", "development"),
```
michellewang: What is `deployment.environment` for?

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            reader = self._metric_reader or self._build_default_reader()
+
+            self._provider = MeterProvider(
+                resource=resource,
+                metric_readers=[reader],
+            )
+            meter = self._provider.get_meter(__name__)
+            self._metrics = self._create_metric_instruments(meter)
+
+            # Flush and export pending metrics on exit (normal exit, Ctrl+C, SIGTERM).
```
michellewang: Why does this comment refer to `SIGTERM` if there is a dedicated handling of `SIGTERM` below?

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+                    original_sigterm(signum, frame)
+                else:
+                    sys.exit(0)
+
+            signal.signal(signal.SIGTERM, _sigterm_handler)
+
+            return True
+
+        except Exception as e:
+            # Fail silently: telemetry issues must never surface to the user.
```
michellewang:
```suggestion
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,277 @@
...
+            )
+            return False
+
+    def _build_default_reader(self) -> MetricReader:
+        """Build the default OTLP/HTTP exporting reader."""
+        otlp_endpoint = self._otlp_endpoint
+        if otlp_endpoint is None:
+            otlp_endpoint = os.getenv(
+                "OTEL_EXPORTER_OTLP_ENDPOINT",
+                "https://telemetry.nipoppy.org",
```
michellewang, line 186: This should be a constant defined in `nipoppy.env`

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        os.environ["OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE"] = "delta"
+        otlp_exporter = OTLPMetricExporter(endpoint=otlp_endpoint)
+
+        # Short export interval so the HTTP session is established before shutdown.
+        # For CLI tools the periodic export rarely fires, but the shutdown flush
+        # reuses the already-open connection — making export reliable even for
+        # 2s commands.
+        return PeriodicExportingMetricReader(
+            otlp_exporter,
+            export_interval_millis=min(self._export_interval_millis, 2000),
```
michellewang: It is misleading to override `self._export_interval_millis` here when the default is 10000

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        )
+
+    def _create_metric_instruments(self, meter) -> MetricInstruments:
+        """Create Nipoppy metric instruments."""
```
michellewang:
```suggestion
        """Create metric instruments."""
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            ),
+            location_by_country=meter.create_counter(
+                name="location.by_country",
+                description="Number of installations per country",
+                unit="installations",
+            ),
+        )
+
+    def record_command_completion(self, command_name: str, return_code: int) -> None:
+        """Emit a commands_completed metric. Fail-safe — never raises."""
```
michellewang:
```suggestion
        """Emit a commands_completed metric."""
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+                return
+            if return_code == ReturnCode.SUCCESS:
+                status = "success"
+            elif return_code in (
+                ReturnCode.PARTIAL_SUCCESS,
+                ReturnCode.NO_PARTICIPANTS_OR_SESSIONS_TO_RUN,
+            ):
+                status = "partial"
+            else:
+                status = "failure"
```
michellewang: Can we set `status = return_code.name` instead?

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+            self._metrics.commands_completed.add(
+                1,
+                attributes={
+                    "command": command_name,
+                    "status": status,
+                    "return_code": str(int(return_code)),
```
michellewang:
```suggestion
                    "return_code": str(return_code.value),
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+                },
+            )
+        except Exception:
+            pass
```
michellewang: Here we can log a debug message

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def record_location(self) -> None:
+        """
+        Perform a GeoIP lookup and record the country metric. Fail-safe.
+
+        Called once during `nipoppy init`.
```
michellewang:
```suggestion
        Perform a country code lookup and record the country metric.
```

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+)
+from opentelemetry.sdk.resources import SERVICE_NAME, SERVICE_VERSION, Resource
+
+from nipoppy.exceptions import ReturnCode
+from nipoppy.logger import get_logger
+
+logger = get_logger()
+
+
+def get_user_country() -> str:
```
michellewang: If this function is only used in this file, it can be private

And we should only have the `try`/`catch` blocks for the public API

nipoppy/telemetry/handler.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+        try:
+            if self._metrics is None:
+                return
+            country_code = get_user_country()
+            self._metrics.location_by_country.add(
+                1,
+                attributes={"country": country_code},
+            )
+        except Exception:
```
michellewang: Here we should also log a debug message

nipoppy/workflows/base.py
```diff
@@ -158,6 +161,10 @@ def process_output(output_source, log_prefix: str, log_level=logging.INFO):
 class BaseWorkflow(Base, ABC):
     """Base workflow class with logging/subprocess/filesystem utilities."""
 
+    # Uninitialized, fail-safe placeholder so self.telemetry always exists (even
+    # if run() is never called); run() replaces it with an initialized handler.
+    telemetry = TelemetryHandler()
```
michellewang: It would be cleaner to have `self.telemetry = TelemetryHandler(service_version=__version__)` inside `__init__` instead of having something at the class level

nipoppy/workflows/base.py
```diff
@@ -196,13 +203,20 @@ def run_cleanup(self):
 
     def run(self):
         """Run the workflow."""
+        self.telemetry = TelemetryHandler(service_version=__version__)
```
michellewang: This can be in `__init__` instead

nipoppy/workflows/base.py
```diff
@@ -196,13 +203,20 @@ def run_cleanup(self):
...
-            raise
         finally:
             self.run_cleanup()
+            # return_code is finalized by the CLI's exception_handler after run()
+            # returns, so detect an in-flight exception here to record failures
+            # accurately. Telemetry stays fail-safe and never owns exit codes.
+            return_code = self.return_code
+            if sys.exc_info()[0] is not None and return_code == ReturnCode.SUCCESS:
+                return_code = ReturnCode.UNKNOWN_FAILURE
+            self.telemetry.record_command_completion(self.name, return_code)
```
michellewang, line 217: We probably want to handle this in a better way in the future. I am not a fan of redundancy between this and the top-level exception handler

nipoppy/workflows/dataset_init.py
```diff
@@ -325,6 +325,23 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
+        # To opt out, set the environment variable before running nipoppy:
+        #   export OTEL_SDK_DISABLED=true
+        # To silence this message without disabling telemetry, comment out the
+        # logger.info line below.
+        logger.info(
+            "Nipoppy collects anonymous usage metrics (command counts, country "
+            "of use). No personal data is collected. "
+            "To opt out: export OTEL_SDK_DISABLED=true"
+        )
+        # ─────────────────────────────────────────────────────────────────────
```
michellewang: This needs to be discussed further. `export OTEL_SDK_DISABLED=true` is not a user-friendly way to disable telemetry

nipoppy/workflows/services/telemetry.py
michellewang, line 1: This file should be moved to `workflows/services/telemetry.py`

michellewang (changes requested):
Thanks @blackpearl006, the refactor into the `TelemetryHandler` class is nice!

Please see my comments/questions. Some of the suggestions are to increase consistency with the rest of the code base (in general I think some comments are too verbose/redundant). And we might want to further discuss format of the telemetry metrics (e.g. what to send when the country code cannot be inferred).

On the server/dashboard side:

1. The time range picker does not seem to be working, unless I am doing something wrong
2. "Unique command seen" is not that useful
3. The two panels for "Geographic distribution" are redundant since the circular chart one has the counts/percentages as well
4. For the "Status per command", I think it would be more useful to have percentage instead of count

A few additional things:

1. The tests are missing some coverage, see https://app.codecov.io/gh/nipoppy/nipoppy/pull/978?dropdown=coverage&src=pr&el=h1
2. The server code should maybe be in another repo, but we should discuss this before making a decision
3. We really need to make sure that the telemetry data persists even after the containers are stopped. Right now `server/README.md` says 'After a fresh start the dashboard panels will show "No data" until nippopy command are run', which is not good
4. I think we mentioned that we could track unique users/datasets as well. But that can be implemented at a later time

pyproject.toml
blackpearl006: the client part does not require any specific version, dropping it

nipoppy/telemetry/handler.py
blackpearl006: Not None: OTel export would drop the 'country' label entirely rather than send it empty; "UNKNOWN" keeps it queryable (e.g. "how many lookups failed").

nipoppy/workflows/base.py
blackpearl006: done

nipoppy/workflows/dataset_init.py
blackpearl006: add to ~/.bashrc or create a small file at a standard user-config location (e.g. ~/.config/nipoppy/config.toml)

michellewang: I think we should have an explicit Nipoppy configuration to disable telemetry. For example in the global config file. Using an env var is still fine as an additional layer. To discuss

We should not have a `~/.config/nipoppy/config.toml` for now

nipoppy/telemetry/handler.py
blackpearl006: done

nipoppy/telemetry/handler.py
blackpearl006: fixed as asked (status = return_code.name), but it likely won't surface all 10+ enum values in practice because of the same redundancy you flagged between this and the top-level exception handler at base.py. Telemetry fires before that handler assigns the precise code.

nipoppy/telemetry/handler.py
blackpearl006: we have changed that to nipoppy env variable

server/README.md
blackpearl006: updated the readme (server ) about this

nipoppy/workflows/services/telemetry.py
blackpearl006: fixed; the SDK returns NoOpMeter

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+logger = get_logger()
+
+
+def _get_user_country(timeout: float = 5) -> str:
+    """
+    Get the user's country code from their public IP address.
+
+    Returns a two-letter ISO country code (e.g. "US", "CA", "IN") or
+    "UNKNOWN" on any failure. Not fail-safe on its own — callers are
+    responsible for exception handling (see `record_location`).
```
michellewang, line 37:
```suggestion
    Returns a two-letter ISO country code (e.g. "US", "CA", "IN") or
    "UNKNOWN" on any failure.
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+    response = httpx.get(f"https://api.db-ip.com/v2/free/{public_ip}", timeout=timeout)
+    response.raise_for_status()
+    data = response.json()
+
+    country_code = data.get("countryCode")
+    if country_code is not None and isinstance(country_code, str):
+        return country_code.upper()
+
+    return "UNKNOWN"
```
michellewang: If we have the `raise_for_status()` calls, that means there will be an exception raised if an error occurs. If we want to return `"UNKNOWN"` on any failure we need to handle this differently

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+class TelemetryHandler:
+    """
+    Self-contained OpenTelemetry metrics handler.
+
+    Each instance owns its own MeterProvider, so it does not touch the global
+    OpenTelemetry state and multiple instances can coexist safely. All public
+    methods are fail-safe and never raise.
+    """
```
michellewang, line 71:
```suggestion
    """Self-contained OpenTelemetry metrics handler."""
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def __init__(
+        self,
+        service_name: str = PROGRAM_NAME,
+        service_version: str | None = None,
```
michellewang, line 76: We should have the default be `nipoppy.env.PROGRAM_VERSION` which was added recently

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+        otlp_endpoint: str | None = None,
+        export_interval_millis: int = TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS,
+        metric_reader: MetricReader | None = None,
+    ) -> None:
+        """Create a telemetry handler (does not initialize OpenTelemetry yet).
```
michellewang, line 81:
```suggestion
        """Create a telemetry handler.
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+    Each instance owns its own MeterProvider, so it does not touch the global
+    OpenTelemetry state and multiple instances can coexist safely. All public
+    methods are fail-safe and never raise.
+    """
+
+    def __init__(
+        self,
+        service_name: str = PROGRAM_NAME,
+        service_version: str | None = None,
+        otlp_endpoint: str | None = None,
```
michellewang, line 77: We can add a constant to `nipoppy.env` for the endpoint and use it as the default here

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+        metric_reader: MetricReader | None = None,
+    ) -> None:
+        """Create a telemetry handler (does not initialize OpenTelemetry yet).
+
+        Parameters
+        ----------
+        service_name : str
+            Service name for metrics (default: `nipoppy.env.PROGRAM_NAME`).
+        service_version : str, optional
+            Version tag, supplied by the base workflow (default: None).
```
michellewang, line 88:
```suggestion
            Version tag (default: None).
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+    def is_enabled(self) -> bool:
+        """True if telemetry is initialized and active."""
+        return self.metrics is not None
+
+    def initialize(self) -> bool:
+        """
+        Initialize the meter provider and metric instruments.
+
+        Safe to call multiple times (only initializes once). Returns False if
+        initialization is disabled or fails; never raises.
```
michellewang, line 119:
```suggestion
        initialization is disabled or fails.
```

nipoppy/workflows/services/telemetry.py
michellewang:
```suggestion
```

I tested manually and it looks like we don't need this. It's handled internally by OpenTelemetry

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+        if self.provider is not None:
+            return True
+
+        if os.getenv("OTEL_SDK_DISABLED", "false").lower() == "true":
+            return False
+
+        try:
+            # OTLP/HTTP exporter uses requests/urllib3 internally; silence both loggers.
+            logging.getLogger("opentelemetry").setLevel(logging.CRITICAL)
+            logging.getLogger("urllib3").setLevel(logging.CRITICAL)
```
michellewang, line 130: This should go in `nipoppy.logger.get_logger`

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+                default_environment = "production"
+            else:
+                default_environment = "development"
+
+            resource = Resource(
+                attributes={
+                    SERVICE_NAME: self.service_name,
+                    SERVICE_VERSION: self.service_version or "unknown",
+                    "deployment.environment": os.getenv(
+                        "ENVIRONMENT", default_environment
```
michellewang, line 142: We should discuss if this is useful given that we already have the version number. If we do choose to use an env var for the environment, it should be something like `NIPOPPY_TELEMETRY_ENVIRONMENT` instead of just `ENVIRONMENT`

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def initialize(self) -> bool:
+        """
+        Initialize the meter provider and metric instruments.
+
+        Safe to call multiple times (only initializes once). Returns False if
+        initialization is disabled or fails; never raises.
+        """
+        if self.provider is not None:
+            return True
```
michellewang, line 122: Is it enough to check the provider or should we also check the metrics?

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+            atexit.register(self.shutdown)
+
+            original_sigterm = signal.getsignal(signal.SIGTERM)
+
+            def _sigterm_handler(signum, frame):
+                self.shutdown()
+                if callable(original_sigterm):
+                    original_sigterm(signum, frame)
+                else:
+                    sys.exit(0)
```
michellewang, line 168: My general preference is for nested function definitions to be at the top, not in the middle of the function

nipoppy/workflows/services/telemetry.py
michellewang: @blackpearl006 reopening as it doesn't look like the telemetry endpoint has been moved to `nipoppy.env`.

I'm wondering if we should even allow a different telemetry endpoint or just force the use of `telemetry.nipoppy.org` all the time. Don't see a need for us to set something else.

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+            otlp_endpoint = os.getenv(
+                "OTEL_EXPORTER_OTLP_ENDPOINT",
+                "https://telemetry.nipoppy.org",
+            )
+
+        # OTLP/HTTP keeps the scheme in the URL (https:// implies TLS). When an
+        # endpoint is passed explicitly the SDK does not append the signal path,
+        # so add /v1/metrics here if the user gave only a base endpoint.
+        if not otlp_endpoint.rstrip("/").endswith("/v1/metrics"):
+            otlp_endpoint = otlp_endpoint.rstrip("/") + "/v1/metrics"
```
michellewang, line 193: Related to the above: if we only use `telemetry.nipoppy.org` then we can simplify this logic

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+        # Force DELTA temporality so the collector's deltatocumulative processor
+        # can accumulate across short-lived process runs. Without this the
+        # OTLPMetricExporter defaults to CUMULATIVE, sending value=1 every time.
+        os.environ["OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE"] = "delta"
+        otlp_exporter = OTLPMetricExporter(endpoint=otlp_endpoint)
+
+        # Short export interval so the HTTP session is established before shutdown.
+        # For CLI tools the periodic export rarely fires, but the shutdown flush
+        # reuses the already-open connection — making export reliable even for
+        # 2s commands.
```
michellewang, line 204:
```suggestion
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+from opentelemetry.sdk.resources import SERVICE_NAME, SERVICE_VERSION, Resource
+
+from nipoppy.env import PROGRAM_NAME, TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS
+from nipoppy.exceptions import ReturnCode
+from nipoppy.logger import get_logger
+
+logger = get_logger()
+
+
+def _get_user_country(timeout: float = 5) -> str:
```
michellewang, line 31: Timeout of 5 seconds is too long. We can do 1s instead.

nipoppy/workflows/services/telemetry.py
michellewang, line 51: Since we are already letting exceptions be raised, I think we might as well only return the correct country code
```suggestion
    return country_code.upper()
```

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+            original_sigterm = signal.getsignal(signal.SIGTERM)
+
+            def _sigterm_handler(signum, frame):
+                self.shutdown()
+                if callable(original_sigterm):
+                    original_sigterm(signum, frame)
+                else:
+                    sys.exit(0)
+
+            signal.signal(signal.SIGTERM, _sigterm_handler)
```
michellewang, line 170: We should discuss if we really want all that extra handling (especially for `SIGTERM`) or if we are okay with losing some of this for the sake of more understandable code

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+        # endpoint is passed explicitly the SDK does not append the signal path,
+        # so add /v1/metrics here if the user gave only a base endpoint.
+        if not otlp_endpoint.rstrip("/").endswith("/v1/metrics"):
+            otlp_endpoint = otlp_endpoint.rstrip("/") + "/v1/metrics"
+
+        # Force DELTA temporality so the collector's deltatocumulative processor
+        # can accumulate across short-lived process runs. Without this the
+        # OTLPMetricExporter defaults to CUMULATIVE, sending value=1 every time.
+        os.environ["OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE"] = "delta"
+        otlp_exporter = OTLPMetricExporter(endpoint=otlp_endpoint)
```
michellewang, line 199: It looks like this can be set when instantiating `OTLPMetricExporter`, which takes a `preferred_temporality` argument

This would be better than changing the environment.

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # Short export interval so the HTTP session is established before shutdown.
+        # For CLI tools the periodic export rarely fires, but the shutdown flush
+        # reuses the already-open connection — making export reliable even for
+        # 2s commands.
+        return PeriodicExportingMetricReader(
+            otlp_exporter,
+            export_interval_millis=min(
+                self.export_interval_millis, TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS
```
michellewang, line 208: What is the default `export_interval_millis` and is `TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS` larger or smaller than it?
Should we set a value for timeout as well?

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+            ),
+        )
+
+    def record_command_completion(
+        self, command_name: str, return_code: ReturnCode
+    ) -> None:
+        """Emit a commands_completed metric."""
+        try:
+            if self.metrics is None:
+                return
```
michellewang, line 233: We might want to have an explicit attribute `initialized` instead of checking if `self.metrics` and/or `self.provided` is set

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+            country_code = _get_user_country()
+            self.metrics.location_by_country.add(
+                1,
+                attributes={"country": country_code},
+            )
+        except Exception as e:
+            logger.debug(f"Country lookup failed: {e}")
+
+    def shutdown(self) -> None:
+        """Flush and shut down the meter provider. Safe to call multiple times."""
```
michellewang, line 259:
```suggestion
        """Flush and shut down the meter provider."""
```

nipoppy/workflows/base.py
```diff
@@ -197,13 +202,19 @@ def run_cleanup(self):
 
     def run(self):
         """Run the workflow."""
+        self.telemetry.initialize()
```
michellewang, line 205: I would put this in `run_setup` instead

nipoppy/workflows/services/telemetry.py
michellewang: Is this redundant with `is_enabled`?

nipoppy/workflows/services/telemetry.py
michellewang: Perhaps the `is_enabled` would do the job?

nipoppy/workflows/base.py
```diff
@@ -197,13 +202,19 @@ def run_cleanup(self):
...
         try:
             self.run_setup()
             self.run_main()
-        except Exception:
-            raise
         finally:
             self.run_cleanup()
+            # return_code is finalized by the CLI's exception_handler after run()
+            # returns, so detect an in-flight exception here to record failures
+            # accurately.
```
michellewang, line 213: @blackpearl006 could you open a follow-up issue to refactor the exception handler in the CLI code?

michellewang (changes requested):
A few more comments/suggestions to make the implementation cleaner.

We need to open these follow-up issues:
- User config/prompting to disable telemetry
- Exception handler
- Move server code to another repo

nipoppy/workflows/base.py
```diff
@@ -178,6 +181,8 @@ def __init__(self, name: str, verbose: bool = False, dry_run: bool = False):
         # for the CLI
         self.return_code = ReturnCode.SUCCESS
 
+        self.telemetry = TelemetryHandler(service_version=__version__)
```
mathdugre: We should have a `get_telemetry_handler` that returns a singleton `TelemetryHandler`. This would avoid needing to create the TelemetryHandler then initializing it.
Note: the `get_telemetry_handler` should return a singleton object, similar to the current `initialize` that is idempotent

nipoppy/workflows/dataset_init.py
```diff
@@ -77,6 +77,7 @@ def run_main(self):
         Copy default config files.
         Copy HPC config files.
         """
+        self.telemetry.record_location()
```
mathdugre: The current call is blocking, which is not ideal.

We should strongly consider having non-blocking call. Something like this (modified from AI):
```py
import threading

class TelemetryHandler:
    def __init__(self, ...):
        ...
        self._location_thread: threading.Thread | None = None

    def record_location_async(self) -> None:
        """Fetch location in a background thread and emit the metric."""
        if not self.is_enabled:
            return

        def _worker():
            try:
                country = _get_user_country(timeout=.5)
                self.metrics.location_by_country.add(1, attributes={"country": country})
            except Exception as e:
                logger.debug(f"Country lookup failed: {e}")

        self._location_thread = threading.Thread(target=_worker, daemon=True, name="nipoppy-geoip")
        self._location_thread.start()

    def shutdown(self) -> None:
        # Give the background lookup a brief window (e.g. 0.2s) to complete before process exit
        if self._location_thread and self._location_thread.is_alive():
            self._location_thread.join(timeout=0.2)
        if self.provider is not None:
            self.provider.shutdown()
```

--- 
In the future, we could also add a local cache (i.e., `~/.nipoppy/country-code` with a cache invalidation of let's say 1 week. This would allow bypassing the network request altogether for users who frequently initialize nipoppy studies.

nipoppy/workflows/services/telemetry.py
mathdugre: I think allowing a different telemetry endpoint could be nice, as user might want to have a private dashboard instead of sending the telemetry data to us. 🤔

nipoppy/workflows/services/telemetry.py
mathdugre: Along the same note. I agree with this comment generated by Gemini:
> self.provider, self.metrics, self.shutdown_called, and is_enabled represent overlapping state. A unified self._initialized boolean and clean shutdown guard simplifies the class lifecycle significantly.

nipoppy/workflows/services/telemetry.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Returns a two-letter ISO country code (e.g. "US", "CA", "IN") or
+    "UNKNOWN" on any failure. Not fail-safe on its own — callers are
+    responsible for exception handling (see `record_location`).
+    """
+    ip_response = httpx.get("https://api.ipify.org", timeout=timeout)
+    ip_response.raise_for_status()
+    public_ip = ip_response.text.strip()
+
+    response = httpx.get(f"https://api.db-ip.com/v2/free/{public_ip}", timeout=timeout)
```
mathdugre: The db-ip docs says we can use "self" instead of an IP to get the country code of the calling IP. This would avoid making the request to ipify.

```suggestion
    response = httpx.get(f"https://api.db-ip.com/v2/free/self", timeout=timeout)
```

mathdugre (changes requested):
Some more high-level changes to be done

nipoppy/workflows/dataset_init.py
blackpearl006: I could not find any threading already written in the nipoppy codebase. Anyways, implementing this "fire-and-forget network call" suggestion. 
changed OTEL_EXPORTER_OTLP_ENDPOINT to a variable in `env.py`
changed is_enabled to is_initialised but does not represent the overlapping state

nipoppy/workflows/services/telemetry.py
blackpearl006: this deployment.environment variable was unnecessary, i have deleted this

nipoppy/workflows/services/telemetry.py
blackpearl006: updated

nipoppy/workflows/services/telemetry.py
blackpearl006: fixed in commit f374cc8192e8bb139f19911ca37ca11f7d3114da

nipoppy/workflows/services/telemetry.py
blackpearl006: fixed in commit f374cc8

nipoppy/workflows/services/telemetry.py
blackpearl006: i have moved the telemetry endpoint to nipoppy.env 
also similar topic ? https://github.com/nipoppy/nipoppy/pull/978#discussion_r4065118626

nipoppy/workflows/services/telemetry.py
blackpearl006: had a discussion with @mathdugre it's fine to leave it for now if in case in the future we would like to support docker as well

nipoppy/workflows/base.py
blackpearl006: 108669d

nipoppy/workflows/dataset_init.py
blackpearl006: #1130

nipoppy/telemetry/handler.py
blackpearl006: outdated

nipoppy/telemetry/handler.py
blackpearl006: now using TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS

nipoppy/workflows/services/telemetry.py
blackpearl006: TELEMETRY_MAX_EXPORT_INTERVAL_MILLIS = 2000 and so is export_interval_millis they were same. Default OTel SDK timeout was 10 seconds now i have taken using TELEMETRY_EXPORT_TIMEOUT_SECONDS to lower that

nipoppy/telemetry/geo.py
blackpearl006: 500/day

PR #979, opened 2026-04-10
[ENH] Expose "length" and number of unique participants in a Study

### LINKED ISSUE #954: [ENH] Expose dataset/manifest length in `NipoppyDataRetriever`

### New feature

Requested by @sharkovsky for Fed-BioMed's `NipoppyDataset`

Expose a length/size measure from `NipoppyDataRetriever` (or possibly `Study`). This needs to be available right after instantiation (i.e., before specifying phenotypes/derivatives). The most straightforward number would be the length of the manifest or the number of unique participants in the study.

@sharkovsky I assume you don't have any requirement for how you would get that number? Like it doesn't necessarily need to be `len(NipoppyDataRetriever())`, right?

### Unclear documentation

_No response_

michellewang:
- Closes #954

- Define `len(Study(...))` to be the length of the manifest file
- Add `Study.get_n_participants()` (and `Manifest.get_participants()`)
    - Not sure about this one: it feels restrictive since it would be more useful to have the full list of participant IDs instead, but then we would have to make the manifest public and we don't want to do that yet. So maybe it's best to remove this for now and reconsider later. Anyway for Fed-BioMed's `NipoppyDataset` the `__len__` is enough. Thoughts @mathdugre?

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--979.org.readthedocs.build/en/979/

### DIFF
--- a/nipoppy/study.py
+++ b/nipoppy/study.py
@@ -44,6 +44,10 @@ class Study(Base):
         self.layout = layout
         self.verbose = verbose
 
+    def __len__(self):
+        """Get the number of unique participant-visit combinations in the study."""
+        return len(self.manifest)
+
     @cached_property
     def config(self) -> Config:
         """The main configuration object."""
--- a/nipoppy/study.py
+++ b/nipoppy/study.py
@@ -40,7 +40,8 @@ class Study(Base):
         """Get the number of unique participant-visit combinations in the study."""
         return len(self.manifest)
 
-    def get_n_participants(self) -> int:
+    @property
+    def n_unique_participants(self) -> int:
         """Get the number of unique participants in the study."""
         return len(set(self.manifest.get_participants()))
 
--- a/nipoppy/study.py
+++ b/nipoppy/study.py
@@ -40,11 +40,6 @@ class Study(Base):
         """Get the number of unique participant-visit combinations in the study."""
         return len(self.manifest)
 
-    @property
-    def n_unique_participants(self) -> int:
-        """Get the number of unique participants in the study."""
-        return len(self.manifest.get_unique_participants())
-
     @cached_property
     def config(self) -> Config:
         """The main configuration object."""
--- a/nipoppy/study.py
+++ b/nipoppy/study.py
@@ -43,7 +43,7 @@ class Study(Base):
     @property
     def n_unique_participants(self) -> int:
         """Get the number of unique participants in the study."""
-        return len(set(self.manifest.get_participants()))
+        return len(self.manifest.get_unique_participants())
 
     @cached_property
     def config(self) -> Config:
--- a/nipoppy/study.py
+++ b/nipoppy/study.py
@@ -40,7 +40,7 @@ class Study(Base):
         """Get the number of unique participant-visit combinations in the study."""
         return len(self.manifest)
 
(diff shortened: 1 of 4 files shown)


nipoppy/study.py
```diff
@@ -36,6 +36,14 @@ def __init__(self, layout: DatasetLayout):
         super().__init__()
         self.layout = layout
 
+    def __len__(self):
+        """Get the number of unique participant-visit combinations in the study."""
+        return len(self.manifest)
+
+    def get_n_participants(self) -> int:
```
mathdugre: I think a `n_participants` property might be a bit more pythonic.

mathdugre: Also, we might want to make it explicit that it's nunique participants.

nipoppy/tabular/manifest.py
```diff
@@ -127,6 +127,10 @@ def get_imaging_subset(self, session_id: Optional[str] = None):
             return manifest[manifest[self.col_session_id] == session_id]
         return manifest
 
+    def get_participants(self) -> list[str]:
+        """Get participant IDs."""
+        return self.loc[:, self.col_participant_id].to_list()
+
```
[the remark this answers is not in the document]
mathdugre: I think this is valid. We might only need the unique participants function, actually.

tests/unit/tabular/test_manifest.py
```diff
@@ -122,6 +122,23 @@ def test_get_imaging_subset(data, session_id, expected_count):
     assert len(manifest_with_imaging_only) == expected_count
 
 
+@pytest.mark.parametrize(
+    "participant_ids",
+    [
+        ["01", "01", "01", "02", "02", "03", "04"],
+    ],
+)
+def test_get_participants(participant_ids):
```
mathdugre: Similar has the other comments, I think we should use nunique participant instead of the list with possible duplicates.

mathdugre (changes requested):

mathdugre (changes requested):
The `__len__` function looks good. I'm not as sure of `get_n_participant` function.

nipoppy/tabular/manifest.py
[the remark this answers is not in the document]
michellewang: @mathdugre do you prefer this to be a property as well? Or do we keep it as a method?

michellewang:
@mathdugre I'm also not sure about whether we even want `Study.n_unique_participants`. Not convinced it's super useful and I think I would rather rely on `Manifest.get_unique_participants()` (or `Manifest.unique_participants` if we decide to make it a property) once that becomes part of the public API

- By the way I fixed the uncollected test reported in #896

mathdugre:
> @mathdugre I'm also not sure about whether we even want `Study.n_unique_participants`. Not convinced it's super useful and I think I would rather rely on `Manifest.get_unique_participants()` (or `Manifest.unique_participants` if we decide to make it a property) once that becomes part of the public API

We could add it in another PR then, if we see need of it in the future
@michellewang

mathdugre (approved):
LGTM 🧑‍🍳

Issue #981, opened 2026-04-10
[ENH] Improve config schema awareness/handling


michellewang:
The latest release (0.4.6) adds a `BIND_ARGS` field to container config, which introduces some potential issues:
- Old versions of pipeline configs will work with 0.4.6 (good)
- Pipeline configs with `BIND_ARGS` will not work with versions <0.4.6

We should make use of the `SCHEMA_VERSION` field to minimize friction:
- if a user tries to run/install a pipeline that has a newer schema version, we should error out with an error message of "update nipoppy or use an older config"
- we could make `nipoppy pipeline search` look for only accepted schema versions, if possible
- the global config schema should also have a version

Not sure about timeline/urgency since we already broke things with 0.4.6 and `BIND_ARGS`, but we should discuss this as it would be good to have a clear process.

Not directly related: when validating a pipeline, we might want to emit (deprecation?) warnings when we detect some legacy features (e.g. use of `--bind-args` or `[[NIPOPPY_CONTAINER_COMMAND]]`), especially on pipeline upload. See #771

mathdugre:
Discussion with @michellewang 

**Breaking:** remove or rename fields
**Supported:** new fields

We want to support previous schema versions. Deprecation cycle may be needed in future.

We would need ~two~ multiple schema version:
1. study
2. pipeline bundle
3. tracker
4. layout

mathdugre:
Dev meeting discussion:

We can handle new optional fields in the config files.

If we remove config option, in the future, we can think of a way to handle this. As of now, we only added config options.

PR #983, opened 2026-04-14
[ENH] Allow comments in configuration files

mathdugre:
> [!warning]
> Blocked by: https://github.com/nipoppy/nipoppy/issues/981

- Closes #973 
- 

Implement and use a JSON-C parser for reading, creating, and updating config files.
This allows for comment or trailing commas in the config files.

## Checklist (for reviewers)

- [x] Tests pass
- [ ] Checks pass

For new features:
- [x] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--983.org.readthedocs.build/en/983/


mathdugre:
@michellewang , requested review mostly for the usage side. I only had a quick glimpse at the internal function implementation created by copilot and will do a bit of refactoring if the usage of the non-internal functions looks good to you.

tldr; When using json-c or json5 style, we will need to use the `update_jsonc_file` or `update_jsonc_text` function to update the key. This is done via by chaining the key hierarchy in an array; e.g., `test.hello: world` === `{"test" : {"hello": "world"}}`

mathdugre:
Comma insertion in dict with existing value works but looks bad:

```jsonc
        "PROCESSING": {
            "fsqc": {
                "2.1.1": {
                    // test test test
                    "FREESURFER_VERSION": null
                }
            }
        ,
            "qsiprep": {
                "1.0.2": {
                    "FREESURFER_LICENSE_FILE": null
                ,
                    "TEMPLATEFLOW_HOME": null}
            }},
```

mathdugre:
Discussion with @michellewang :
- We should allow invocation to use JSON-C
- We could use the .jsonc extension since we will do a breaking change to the layout

michellewang:
Mentioned by @nx10: https://github.com/spyoungtech/json-five

nipoppy/utils/json5.py
michellewang, line 1: Would it be okay to name this file `jsonc.py` instead?

nipoppy/utils/utils.py
```diff
@@ -69,30 +71,41 @@ def get_pipeline_tag(
...
+    parser = json5.loads if allow_json5 else json.loads
+    try:
+        return parser(json_text, **kwargs)
+    except Exception as e:
+        pos = getattr(e, "pos", 0)
+        raise json.JSONDecodeError(
+            f"Error loading JSON file at {fpath}",
+            json_text,
+            pos,
+        ) from e
```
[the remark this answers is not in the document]
michellewang: I agree with this comment, we should only be catching `json.JSONDecodeError` and whatever the other parser library uses

nipoppy/workflows/pipeline_store/install.py
```diff
@@ -121,7 +122,26 @@ def _update_config_and_save(self, pipeline_config: BasePipelineConfig) -> Config
...
+                            [
+                                "PIPELINE_VARIABLES",
+                                pipeline_variables_key,
+                                pipeline_config.NAME,
+                                pipeline_config.VERSION,
+                                variable_name,
+                            ],
+                            variable_value,
+                        )
+                    )
```
michellewang: Is it possible to have this in the loop above?

tests/unit/config/test_main.py
```diff
@@ -242,6 +242,22 @@ def test_load(path):
         assert hasattr(config, field)
 
 
+def test_load_json5_with_comments_and_trailing_commas(tmp_path: Path):
+    fpath_config = tmp_path / "global_config.json"
+    fpath_config.write_text("""
```
michellewang: Is it good style to have triple-quoted strings like this?

nipoppy/utils/json5.py
michellewang: Or `json5` actually

michellewang (commented):
Some thoughts:
- An alternative parser would be `hjson`: https://github.com/hjson/hjson-py, [recommended by one of the devs for `pyjson5`](https://github.com/dpranke/pyjson5/issues/28#issuecomment-3663185259)
- Need to double-check if all the added tests for existing modules are useful/correct.
- The implementation is very obscure which is not good for maintenance. Especially since we already found a bug (it would be interesting to see how hard it is to fix the bug). At the very least we would need a test suite that is more comprehensive than the current one.

mathdugre:
An alternative parser would be `hjson`: https://github.com/hjson/hjson-py, [recommended by one of the devs for `pyjson5`](https://github.com/dpranke/pyjson5/issues/28#issuecomment-3663185259)
`hjson` seems to have issues open about vulnerability for over a year without fix. I would be worried about the maintenance of the package. Also, they only read files with comments and not round-trip comments.

nipoppy/utils/json5.py
michellewang: @mathdugre

nipoppy/utils/utils.py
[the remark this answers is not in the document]
michellewang: Should we use `nipoppy.exceptions.JSONError` here instead?

nipoppy/workflows/pipeline_store/install.py
michellewang: @mathdugre this was not addressed. I think it would be better to build the `updates` list in the `for variable_name in added_variables:` loop, and have the `if not self.dry_run` only do the JSONC update call.

tests/unit/utils/test_jsonc_edit.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_update_jsonc_text_updates_existing_value_and_preserves_comment():
+    text = """
+{
+  // keep this comment
+  "NAME": "old_name",
+  "VERSION": "1.0.0",
+}
+""".strip()
```
michellewang: Really not a fan of this way of having `text`, but the alternative would be to store in separate files which could be annoying to check...

tests/unit/workflows/pipeline_store/test_install.py
```diff
@@ -231,6 +223,26 @@ def test_update_config_and_save_no_overwrite(
...
+def test_update_config_and_save_preserves_jsonc(
+    workflow: PipelineInstallWorkflow,
+    pipeline_config: ProcessingPipelineConfig,
+):
+    fpath_config = workflow.study.layout.fpath_config
+    jsonc_text = fpath_config.read_text().replace(
+        '"PIPELINE_VARIABLES": {',
+        '"PIPELINE_VARIABLES": {\n        // keep this comment',
+        1,
+    )
```
michellewang: This is not very clean

michellewang (changes requested):
I agree that the code is cleaner now, thanks.

A few more comments, some from the previous review have not been addressed yet.

Some merge conflicts.

And we need to fix the formatting issue.

nipoppy/utils/json5_edit.py
```diff
@@ -0,0 +1,669 @@
+"""Utilities to edit JSON5 objects while preserving comments.
+
+Overview
+--------
+This module applies targeted edits to JSON5 text *without* reformatting
+it, so comments and existing layout survive. It never rebuilds the document
+from a parsed tree. Instead it works in two phases:
```
michellewang:
```suggestion
This module applies targeted edits to JSON5 text *without* reformatting
it, so comments and existing layout survive. Edits work in two phases:
```

nipoppy/workflows/pipeline_store/install.py
```diff
@@ -110,9 +111,24 @@ def _update_config_and_save(self, pipeline_config: BasePipelineConfig) -> Config
             logger.warning(
                 f"Adding {len(added_variables)} variable(s) to the global config file:"
             )
+            json5_updates = []
```
michellewang: Can we rename to just `updates`?

tests/unit/utils/test_json5.py
```diff
@@ -0,0 +1,277 @@
...
+""".strip()
+
+
+def test_update_json5_text_raises_on_empty_key_path():
+    with pytest.raises(ValueError, match="Key path cannot be empty"):
+        update_json5_text('{"A": 1}', [([], "x")])
+
+
+def test_update_json5_text_raises_on_invalid_json5():
+    with pytest.raises(ValueError):
```
michellewang, line 179: This should be `json.JSONDecodeError`

tests/unit/utils/test_json5.py
```diff
@@ -0,0 +1,277 @@
...
+
+# ---------------------------------------------------------------------------
+# Coverage: defensive guards in private helpers. These are not reachable
+# through the validated public API (json5.loads rejects malformed input
+# before any of these guards would trigger), so they are unit-tested
+# directly against the private helpers.
+# ---------------------------------------------------------------------------
+
+
+def test_scanner_find_string_end_index_raises_when_not_on_quote():
```
michellewang, line 207: These should possibly also be `json.JSONDecodeError`s

michellewang (changes requested):
Almost there :)

tests/unit/utils/test_json5.py
mathdugre: > `ValueError` if given an invalid document. This is different from the `json` module, which raises `json.JSONDecodeError`.

From the json5.load documentation

tests/unit/utils/test_json5.py
mathdugre: I disagree. We are testing the scanner method to retrieve a string end index and has nothing to do with the JSON decoding. The test is ensuring that the `find_string_end_index` raise an error when called with the scanner pointing at a non-quote character; that is, not a string.

nipoppy/utils/utils.py
```diff
@@ -69,26 +71,36 @@ def get_pipeline_tag(
...
-            return json.load(file, **kwargs)
-        except json.JSONDecodeError as e:
-            raise JSONError(e, fpath=Path(fpath)) from e
+    fpath = Path(fpath)
+    json_text = fpath.read_text()
+    parser = json5.loads if allow_json5 else json.loads
+    try:
+        return parser(json_text, **kwargs)
+    except json.JSONDecodeError as e:
+        raise JSONError(e, fpath=fpath) from e
```
michellewang: This will raise (non-Nipoppy) `ValueError` if `json5.loads` fails.

It's not easy to coerce the `json5`'s `ValueError` into a `json.JSONDecodeError`.

What do you think of this:

```python
class _JSONError(NipoppyError):
    code = ReturnCode.UNKNOWN_FAILURE
    default_hint = "Suggested fix: Check the JSON file for syntax errors, such as missing commas or mismatched brackets."  # noqa:E501


class JSONError(_JSONError, json.JSONDecodeError):
    """Exception raised for JSON parsing errors, with context about the file path."""

    def __init__(
        self,
        e: json.JSONDecodeError,
        *,
        fpath: StrOrPathLike,
        hint: str | None = None,
    ):
        e.msg += f": {fpath}"
        json.JSONDecodeError.__init__(self, e.msg, e.doc, e.pos)

        # self.args[0] is the JSONDecodeError error message
        _JSONError.__init__(self, self.args[0], hint=hint)


class JSON5Error(_JSONError):
    """Exception raised for JSON5 parsing errors, with context about the file path."""

    def __init__(
        self,
        e: ValueError,
        *,
        fpath: StrOrPathLike,
        hint: str | None = None,
    ):
        message = f"{e}: {fpath}"
        _JSONError.__init__(self, message, hint=hint)
```

Then `load_json` could have something like this:

```python
    fpath = Path(fpath)
    json_text = fpath.read_text()
    if allow_json5:
        try:
            return json5.loads(json_text, **kwargs)
        except ValueError as e:
            raise JSON5Error(e, fpath=fpath) from e
    else:
        try:
            return json.loads(json_text, **kwargs)
        except json.JSONDecodeError as e:
            raise JSONError(e, fpath=fpath) from e
```

And possibly also apply this to `json5_edit`'s public API

mathdugre: Implemented in 
1121ccf

nipoppy/exceptions.py
```diff
@@ -76,7 +80,23 @@ def __init__(
...
+    def __init__(
+        self,
+        e: ValueError,
+        *,
+        fpath: StrOrPathLike,
+        hint: str | None = None,
+    ):
+        msg = f"{str(e)}: {fpath}"
+
+        # self.args[0] is the JSONDecodeError error message
```
michellewang:
```suggestion
        # self.args[0] is the ValueError error message
```

michellewang (changes requested):
@mathdugre there is a merge conflict

michellewang (commented):
@mathdugre I think this is basically ready for approval/merge, but I don't like the discrepancy in test design in these files:
- `test/units/config/test_main.py`
- `tests/unit/test_pipeline_validation.py`

because the existing tests load valid/invalid files in `tests/data`, so I think we should create new JSON files and put them in the same place and use `pytest.mark.parametrize` there. What do you think?

I'm fine with the other occurrences of the triple-quoted blocks in the other files.

mathdugre:
> @mathdugre I think this is basically ready for approval/merge, but I don't like the discrepancy in test design in these files:
> 
> * `test/units/config/test_main.py`
> * `tests/unit/test_pipeline_validation.py`
> 
> because the existing tests load valid/invalid files in `tests/data`, so I think we should create new JSON files and put them in the same place and use `pytest.mark.parametrize` there. What do you think?
> 
> I'm fine with the other occurrences of the triple-quoted blocks in the other files.

I have a strong preference for having the json directly in the test, since it's only a few line snippets. It's faster to understand the tests when looking at them and the snippets are never reused. Also, the parametrization would be very convoluted given that the tests use different functions, with and without exceptions, etc.

michellewang:
> Also, the parametrization would be very convoluted given that the tests use different functions, with and without exceptions, etc.

I am just talking about the tests that check if a config is valid. E.g.

```python
def test_load_pipeline_config_file():
    assert isinstance(
        _load_pipeline_config_file(DPATH_TEST_DATA / "pipeline_config-valid.json"),
        BasePipelineConfig,
    )


def test_load_pipeline_config_file_json5(tmp_path: Path):
    fpath = tmp_path / "pipeline_config.json"
    fpath.write_text("""
{
  // comments and trailing commas should be accepted
  "NAME": "test_pipeline",
  "VERSION": "test_version",
  "SCHEMA_VERSION": "1",
  "PIPELINE_TYPE": "processing",
}
""".strip())

    config = _load_pipeline_config_file(fpath)
    assert isinstance(config, BasePipelineConfig)

```

These two tests can be merged into a single one, and actually the parametrization might not even be needed if we just change the content of `pipeline_config-valid.json` to have comments.

tests/unit/test_pipeline_validation.py
```diff
@@ -49,25 +49,25 @@ def valid_config_data():
 
 def test_load_pipeline_config_file():
     assert isinstance(
-        _load_pipeline_config_file(DPATH_TEST_DATA / "pipeline_config-valid.json"),
+        _load_pipeline_config_file(DPATH_TEST_DATA / "pipeline_config-valid.json5"),
```
michellewang, line 52: `pipeline_config-valid.json` is no longer used. We should either remove it from the repo or change the test to use `pytest.mark.parametrize`

tests/unit/test_pipeline_validation.py
```diff
@@ -49,25 +49,25 @@ def valid_config_data():
...
+}
+""".strip())
 
-    _load_pipeline_config_file(fpath_config, strict=strict)
-
-    mocked_ensure_schema_version_exists.assert_called_once_with(
-        fpath_config, ConfigType.PIPELINE, strict=strict
-    )
+    config = _load_pipeline_config_file(fpath)
+    assert isinstance(config, BasePipelineConfig)
```
michellewang: This is redundant with `test_load_pipeline_config_file` now, no? Should be removed?

tests/unit/test_pipeline_validation.py
```diff
@@ -147,6 +155,20 @@ def test_check_invocation_file(descriptor_str):
...
+    fpath = tmp_path / "invocation.json"
+    fpath.write_text('{"title": true,}')  # trailing comma makes it invalid JSON
+
+    with pytest.raises(
+        ConfigError,
+        match="Invocation file .* is invalid:",
+    ):
+        _check_invocation_file(fpath, descriptor_str)
+
+
```
michellewang: This test is incorrect: we allow JSON5 in invocation files. The `ConfigError` is thrown not because of the trailing comma, but because the invocation file doesn't have the required key from the descriptor.

A valid invocation is:
```json
{
    "basic_param2": "mychoice1.log"
}
```

But in any case we should flip this test so that it ensures that the trailing comma does not make it produce a `ConfigError`

michellewang: Related: can we refactor this and `test_check_invocation_file` in the same way as `test_load_pipeline_config_file`/`test_load_pipeline_config_file_json5`?

tests/unit/test_pipeline_validation.py
```diff
@@ -169,6 +191,18 @@ def test_check_hpc_config_file():
...
+    fpath.write_text("""
+{
+  // comments and trailing commas should be accepted
+  "walltime": "00:10:00",
+}
+""".strip())
+
+    _check_hpc_config_file(fpath)
+
+
```
michellewang: Can we refactor this and `test_check_hpc_config_file` in the same way as `test_load_pipeline_config_file`/`test_load_pipeline_config_file_json5`?

tests/unit/test_pipeline_validation.py
```diff
@@ -189,6 +223,20 @@ def test_check_tracker_config_file():
...
+  // comments and trailing commas should be accepted
+  "PATHS": [
+    "output/sub-[[NIPOPPY_PARTICIPANT_ID]]",
+  ],
+}
+""".strip())
+
+    _check_tracker_config_file(fpath)
+
+
```
michellewang: Can we refactor this and `test_check_tracker_config_file` in the same way as `test_load_pipeline_config_file`/`test_load_pipeline_config_file_json5`?

tests/unit/test_pipeline_validation.py
```diff
@@ -209,6 +257,18 @@ def test_check_pybids_ignore_file():
...
+    fpath.write_text("""
+[
+  // comments and trailing commas should be accepted
+  "^/code",
+]
+""".strip())
+
+    _check_pybids_ignore_file(fpath)
+
+
```
michellewang: Can we refactor this and `test_check_pybids_ignore_file` in the same way as `test_load_pipeline_config_file`/`test_load_pipeline_config_file_json5`?

tests/unit/config/test_main.py
```diff
@@ -252,6 +252,22 @@ def test_load(path):
...
+    "module load apptainer",
+  ],
+}
+""".strip())
+
+    config = Config.load(fpath_config)
+    assert isinstance(config, Config)
+    assert config.HPC_PREAMBLE == ["module load apptainer"]
+
+
```
michellewang: Can we refactor this and `test_load_main` in the same way as `test_load_pipeline_config_file`/`test_load_pipeline_config_file_json5` in `test_pipeline_validation`?

michellewang (changes requested):
Only one of the tests that I previously mentioned was refactored, and in a way that left a `tests/data/` file unused. We should make all the tests consistent.

Also caught a problem in a test that wrongly assessed that invocation files can't be JSON5 even though we allow it.

tests/unit/test_pipeline_validation.py
mathdugre: I think it reappeared with the merge conflict. My bad

tests/unit/test_pipeline_validation.py
mathdugre: Agree! I think the merge conflict failed for this file in my vscode :( 
I made the change in 655686be66825f5d144768ef45a49e91b4211773 and it got reverted.

mathdugre:
> Only one of the tests that I previously mentioned was refactored, and in a way that left a `tests/data/` file unused. We should make all the tests consistent.
> 
> Also caught a problem in a test that wrongly assessed that invocation files can't be JSON5 even though we allow it.

~~Can you explain the refactor you would like to make to these tests?~~
Nvm, I got it

tests/data/invocation-invalid2.json
michellewang, line 1: This doesn't seem to be used anywhere. Should it be a parametrization in `test_pipeline_validation.test_check_invocation_file_invalid`?

michellewang (changes requested):
Thanks, just one last thing: it seems the new `invocation-invalid2.json` file is not used anywhere

tests/data/invocation-invalid2.json
mathdugre: Fixed it in the latest commit.

mathdugre: I was planning to add a parametrization, but realized that it didn't make sense. I forgot to remove the extra file. Good catch.

michellewang (approved):
🧑‍🍳 🎉 🎉 🎉

Issue #937, opened 2026-03-25
[MAINT] Harmonize use of `if X is not None` vs. ` if X`


mathdugre:
Right now we don't have a clear convention check for non-null object. We either do:
- `if X is not None: ...`
- `if X: ...`

Achieving a consensus on which method to use and then standardizing it within the codebase would be nice.

Example:
https://github.com/mathdugre/nipoppy/blob/c9271e65e483f132020145f144daf45e853a9364/nipoppy/workflows/runner.py#L265-L271

note: This could be an easy copilot task.

PR #984, opened 2026-04-14
[ENH] Add general troubleshooting hint on known exception

### LINKED ISSUE #519: [DOC] Improve error messages from Exceptions

- [ ] I have searched the existing issues

### New feature

e.g.

- [ ] When `--dataset` is not passed and the command failed due to dir structure, we could suggest: "Did you forget to pass `--dataset` arg" or "Are you in the right dir", etc.
- [ ] Generally: Suggest possible cause of error and possible fix.
- [ ] Propose potential next step after sucessful command completion.

We could also link the doc of the respective command.

### Unclear documentation

_No response_

mathdugre:
- Closes #519 

- Add a way to pass troubleshooting hint to custom Nipoppy exceptions.

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--984.org.readthedocs.build/en/984/


mathdugre:
@michellewang What do you think of this implementation for passing troubleshooting hints for known/common exceptions.

nipoppy/cli/__init__.py
```diff
@@ -11,6 +11,38 @@
...
+)
+DISCORD_URL = "https://discord.gg/2VMKFRpjkm"
+
+
+def _log_known_error(exception: NipoppyError) -> None:
+    """Log known Nipoppy errors with actionable guidance."""
+    logger.error(exception)
+    hint = exception.troubleshooting_hint
+    if hint:
+        logger.warning(f"Suggested fix: {hint}")
```
michellewang: It feels weird to me that the fix is emitted as a warning. I would lean towards having the fix as a continuation of the exception (i.e. error level). We can experiment with concatenating the fix to the exception message and see how that looks. Not sure about the ones where we call `logger.exception()` though as that will print the entire traceback

mathdugre: I've switched them to INFO level. I think ERROR would be weird since the messages are not quite errors.

nipoppy/exceptions.py
```diff
@@ -40,14 +40,21 @@ class NipoppyError(Exception):
...
+        self.hint = hint
         super().__init__(self.message)
 
     def __str__(self):
         return self.message
 
+    @property
+    def troubleshooting_hint(self) -> str:
+        """Return the troubleshooting hint attached to this error."""
+        return self.hint or self.default_hint
```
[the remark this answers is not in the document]
mathdugre: I think this is valid. @michellewang , do you think we will want to suppress the default hint tho?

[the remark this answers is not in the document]
michellewang: Yes, I think that would be cleaner. There might be situations where the default hint is invalid, though in those cases we probably want to override the hint instead of suppressing it completely. But still it would be cleaner to allow suppression via `""`

nipoppy/cli/__init__.py
```diff
@@ -11,6 +11,11 @@
 
 logger = get_logger()
 
+BUG_REPORT_URL = (
+    "https://github.com/nipoppy/nipoppy/issues/new/choose?template=bug_report.yml"
```
michellewang: We changed the name of that file:

```suggestion
    "https://github.com/nipoppy/nipoppy/issues/new/choose?template=1-bug.yml"
```

nipoppy/cli/__init__.py
```diff
@@ -11,6 +11,11 @@
...
+)
+DISCORD_URL = "https://discord.gg/2VMKFRpjkm"
```
michellewang: I would put these in `nipoppy.env` instead

nipoppy/cli/__init__.py
```diff
@@ -22,18 +27,26 @@ def exception_handler(workflow):
     except NipoppyError as e:
         workflow.return_code = e.code
         logger.error(e)
+        hint = e.troubleshooting_hint
+        if hint:
+            logger.info(f"Suggested fix: {hint}")
     except ValidationError as e:
```
michellewang, line 29: Should we also catch `json.JSONDecodeError`s? When testing locally I broke a config file on purpose and got `This failure was unexpected ...`, which is not really what we should get

nipoppy/exceptions.py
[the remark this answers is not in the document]
michellewang: - In any case doing `self.hint is None or self.default_hint` is consistent with #937

nipoppy/cli/__init__.py
```diff
@@ -22,18 +27,26 @@ def exception_handler(workflow):
     except NipoppyError as e:
         workflow.return_code = e.code
         logger.error(e)
+        hint = e.troubleshooting_hint
+        if hint:
```
michellewang:
```suggestion
        if hint is not None:
```

nipoppy/exceptions.py
```diff
@@ -64,13 +75,23 @@ class TerminatedByUserError(NipoppyError):
     code = ReturnCode.TERMINATED
 
 
-class FileOperationError(NipoppyError, IOError): ...  # noqa: D101, E701
+class FileOperationError(NipoppyError, IOError):
+    """Exception raised for file system operations that fail."""
+
+    default_hint = (
+        "Confirm all input/output paths exist and you have the required file "
```
michellewang: Possibly misleading because some file operations do not want the output path to exist. Maybe change to `Confirm all input/output paths are correct`

nipoppy/exceptions.py
```diff
@@ -64,13 +75,23 @@ class TerminatedByUserError(NipoppyError):
...
+        "permissions."
+    )
 
 
 class WorkflowError(NipoppyError, RuntimeError):
     """Base exception class for workflow-related errors."""
 
     code = ReturnCode.WORKFLOW_FAILURE
+    default_hint = (
+        "Check the workflow arguments and dataset state, then rerun the command "
```
michellewang: "Dataset state" is not really a term we use, could be confusing.

I wonder if the hint can just be "rerun with --verbose"

nipoppy/exceptions.py
```diff
@@ -80,21 +101,37 @@ class ContainerError(NipoppyError):
...
+    )
 
 
 class ExecutionError(NipoppyError, RuntimeError):
     """Exception for pipeline execution errors."""
 
     code = ReturnCode.PIPELINE_EXECUTION_ERROR
+    default_hint = (
+        "Inspect the pipeline logs to locate the failed step, then rerun after "
+        "fixing the reported command or input."
```
michellewang: Do you think "rerun the command" is a useful thing to say? I feel it is obvious that they will need to rerun...

nipoppy/exceptions.py
```diff
@@ -80,21 +101,37 @@ class ContainerError(NipoppyError):
...
+    )
 
 
 class LayoutError(NipoppyError, ValueError):
     """Exception for layout validation errors."""
 
     code = ReturnCode.INVALID_LAYOUT
+    default_hint = (
+        "Check the dataset structure and required Nipoppy layout files, then "
```
michellewang: Suggest changing to "Make sure this command is being run on a valid Nipoppy dataset"

tests/unit/test_cli.py
```diff
@@ -408,11 +408,27 @@ def test_context_manager_nipoppy_exception(mocker, exception):
     mock_exit.assert_called_once_with(exception.code)
 
 
+def test_context_manager_nipoppy_exception_logs_actionable_hint(mocker, caplog):
```
michellewang: Should this test be parametrized for all the possible Nipoppy exception classes?

michellewang (changes requested):
Looks good overall. We should discuss exact wording of the hints. For example I'm not sure how useful it is to tell the user to rerun a command after fixing things.

tests/unit/test_cli.py
mathdugre: I think we could only test on the parent `NipoppyError` exception since the behavior is only defined in that class.
Do you think that makes sense?

michellewang: Ok for testing on `NipoppyError`. I just thought it was weird to only test on `ConfigError`

nipoppy/cli/__init__.py
mathdugre: I implemented a version in 9aaf7d71df271d93924d03a73a51e0da6fddc4e7.

I'm not quite satisfied with it since we rely on the lower level code to pass the file name ; i.e., overwritting the exception message when initially raised. I tried to declare a new custom exception type., but it was without success.
Do you have some ideas for a better implementation? @michellewang

michellewang: What do you think of this?

```python
class NipoppyError(Exception):
    """Base exception class for all nipoppy errors."""

    code = ReturnCode.KNOWN_FAILURE
    default_hint = ""

    def __init__(self, message: str = "", hint: str | None = None):
        self.message = message
        self.hint = hint
        Exception.__init__(self, self.message)

    def __str__(self):
        return self.message

    @property
    def troubleshooting_hint(self) -> str:
        """Return the troubleshooting hint attached to this error."""
        return self.default_hint if self.hint is None else self.hint

class JSONError(NipoppyError, json.JSONDecodeError):
    def __init__(
        self,
        message: str,
        doc: str,
        pos: int,
        filepath: StrOrPathLike | None = None,
        hint: str | None = None,
    ):
        if filepath is not None:
            message = f"{message} (file: {filepath})"

        json.JSONDecodeError.__init__(self, message, doc, pos)

        # self.args[0] is the JSONDecodeError error message
        NipoppyError.__init__(self, self.args[0], hint)
```

michellewang (approved):

PR #985, opened 2026-04-16
[ENH] Add `NipoppyDataRetriever.get_all_phenotypes()` method

### LINKED ISSUE #980: [ENH] Expose full phenotype table (from `harmonized.tsv`) in public API?

### New feature

See @sharkovsky's comment in https://github.com/michellewang/fedbiomed/pull/1#discussion_r3064854988:

> Would you be open to adding something like this to the NipoppyDataRetriever?
> 
>  ```python
>    def get_phenotypes_schema(self) -> Dict[str, type]:
>        """Get the schema (column names and data types) of the harmonized phenotypes TSV file.
>        """  # noqa E501
>        df = self._load_tsv(
>            self._study.layout.fpath_harmonized, index_cols=self._index_cols_phenotypes
>        )
>        return {col: df[col].dtype for col in df.columns}
> ```
>
>It's an easy and fbm-compatible way to list all phenotypic variables and their dtype.

I think there would be limited use for a method that just returns the schema/datatypes, but it could be useful to allow users to get the full phenotypic data without having to specify any columns in advance. Maybe we could add a `get_all_phenotypes() -> pd.DataFrame` which is then called by `get_phenotypes()`.

### Unclear documentation

_No response_

michellewang:
- Closes #980

- 

## Checklist (for reviewers)

- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--985.org.readthedocs.build/en/985/


nipoppy/_data_retriever.py
mathdugre, line 193: It would be nice for the example so the output is not the same as with the `get_all_phenotypes` function

mathdugre (changes requested):
Minor comment, otherwise LGTM

mathdugre (approved):
LGTM 🧑‍🍳

PR #991, opened 2026-04-21
[MAINT] Extract tar dir function to fileops module

### LINKED ISSUE #888: [MAINT] Decouple Tar functionality

mathdugre:
- Closes #888 

Move the`tar_directory` function from the ProcessingRunner to the fileops module.

Also moved the `run_command` function from the base workflow module to a dedicated new module. Otherwise, there were issue with circular dependency.

## Checklist (for reviewers)

- [x] Tests pass
- [x] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:

----
📚 Documentation preview 📚: https://nipoppy--991.org.readthedocs.build/en/991/

### DIFF
--- a/nipoppy/utils/fileops.py
+++ b/nipoppy/utils/fileops.py
@@ -2,13 +2,13 @@
 
 import errno
 import shutil
-import subprocess
 from pathlib import Path
 from tarfile import is_tarfile
 
 from nipoppy.env import EXT_TAR
 from nipoppy.exceptions import FileOperationError
 from nipoppy.logger import get_logger
+from nipoppy.utils.subprocess_runner import run_command
 
 logger = get_logger()
 
@@ -102,11 +102,12 @@ def tar_directory(dpath: Path, dry_run: bool = False) -> Path:
 
     fpath_tarred = dpath.with_suffix(EXT_TAR)
     logger.debug(f"Tarring {dpath} to {fpath_tarred}")
+    run_command(
+        ["tar", "-cvf", str(fpath_tarred), "-C", str(dpath.parent), dpath.name],
+        check=True,
+        dry_run=dry_run,
+    )
     if not dry_run:
-        subprocess.run(
-            ["tar", "-cvf", str(fpath_tarred), "-C", str(dpath.parent), dpath.name],
-            check=True,
-        )
         if not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
             raise FileOperationError(f"Failed to tar {dpath} to {fpath_tarred}")
 
--- a/nipoppy/utils/fileops.py
+++ b/nipoppy/utils/fileops.py
@@ -2,8 +2,11 @@
 
 import errno
 import shutil
+import subprocess
 from pathlib import Path
+from tarfile import is_tarfile
 
+from nipoppy.env import EXT_TAR
 from nipoppy.exceptions import FileOperationError
 from nipoppy.logger import get_logger
 
@@ -68,6 +71,46 @@ def _ignore_oserror_empty_dir(function, path, excinfo):
     if isinstance(exception, OSError) and exception.errno == errno.ENOTEMPTY:
         return
     raise exception
+
+
+def tar_directory(dpath: Path, dry_run: bool = False) -> Path:
+    """Create a tarball of a directory.
+
+    The tarball is created in the same parent directory as the original directory,
+    with the same name and a .tar extension.
+
+    The original directory is NOT removed after tarring. The caller is responsible for
+    removing the original directory if desired.
+
+    Parameters
+    ----------
+    dpath: Path
+        Path to the directory to tar.
+    dry_run: bool, optional
+        If True, do not actually perform the tarring operation, just log it. Default is
+        False.
+
+    Returns
+    -------
+    Path
+        Path to the tarball.
+    """
+    if not dpath.exists():
+        raise FileOperationError(f"Dir does not exist: {dpath}")
+    if not dpath.is_dir():
+        raise FileOperationError(f"Cannot tar non-directory: {dpath}")
+
+    fpath_tarred = dpath.with_suffix(EXT_TAR)
+    logger.debug(f"Tarring {dpath} to {fpath_tarred}")
+    if not dry_run:
+        subprocess.run(
+            ["tar", "-cvf", str(fpath_tarred), "-C", str(dpath.parent), dpath.name],
+            check=True,
+        )
+        if not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
+            logger.error(f"Failed to tar {dpath} to {fpath_tarred}")
+
+    return fpath_tarred
 
 
 def rm(path: Path, dry_run=False):
--- a/nipoppy/utils/fileops.py
+++ b/nipoppy/utils/fileops.py
@@ -108,7 +108,7 @@ def tar_directory(dpath: Path, dry_run: bool = False) -> Path:
             check=True,
         )
         if not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
-            logger.error(f"Failed to tar {dpath} to {fpath_tarred}")
+            raise FileOperationError(f"Failed to tar {dpath} to {fpath_tarred}")
 
     return fpath_tarred
 
--- a/nipoppy/utils/fileops.py
+++ b/nipoppy/utils/fileops.py
@@ -96,7 +96,7 @@ def tar_directory(dpath: Path, dry_run: bool = False) -> Path:
         Path to the tarball.
     """
     if not dpath.exists():
-        raise FileOperationError(f"Dir does not exist: {dpath}")
+        raise FileOperationError(f"Directory does not exist: {dpath}")
     if not dpath.is_dir():
         raise FileOperationError(f"Cannot tar non-directory: {dpath}")
 
--- a/nipoppy/utils/fileops.py
+++ b/nipoppy/utils/fileops.py
@@ -107,9 +107,8 @@ def tar_directory(dpath: Path, dry_run: bool = False) -> Path:
         check=True,
         dry_run=dry_run,
     )
-    if not dry_run:
-        if not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
-            raise FileOperationError(f"Failed to tar {dpath} to {fpath_tarred}")
+    if not dry_run and not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
+        raise FileOperationError(f"Failed to tar {dpath} to {fpath_tarred}")
 
     return fpath_tarred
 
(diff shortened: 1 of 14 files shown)


tests/unit/utils/test_subprocess_runner.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def test_run_command_quiet(caplog: pytest.LogCaptureFixture):
+    message = "This should be printed"
+    run_command(["echo", message], quiet=True)
+    assert LogPrefix.RUN not in caplog.text
+    assert message in caplog.text
+
+
+def test_run_command_streams_final_lines(
```
mathdugre, line 84: Only this test was added. Everything else was moved from `tests/unit/workflows/test_base.py`

nipoppy/utils/subprocess_runner.py
mathdugre, line 1: Had to move the `run_command` function from base because of circular dependency.

nipoppy/utils/fileops.py
```diff
@@ -70,6 +73,47 @@ def _ignore_oserror_empty_dir(function, path, excinfo):
...
+        If True, do not actually perform the tarring operation, just log it. Default is
+        False.
+
+    Returns
+    -------
+    Path
+        Path to the tarball.
+    """
+    if not dpath.exists():
+        raise FileOperationError(f"Dir does not exist: {dpath}")
```
michellewang:
```suggestion
        raise FileOperationError(f"Directory does not exist: {dpath}")
```

nipoppy/utils/fileops.py
```diff
@@ -70,6 +73,47 @@ def _ignore_oserror_empty_dir(function, path, excinfo):
...
+    if not dpath.is_dir():
+        raise FileOperationError(f"Cannot tar non-directory: {dpath}")
+
+    fpath_tarred = dpath.with_suffix(EXT_TAR)
+    logger.debug(f"Tarring {dpath} to {fpath_tarred}")
+    run_command(
+        ["tar", "-cvf", str(fpath_tarred), "-C", str(dpath.parent), dpath.name],
+        check=True,
```
michellewang, line 107: Is this necessary? Seems redundant with the check below

nipoppy/utils/fileops.py
```diff
@@ -70,6 +73,47 @@ def _ignore_oserror_empty_dir(function, path, excinfo):
...
+        dry_run=dry_run,
+    )
+    if not dry_run:
+        if not (fpath_tarred.exists() and is_tarfile(fpath_tarred)):
```
michellewang: Merge into a single `if`?

nipoppy/utils/subprocess_runner.py
michellewang: I am not a fan of calling this a "runner" since it is not a class, and the other runners (I think) are all classes

nipoppy/utils/subprocess_runner.py
```diff
@@ -0,0 +1,149 @@
...
+import shlex
+import subprocess
+from typing import Protocol, Sequence
+
+from nipoppy.logger import get_logger
+
+logger = get_logger()
+
+
+class LogPrefix:
```
michellewang: I think we could make this private

nipoppy/utils/subprocess_runner.py
```diff
@@ -0,0 +1,149 @@
...
+            line = line.strip("\n")
+            # using extra={"markup": False} in case the output contains substrings
+            # that would be interpreted as closing tags by the RichHandler
+            logger.log(
+                level=log_level,
+                msg=f"{log_prefix} {line}",
+                extra={"markup": False},
+            )
+
+    def drain_process(process: subprocess.Popen[str]):
```
michellewang: @mathdugre is "draining" a standard term? Plus: it's not the process that's being drained, but rather its output. Is this helper function useful if it is only called once?

nipoppy/utils/subprocess_runner.py
```diff
@@ -0,0 +1,149 @@
...
+        **kwargs,
+    )
+
+    while process.poll() is None:
+        drain_process(process)
+
+    # final drain: the poll() loop can exit before the last lines written
+    # just prior to process termination are read, so flush any remaining
+    # buffered output now that the process has exited.
+    drain_process(process)
```
michellewang: Is this true?

nipoppy/utils/subprocess_runner.py
```diff
@@ -0,0 +1,149 @@
...
+            process.stdout,
+            LogPrefix.RUN_STDOUT,
+        )
+        process_output(
+            process.stderr,
+            LogPrefix.RUN_STDERR,
+            log_level=logging.ERROR,
+        )
+
+    def build_cmd(command_or_args: Sequence[str] | str) -> tuple[str, list[str]]:
```
michellewang: If we are going to have that many nested functions, I am wondering if it would not be better to move them out of the parent function

nipoppy/utils/subprocess_runner.py
```diff
@@ -0,0 +1,149 @@
...
+        *,
+        check: bool = True,
+        quiet: bool = False,
+        dry_run: bool = False,
+    ) -> subprocess.Popen[str] | str:
+        """Run a command in a subprocess, with logging and dry-run support."""
+        ...
+
+
+def run_command(
```
michellewang, line 126: Is there a reason why this is no longer `_run_command`?

nipoppy/workflows/base.py
```diff
@@ -30,131 +27,6 @@
...
-        run_output = process
-
-    else:
-        run_output = command
-
-    return run_output
-
-
 class BaseWorkflow(Base, ABC):
     """Base workflow class with logging/subprocess/filesystem utilities."""
```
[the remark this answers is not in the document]
michellewang: @mathdugre this seems valid

nipoppy/workflows/processing_runner.py
```diff
@@ -214,9 +187,12 @@ def run_single(self, participant_id: str, session_id: str):
                     session_id=session_id,
                 )
             )
-            self.tar_directory(
+            dpath_to_tar = (
                 self.dpath_pipeline_output / tracker_config.PARTICIPANT_SESSION_DIR
             )
+            fpath_tarred = fileops.tar_directory(dpath_to_tar, dry_run=self.dry_run)
+            if fpath_tarred.exists():
```
michellewang, line 194: Is it necessary to check if the tarball exists?

tests/unit/utils/test_fileops.py
```diff
@@ -229,3 +230,71 @@ def test_symlink(self, tmp_path: Path):
...
+def dpath_to_tar(tmp_path: Path) -> Path:
+    """Return a temporary directory with nested files, ready to be tarred."""
+    dpath = tmp_path / "my_data"
+    (dpath / "dir1").mkdir(parents=True)
+    (dpath / "dir1" / "file1.txt").touch()
+    (dpath / "file2.txt").touch()
+    return dpath
+
+
+class TestTarDirectory:
```
michellewang, line 245: We should discuss this more for the refactor. I don't like that we have a mix of test classes and functions

tests/unit/utils/test_fileops.py
```diff
@@ -229,3 +230,71 @@ def test_symlink(self, tmp_path: Path):
...
+    def test_tar_directory(self, dpath_to_tar: Path):
+        """Tarball is created with correct contents; source directory is left intact."""
```
michellewang, line 247: This docstring technically doesn't respect PEP conventions I think

tests/unit/utils/test_fileops.py
```diff
@@ -229,3 +230,71 @@ def test_symlink(self, tmp_path: Path):
 
         assert symlink.is_symlink()
         assert symlink.read_text() == expected_content
+
+
+@pytest.fixture
+def dpath_to_tar(tmp_path: Path) -> Path:
+    """Return a temporary directory with nested files, ready to be tarred."""
```
michellewang:
```suggestion
    """Return a temporary directory with nested files."""
```

nipoppy/utils/fileops.py
```diff
@@ -70,6 +73,47 @@ def _ignore_oserror_empty_dir(function, path, excinfo):
...
+            raise FileOperationError(f"Failed to tar {dpath} to {fpath_tarred}")
```
michellewang: This is a different behaviour from before, and we didn't discuss this. Currently we don't raise an error on tar failure because this can happen if the pipeline itself crashed, and we still want to run the rest of the Nipoppy workflow.

Similar for the `check=True` actually: I don't think we should have it. Or at least we should have this be toggle-able when calling the function.

tests/unit/utils/test_fileops.py
```diff
@@ -229,3 +230,71 @@ def test_symlink(self, tmp_path: Path):
...
+            fileops.tar_directory(dpath_to_tar)
+
+    @pytest.mark.parametrize(
+        "make_path,match",
+        [
+            (lambda p: p / "nonexistent", "Dir does not exist:"),
+            (
+                lambda p: (p / "file.txt").touch() or p / "file.txt",
+                "Cannot tar non-directory:",
+            ),
```
michellewang, line 281: We could have the parametrized arg only be the filename and do `(p/"file.txt").touch()` inside the test function, no? I feel that would be more readable.

tests/unit/utils/test_subprocess_runner.py
michellewang, line 1: This lost all of the `pytest.mark.no_xdist`: on purpose or no?

tests/unit/utils/test_subprocess_runner.py
```diff
@@ -0,0 +1,122 @@
...
+    caplog: pytest.LogCaptureFixture, monkeypatch: pytest.MonkeyPatch
+):
+    """The final drain ensures the last buffered lines are logged.
+
+    Previously the `while process.poll() is None:` loop could exit before
+    reading lines written just prior to process termination; this test
+    guards against regressions of that fix.
```
michellewang, line 91: If this happened only in Python <3.10 then I don't think we need it, @mathdugre can you look into this?

michellewang (changes requested):
Looks okay overall, some choices regarding subprocess exit code checking need to be further discussed

nipoppy/utils/fileops.py
mathdugre: I not sure to understand this comment. Do you remember what change you wanted here?

michellewang: Not 100% sure. Maybe it was about the debug logging, since `run_command` already logs the full command at debug level. So having the extract `logger.debug` is slightly redundant

Issue #683, opened 2025-07-10
[MAINT] Refactor initWorkflow to reduce function complexity


mathdugre:
Flake8 error from #682 
```
nipoppy/workflows/dataset_init.py:58:5: C901 'InitWorkflow.run_main' is too complex (13)
```

https://github.com/asmacdo/nipoppy/blob/89322e59f12cd0c3235b24b85f9841ed3f76df2b/nipoppy/workflows/dataset_init.py#L58

PR #1017, opened 2026-05-17
[ENH] Add `--container-store` argument to `nipoppy init`

### LINKED ISSUE #1008: [ENH] Allow configuration of the container store path during `init`

### New feature

We should have a `--container-store option` for the init command to specify the location of an share container store. Configuration via an env var (e.g., NIPOPPY_CONTAINER_STORE) should be allowed.
A symlink should be created to link the container store (source) to the containers dir of the study (target).

### Unclear documentation

_No response_

### LINKED ISSUE #1056: [BUG] `nipoppy init` should not fail if there is no network connection

### Describe the bug

Running `nipoppy init --bids-source` on a machine that doesn't have Internet connection results in a crash because we do an HTTP request to get the template/default `README.md` from the BIDS website.

It would be better to not crash on timeout/unreachable network.

### To reproduce

`nipoppy init --bids-source ...` in a Compute Canada interactive job

### Environment

_No response_

michellewang:
- Closes #1008
- Closes #1056 (removes HTTP request)
- Related to #683

## Changes
- Add `container_store` arg and `_handle_container_store()` method to workflow
- Add `--container-store` arg to CLI with env var support `NIPOPPY_CONTAINER_STORE`
- Reduce complexity of `run_main()` by extracting code block that checked if an existing study root directory was allowed
- Change logic of README writing so that the directories that undergo special handling (BIDS directory and containers directory) do not have READMEs written to them

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1017.org.readthedocs.build/en/1017/


nipoppy/cli/cli.py
```diff
@@ -135,6 +127,21 @@ def cli():
         "If using a BIDS source, specify whether to copy, move, or symlink the files."
     ),
 )
+@click.option(
+    "--container-store",
+    type=click.Path(exists=True, file_okay=False, path_type=Path, resolve_path=True),
+    help="Path to an existing (shared) directory for container image files.",
```
mathdugre: 🍒 We could just say "container images" instead of "container image files"; container images are always files.

nipoppy/workflows/dataset_init.py
```diff
@@ -102,35 +102,24 @@ def run_main(self):
         Copy default config files.
         Copy HPC config files.
         """
-        # dataset must not already exist
         if self.dpath_root.exists():
```
mathdugre: We might want to move the condition inside the extracted function.
Then, the function name could be `_raise_on_invalid_init_dir` or something alike.

nipoppy/workflows/dataset_init.py
```diff
@@ -211,33 +217,22 @@ def handle_bids_source(self) -> None:
...
-                        "Skipping README creation."
-                    )
+    def _handle_container_store(self) -> None:
+        dpath = self.study.layout.dpath_containers
+
+        # delete existing if needed
+        if dpath.exists() and self.force:
+            fileops.rm(dpath, dry_run=self.dry_run)
+
+        fileops.mkdir(dpath.parent, dry_run=self.dry_run)
```
mathdugre: This should be moved inside the `fileops.symlink` function. We can add a `force: bool = False` argument to the function signature.

We should add a test creating a symlink inside a non-existing directory, to ensure we handle the creation of the missing parent directories.

nipoppy/layout.py
```diff
@@ -206,96 +206,96 @@ def __init__(
...
                 self.config.fpath_bids_dataset_description.path
             )
         if self.config.fpath_bidsignore is not None:
-            self.fpath_bidsignore: Path = self.get_full_path(
+            self.fpath_bidsignore: Path = self._get_full_path(
                 self.config.fpath_bidsignore.path
             )
 
-    def get_full_path(self, path: StrOrPathLike) -> Path:
+    def _get_full_path(self, path: StrOrPathLike) -> Path:
```
mathdugre: Should we rename this to **absolute** path? The term is more common.

tests/unit/workflows/test_init.py
```diff
@@ -102,24 +102,17 @@ def _assert_manifest_creation(
...
 
-def exist_or_none(o: object, s: str) -> bool:
-    # walrus operator ":=" does assignment inside the "if" statement
-    if attr := getattr(o, s, None):
-        return attr.exists()
-    return True
-
-
-def assert_layout_creation(workflow, dpath_root):
+def _assert_layout_creation(workflow):
```
mathdugre, line 112: I don't think we need to mark the test functions as private.

nipoppy/workflows/dataset_init.py
mathdugre: Afterwards, I don't think the function would be needed. We can inline the symlink at the container condition [here](https://github.com/nipoppy/nipoppy/pull/1017/changes#diff-56301dfefdf30f3489619572e91993f510b4be871e0a939777bd40c6e7d08a97R115) .

tests/unit/workflows/test_init.py
```diff
@@ -187,7 +180,50 @@ def test_init_twice_force(workflow: InitWorkflow):
...
+    workflow.force = True
+    workflow.study.layout.dpath_containers.mkdir()
+
+    workflow._handle_container_store()
+
+    assert workflow.study.layout.dpath_containers.is_symlink()
+    assert (
+        workflow.study.layout.dpath_containers.resolve()
+        == workflow.container_store.resolve()
+    )
```
mathdugre: This won't be needed if moving the `force` check in the symlink function.

**note:** related to above comment

tests/unit/workflows/test_init.py
```diff
@@ -187,7 +180,50 @@ def test_init_twice_force(workflow: InitWorkflow):
...
+    workflow: InitWorkflow, tmp_path: Path, mocker: pytest_mock.MockerFixture
+):
+    workflow.container_store = tmp_path / "container_store"
+
+    mocked = mocker.patch.object(
+        workflow, "_handle_container_store", wraps=workflow._handle_container_store
+    )
+
+    workflow.run_main()
+    mocked.assert_called_once()
```
mathdugre: In the future, this might be better as an integration test.

mathdugre (changes requested):
Mostly changes related to the behavior of the `fileops.symlink` function, and a few nitpicks

nipoppy/layout.py
michellewang: It's not necessarily absolute though: might not start with `/` if `dpath_root` is relative

michellewang: E.g.:

```
>>> from nipoppy.layout import DatasetLayout
>>> layout = DatasetLayout('.')
>>> layout.dpath_bids
PosixPath('bids')
```

tests/unit/workflows/test_init.py
michellewang: If you don't have strong feelings against it, I would prefer having them private since that makes them easier to differenciate with the test functions

nipoppy/layout.py
mathdugre: How about `_preprend_study_path()` then?

Although, it might be clearer to remove the function altogether and use `self.dpath_root / <path>` direclty.

tests/unit/workflows/test_init.py
mathdugre: I find it a bit verbose and confusing, but I won't block the PR over this

tests/unit/utils/test_fileops.py
```diff
@@ -217,15 +217,32 @@ def test_mv_directory(self, tmp_path: Path):
...
 
-        symlink = tmp_path / "symlink_to_target"
+        symlink = tmp_path / target_path_relative
 
         fileops.symlink(source=source_file, target=symlink)
 
         assert symlink.is_symlink()
         assert symlink.read_text() == expected_content
+
+    def test_symlink_force(self, tmp_path: Path):
```
mathdugre: I think we would need another test to also verify that we don't overwrite the target path when `force=False`.

michellewang: @mathdugre right now we don't do any explicit handling of this in `fileops.symlink`. So we would essentially be testing the behaviour of the underlying `Path.symlink_to`, which I don't think is very useful. Do you want to explicitly raise a Nipoppy `FileOperationError` instead? Otherwise `Path.symlink_to` will throw a `FileExistsError`

nipoppy/layout.py
```diff
@@ -207,80 +207,86 @@ def __init__(
...
                 self.config.fpath_bids_dataset_description.path
             )
         if self.config.fpath_bidsignore is not None:
-            self.fpath_bidsignore: Path = self._get_full_path(
+            self.fpath_bidsignore: Path = self._prepend_root_path(
                 self.config.fpath_bidsignore.path
             )
 
-    def _get_full_path(self, path: StrOrPathLike) -> Path:
+    def _prepend_root_path(self, path: StrOrPathLike) -> Path:
```
mathdugre: `root` can be ambiguous here—it could be the `/` of the filesystem. We could use `study` or `study_root` instead.

mathdugre (commented):
Minor comment, otherwise LGTM

nipoppy/workflows/dataset_init.py
```diff
@@ -107,35 +108,26 @@ def run_main(self):
...
         fileops.mkdir(self.dpath_root / NIPOPPY_DIR_NAME, dry_run=self.dry_run)
         for dpath in self.study.layout.get_paths(directory=True, include_optional=True):
             if self.bids_source is not None and dpath == self.study.layout.dpath_bids:
-                self.handle_bids_source()
+                self._handle_bids_source()
+            elif (
+                self.container_store is not None
+                and dpath == self.study.layout.dpath_containers
+            ):
+                fileops.mkdir(dpath.parent, dry_run=self.dry_run)
```
mathdugre: This check should be done inside the symlink function and have a comment to indicate the parent dir of the symlink needs to be created before creating the symlink itself.

nipoppy/workflows/dataset_init.py
```diff
@@ -188,7 +180,26 @@ def run_main(self):
 
         logger.success(f"Successfully initialized a dataset at {self.dpath_root}!")
 
-    def handle_bids_source(self) -> None:
+    def _raise_on_invalid_existing_root_dir(self) -> None:
```
mathdugre: Maybe `_validate_study_root_for_initialization` instead?

mathdugre (changes requested):
Minor comments, otherwise LGTM

nipoppy/workflows/dataset_init.py
michellewang: @mathdugre I'm not sure I understand this comment. Are you saying we should move the `mkdir` call inside `symlink`? I can implement that but could you clarify where you want the comment to be?

mathdugre: Yes, the `mkdir` should be inside the `symlink` function. If needed, a comment could be added to the `symlink` function to explain why we do the `mkdir` to ensure the parent dir of the target exists.

tests/unit/utils/test_fileops.py
```diff
@@ -220,7 +220,7 @@ class TestSymlink:
     @pytest.mark.parametrize(
         "target_path_relative", ["symlink_to_target", "parent_dir/symlink_to_target"]
     )
-    def test_symlink(self, target_path_relative: str, tmp_path: Path):
+    def test_file(self, target_path_relative: str, tmp_path: Path):
```
mathdugre: We should keep the `test_symlink_*` name for those tests, to be consistent with the other ones in this file.

mathdugre (changes requested):
🍒 Minor comment only. Almost there :D

tests/unit/utils/test_fileops.py
michellewang: Ok, but this is redundant with the class name

mathdugre (approved):

PR #1019, opened 2026-05-18
[ENH] Warn when participant ID still starts with 'sub' after stripping prefix

### LINKED ISSUE #559: [ENH] Warn if `participant_id` in manifest has `sub` in it
> I still think we should warn against `sub-sub123`. But up to you.

_Originally posted by @nikhil153 in https://github.com/nipoppy/nipoppy/pull/546#pullrequestreview-2703376862_

laurapiro17:
- Closes #559

## Changes

`check_participant_id` now emits a `UserWarning` when the ID still starts with `sub` after the BIDS `sub-` prefix has been stripped. Catches the two common slip-ups from the issue:

- `sub-sub123` → stripped to `sub123` → warns
- `sub123` (no dash) → no strip → warns

The warning message is intentionally generic (it doesn't include the offending ID), so Python's default warnings filter deduplicates it by `(message, module, lineno)`. A manifest with 500 problematic IDs produces **one** warning, not 500 — addresses @michellewang's concern in the issue thread.

Plain IDs (`01`, `P01`, `sub-01`) are unaffected.

Tests added:
- `test_check_participant_id_warns_on_sub_prefix` — covers `sub-sub123`, `sub123`, `SUB01` (case-insensitive)
- `test_check_participant_id_no_warning` — covers `01`, `sub-01`, `P01`, `None` (turns `UserWarning` into an error to assert silence)

----
📚 Documentation preview 📚: https://nipoppy--1019.org.readthedocs.build/en/1019/

### DIFF
--- a/nipoppy/utils/bids.py
+++ b/nipoppy/utils/bids.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import re
+import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Optional, Sequence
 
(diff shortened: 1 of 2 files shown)


michellewang:
Please ensure to follow the contributing guide; specifically the AI policy section and strictly follow the PR template omitting no sections.
We will re-consider the PR once those requirements are met.

PR #1023, opened 2026-05-22
[ENH] Allow users to configure environment variables via `.env` files

### LINKED ISSUE #910: [ENH] Read environment variables from user- and study-level `.env` files
From lowest to higher priority, we will have:
1. User-level `.env` file at `~/.nipoppy/.env`? Or `~/.config/nipoppy/.env`? What is the more common practice?
2. Study-level `.env` file at `<NIPOPPY_ROOT>/.env`
3. Actual environment variables
4. CLI options specified at runtime

michellewang:
- Closes #910

## Changes
- Users can now use `.env` files at multiple levels (system, user, and project, with project having the highest priority) for CLI options that can be set through environment variables
- Introduces new dependency `python-dotenv` (which does not have any required dependencies)
- ~The implementation is slightly hacky because it relies on defining a hidden Click option (which I named `--_env`). This had to be done because the path to the project-level `.env` file requires knowledge about the Nipoppy root directory, which requires `--dataset` to be parsed already, but the `.env` file needs to be loaded before the rest of the CLI options are parsed~
- ~New implementation relies on a decorator that is added to all Click groups (`cli` and `pipeline` for now). This decorator will load the dotenv files **if and only if** the subcommand is not a group itself and the subcommand does not have any of the dataset params. Otherwise the `.env` file loading is deferred in a callback that has been added to both dataset params (which are now both eager as well)~
- New new implementation: have dummy CLI to pre-parse the dataset option, then override the CLI group's `make_context`
- Some adjustments to the logger so that the `python-dotenv` logger outputs are formatted the same as the Nipoppy logs
- New How-to guide explaining how the env files work (not sure this is the right place to put that information)

## Checklist

### All PRs
- [x] Tests pass
- [x] Checks pass

### If new feature
- [x] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1023.org.readthedocs.build/en/1023/


nipoppy/cli/options.py
```diff
@@ -2,14 +2,60 @@
...
+        raise ctx.fail(
+            f"The {param.opts[0]} option exists for internal reasons and should never be used on the command-line."  # noqa: E501
+        )
+
+    # Nipoppy root directory
+    dpath_root = ctx.params.get("dataset_argument") or ctx.params.get("dpath_root")
+    if dpath_root is not None:
+        dpath_root = is_nipoppy_project(dpath_root) or dpath_root
+    else:
+        dpath_root = Path.cwd()
```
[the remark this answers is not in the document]
michellewang: Incorrect because `dpath_root` is `Path.cwd()` by default

michellewang:
~Some old CLI tests are failing, but I'm unable to reproduce the failures locally. @mathdugre could you try on your side?~ Nevermind, reproduced them

michellewang:
Ok, ready for review @mathdugre!

docs/source/how_to_guides/configuration/index.md
```diff
@@ -0,0 +1,19 @@
...
+Some {term}`CLI` options, such as `--layout`, can be configured via environment variables.
+This will be indicated in the usage message when running `nipoppy <SUBCOMMAND> --help`.
+
+It is also possible to set these options using `.env` environment files.
+By default, Nipoppy searches the following locations (in order of decreasing priority):
+- {{dpath_root}}`/.env`: project-level
+- `~/.nipoppy/.env`: user-level
+- `/etc/nipoppy/.env`: system-level
+
+Content from these files has lower priority compared to already-defined environment variables, which themselves can be overridden by explicitly passing CLI options.
```
mathdugre: I think we should have an ordered list to showcase the priorities. It feels easier to parse through quickly compared to a sentence.

nipoppy/cli/options.py
```diff
@@ -2,14 +2,60 @@
...
+from nipoppy.utils.utils import is_nipoppy_project, process_template_str
 
 logger = get_logger()
 
+# from highest to lowest priority
+DEFAULT_DOTENV_PATHS_LIST = [
+    "[[NIPOPPY_DPATH_ROOT]]/.env",
+    "~/.nipoppy/.env",
+    "/etc/nipoppy/.env",
+]
```
mathdugre: Move this to `nipoppy.env`?

docs/source/how_to_guides/configuration/index.md
```diff
@@ -0,0 +1,19 @@
+# Configuring projects
+
+This guide describes advanced configuration methods.
```
mathdugre: I think this could be more concise. Maybe a single title; e.g., `Advanced Study Configuration` or similar

tests/unit/test_cli.py
```diff
@@ -505,3 +527,103 @@ def test_cli_params_match_workflows(command_name):
...
+        (
+            "TEST_PARAM='dotenv_global'",
+            "TEST_PARAM='dotenv_local'",
+            {},
+            [],
+            "dotenv_local",
+        ),
+        ("TEST_PARAM='dotenv_global'", "", {}, [], "dotenv_global"),
+        ("TEST_PARAM='dotenv_global'", None, {}, [], "dotenv_global"),
+        ("", None, {}, [], "default"),
```
mathdugre: We should put the `"default"` in a var, otherwise it's not obvious where it comes from.

tests/unit/test_cli.py
```diff
@@ -505,3 +527,103 @@ def test_cli_params_match_workflows(command_name):
...
+        in RE_ANSI.sub("", result.output)
+    )
+
+
+@pytest.mark.parametrize(
+    "command_name",
+    [
+        command
+        for command in list_cli_commands(cli)
+        if command not in ("gui", "pipeline")
```
mathdugre: We can use the `include_hidden=False` to remove `gui`. We probably would want to have a `include_group` arg as well to hidden parent commands (e.g., `pipeline`)

docs/source/how_to_guides/configuration/index.md
michellewang: I thought about that but it breaks the pattern of all how-to guides so far starting with a verb like "initializing", "creating", etc.

michellewang: TODO make sidebar still show "Configuring projects" but page title can be "Advanced study configuration"

michellewang:
@mathdugre I changed the implementation to be less hacky, let me know if it makes sense

michellewang:
~@mathdugre it seems the shell completion for `--dataset` doesn't work with this latest version, at least on my side...~

Update: fixed :)

mathdugre (approved):
LGTM 🧑‍🍳

PR #1027, opened 2026-05-28
[ENH] Add injection of BIDS file paths based on PyBIDS database

### LINKED ISSUE #1022: [ENH] BIDS filter support for non-BIDS-App pipelines

### New feature

I have been thinking about this issue faced by many Nipoppy users who want to add their own pipeline to Nipoppy, where the pipeline is not a BIDS App and instead takes e.g. a single T1w NIfTI file, which means that the invocation likely has to be modified for each dataset (“best case” scenario) or for each subject. Examples include FreeSurfer `recon-all`. 

NiPreps BIDS Apps (and other BIDS Apps like QSIPrep) are able to get around this issue (most of the time) by accepting a BIDS filter file or a PyBIDS database directly. Is there a way to easily integrate something flexible like this for the average Nipoppy pipeline config creator? Maybe making use of the PyBIDS database since that is something that Nipoppy already uses.

A reasonable use-case would be:
1. I have made sure that the PyBIDS database created by Nipoppy will only have a single T1w per participant/session
2. I want to inject the path to this T1w image into the invocation file

Probably this can be achieved with an additional pipeline configuration file (to be specified by the dev). Something like:
```yaml
{
    # keys are placeholders in invocation file 
    # values are some BIDS entity/kwarg filters passed to PyBIDS bids.layout.BIDSLayout.get()
    "KEY1": {
        "extension": "nii.gz",
        "suffix": "T1w"
    }
{
```

Then in the invocation file we can have something like this (modified from https://zenodo.org/records/18488207):
```json
{
    "subjid": "[[NIPOPPY_BIDS_PARTICIPANT_ID]]_[[NIPOPPY_BIDS_SESSION_ID]]",
    "input": "[[NIPOPPY_PYBIDS_DATABASE_KEY1]]",
    "outdirname": "[[NIPOPPY_DPATH_PIPELINE_OUTPUT]]",
    "directive": "-all"
}
```

On the Nipoppy side, we
1. do a `layout.get(return_type="filename", **kwargs)` call for each key
2. check that there is only 1 file that matches per key
3. treat is as regular substitution and inject in the invocation file

There would still be the problem of datasets that having multiple T1w files and needing to pick different ones for different subjects, which technically can be solved through PyBIDS ignore regex patterns but it would be very ugly.

_Originally posted by @michellewang in https://github.com/nipoppy/nipoppy/discussions/890_

michellewang:
- Closes #1022

## Changes
- `ProcessingPipelineConfig` now has an extra field `BIDS_PATH_INJECTION_MAP` (open to a better name)
    - I put it at the pipeline level instead of the pipeline step level because `VARIABLES` is also at the pipeline level. But open to discuss
- During `nipoppy process`, if `BIDS_PATH_INJECTION_MAP` is defined (and `GENERATE_PYBIDS_DATABASE` is `True`), new replacement keys for e.g. the invocation file will become available based on `BIDSLayout.get`

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1027.org.readthedocs.build/en/1027/


nipoppy/config/pipeline.py
```diff
@@ -187,6 +188,42 @@ class ProcessingPipelineConfig(BasePipelineConfig):
...
+    )
+
+    @model_validator(mode="after")
+    def validate_after(self):
+        """
+        Validate the pipeline config after instantiation.
+
+        Specifically:
+        - Make sure that BIDS_PATH_INJECTION_MAP entries have valid keys and do not
+          use reserved keywords
```
mathdugre: We shouldn't use a list for a single entry. 
```suggestion
        Specifically, make sure that BIDS_PATH_INJECTION_MAP entries have valid keys and do not
          use reserved keywords
```

nipoppy/config/pipeline.py
```diff
@@ -187,6 +188,42 @@ class ProcessingPipelineConfig(BasePipelineConfig):
...
+        ),
+    )
+
+    @model_validator(mode="after")
+    def validate_after(self):
+        """
+        Validate the pipeline config after instantiation.
+
+        Specifically:
+        - Make sure that BIDS_PATH_INJECTION_MAP entries have valid keys and do not
```
mathdugre: We should clarify what is a valid key

nipoppy/config/pipeline.py
```diff
@@ -187,6 +188,42 @@ class ProcessingPipelineConfig(BasePipelineConfig):
...
+          use reserved keywords
+        """
+        super().validate_after()
+        for key, kwargs in self.BIDS_PATH_INJECTION_MAP.items():
+            if not key.isidentifier():
+                raise ConfigError(
+                    f'Invalid key "{key}" in BIDS_PATH_INJECTION_MAP: '
+                    "keys must be valid Python identifiers."
```
mathdugre: The meaning of a Python identifier will be unclear to most people. We should indicate was is a valid key instead.
(Related to above comment)

nipoppy/config/pipeline.py
```diff
@@ -187,6 +188,42 @@ class ProcessingPipelineConfig(BasePipelineConfig):
...
+
+    BIDS_PATH_INJECTION_MAP: dict[str, dict] = Field(
+        default={},
+        description=(
+            "Filters for selecting paths to individual BIDS files to inject into the invocation file."  # noqa: E501
+            " This should be a dictionary where keys correspond to placeholders in the invocation file and values are keyword arguments to be passed to PyBIDS' bids.layout.BIDSLayout.get()."  # noqa: E501
+            f' For example, if the invocation file has a placeholder [[NIPOPPY_{BIDS_PATH_INJECTION_PREFIX}KEY1]] and the BIDS_PATH_INJECTION_MAP has entry {{"KEY1": {{"extension": "nii.gz", "suffix": "T1w"}}}},'  # noqa: E501
+            f" then [[NIPOPPY_{BIDS_PATH_INJECTION_PREFIX}KEY1]] will be replaced with the path to a T1w NIfTI file (provided there is only one match)."  # noqa: E501
+            " This can be useful for pipelines that are not BIDS Apps and that take as input specific files rather than the root directory of the BIDS dataset."  # noqa: E501
+            " Note: this map is only used for pipeline steps that have GENERATE_PYBIDS_DATABASE set to true."  # noqa: E501
```
mathdugre: We should break the paragraph over multiple line instead of using `# noqa: E501`. Otherwise, it makes it hard to read in the code and on GitHub diff.

nipoppy/workflows/processing_runner.py
```diff
@@ -1,21 +1,43 @@
...
+    bids_paths = {}
+    for key, kwargs in injection_map.items():
+        bids_path = bids_layout.get(**kwargs, return_type="filename")
+        if len(bids_path) != 1:
+            raise WorkflowError(
+                f"Expected exactly one file to match criteria {kwargs}, got: {bids_path}"  # noqa: E501
+            )
+
+        bids_path = bids_path[0]
+        bids_paths[f"{BIDS_PATH_INJECTION_PREFIX}{key}".lower()] = bids_path
```
mathdugre: It can be confusing to switch the var type from list to str/path. Instead we can directly assign the path to the `bids_paths` dictionary.
```suggestion
        bids_paths[f"{BIDS_PATH_INJECTION_PREFIX}{key}".lower()] = bids_path[0]
```

tests/unit/config/test_pipeline.py
```diff
@@ -186,6 +191,30 @@ def test_error_schema_version():
...
+            {"KEY1": {"return_value": "test"}},
+            'Invalid entry for key "KEY1" in BIDS_PATH_INJECTION_MAP: "return_value" is a reserved keyword argument and cannot be set externally.',  # noqa: E501
+        ),
+        (
+            {"123": {"extension": "nii.gz", "suffix": "T1w"}},
+            'Invalid key "123" in BIDS_PATH_INJECTION_MAP: keys must be valid Python identifiers.',  # noqa: E501
+        ),
+    ],
+)
+def test_errors_bids_path_injection_map_reserved_keyword(
```
mathdugre: The second parametrization is not about reserved keywords. `invalid` might make more sense.

```suggestion
def test_errors_bids_path_injection_map_invalid_keyword(
```

tests/unit/workflows/test_processing_runner.py
```diff
@@ -128,6 +132,63 @@ def runner(tmp_path: Path, mocker: pytest_mock.MockFixture) -> ProcessingRunner:
...
+
+    result = _get_bids_paths_to_inject(
+        bids_layout=mocked_layout,
+        injection_map=injection_map,
+    )
+    assert result == expected
+
+
+@pytest.mark.parametrize("returned_filenames", [[], ["file1.nii.gz", "file2.nii.gz"]])
+def test_get_bids_paths_to_inject_error(
```
mathdugre: It is unclear what the test is doing from the name. Maybe `test_get_bids_paths_errors_on_non_unique_filename` or alike?

tests/unit/workflows/test_processing_runner.py
```diff
@@ -702,6 +763,44 @@ def test_run_single_pybids_db(
...
+    mocked_get_bids_paths_to_inject = mocker.patch(
+        "nipoppy.workflows.processing_runner._get_bids_paths_to_inject",
+        return_value={"key1": "path1", "key2": "path2"},
+    )
+
+    # mock the Boutiques run outcome
+    mocked_launch_boutiques_run = mocker.patch.object(runner, "launch_boutiques_run")
+
+    participant_id = "01"
+    session_id = "1"
```
mathdugre: These could be passed inline to the function, since they are not reused afterwards.

tests/unit/workflows/test_processing_runner.py
```diff
@@ -702,6 +763,44 @@ def test_run_single_pybids_db(
...
+            bids_layout=mocked_bids_layout,
+            injection_map=runner.pipeline_config.BIDS_PATH_INJECTION_MAP,
+        )
+
+        launch_boutiques_run_kwargs = mocked_launch_boutiques_run.call_args[1]
+        for key, value in mocked_get_bids_paths_to_inject.return_value.items():
+            assert key in launch_boutiques_run_kwargs
+            assert launch_boutiques_run_kwargs[key] == value
+    else:
+        mocked_get_bids_paths_to_inject.assert_not_called()
```
mathdugre: imo, it would be better to have two tests without a condition statement

mathdugre (changes requested):
A couple of things to clarify and some nitpicks, but should be ready to merge soon.

nipoppy/config/pipeline.py
michellewang: I think for the docstring we can be more technical and just say that it needs to be a valid Python identifier. For the warning I agree we should be more descriptive. What do you think?

tests/unit/workflows/test_processing_runner.py
michellewang: How's `test_get_bids_paths_to_inject_no_single_match`?

michellewang:
@mathdugre should we include this in the next release or wait until the one after?

tests/unit/workflows/test_processing_runner.py
```diff
@@ -702,6 +763,58 @@ def test_run_single_pybids_db(
         mocked_set_up_bids_db.assert_not_called()
 
 
+def test_run_single_bids_path_injection(
```
mathdugre:
```suggestion
def test_run_single_bids_path_injection_with_pybids_database(
```

mathdugre (changes requested):
🍒 nitpicks

tests/unit/workflows/test_processing_runner.py
mathdugre: I think we should still specify that we expect an error in the test name

mathdugre:
> @mathdugre should we include this in the next release or wait until the one after?

I would include it in this release

mathdugre (approved):
LGTM 🧑‍🍳

PR #1032, opened 2026-06-05
[ENH] Add schema version field in more config files

### LINKED ISSUE #981: [ENH] Improve config schema awareness/handling

mathdugre:
- Closes #981

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1032.org.readthedocs.build/en/1032/


nipoppy/config/schema.py
```diff
@@ -0,0 +1,32 @@
...
+SCHEMA_VERSION_FIELD = "SCHEMA_VERSION"
+DEFAULT_SCHEMA_VERSION = "1.0.0"
+
+
+def _version_parts(version: str) -> tuple[int, ...]:
+    """Convert a schema version string to comparable integer parts."""
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError as exception:
+        raise ConfigError(f"Invalid schema version: {version}") from exception
```
mathdugre: We could check for a built-in function to do this

michellewang: https://packaging.pypa.io/en/stable/version.html

nipoppy/env.py
```diff
@@ -10,13 +10,15 @@
...
 ZENODO_COMMUNITY_ID = "1c136bd0-655e-495f-8460-884751d4fdf4"
 
-BUG_REPORT_URL = (
-    "https://github.com/nipoppy/nipoppy/issues/new/choose?template=1-bug.yml"
-)
-DISCORD_URL = "https://discord.gg/2VMKFRpjkm"
 
-# pipeline config schema version
-CURRENT_SCHEMA_VERSION = "1"
+class CURRENT_SCHEMA_VERSION(str, Enum):
```
michellewang: Should this be plural instead?

nipoppy/config/schema.py
```diff
@@ -0,0 +1,32 @@
+"""Helpers for configuration schema version checks."""
+
+from __future__ import annotations
+
+from nipoppy.env import CURRENT_SCHEMA_VERSION
+from nipoppy.exceptions import ConfigError
+
+SCHEMA_VERSION_FIELD = "SCHEMA_VERSION"
+DEFAULT_SCHEMA_VERSION = "1.0.0"
```
michellewang: I don't think it makes sense to have a `DEFAULT_SCHEMA_VERSION` since the versions might not be in sync for different config files. And we should really just be using `CURRENT_SCHEMA_VERSION.<>` instead for the defaults, no?

nipoppy/config/schema.py
```diff
@@ -0,0 +1,32 @@
...
+        return tuple(int(part) for part in version.split("."))
+    except ValueError as exception:
+        raise ConfigError(f"Invalid schema version: {version}") from exception
+
+
+def check_schema_version(
+    schema_version: str,
+    current_version: CURRENT_SCHEMA_VERSION,
+) -> None:
+    """Validate SCHEMA_VERSION after model validation."""
```
michellewang: This docstring is not good because this function is independent of Pydantic model validation or field name.

nipoppy/config/schema.py
```diff
@@ -0,0 +1,32 @@
...
+    """Convert a schema version string to comparable integer parts."""
+    try:
+        return tuple(int(part) for part in version.split("."))
+    except ValueError as exception:
+        raise ConfigError(f"Invalid schema version: {version}") from exception
+
+
+def check_schema_version(
+    schema_version: str,
+    current_version: CURRENT_SCHEMA_VERSION,
```
michellewang: Type annotation would be helpful here

michellewang: The default is not valid because `current_version.name` would fail

michellewang: This should be `config_type` instead of `current_version`

nipoppy/env.py
michellewang: I don't think we should store the versions themselves in an `Enum`

I would favour having an enum for the different types of versioned config files, and a mapping between those and their current version.

```python
class CONFIG_TYPE(Enum):
    STUDY = ...
    PIPELINE = ...
    TRACKER = ...
    LAYOUT = ...

CURRENT_SCHEMA_VERSIONS: dict[CONFIG_TYPE, str]: {
    CONFIG_TYPE.STUDY: "1.0.0",
    CONFIG_TYPE.PIPELINE: "1.0.0",
    CONFIG_TYPE.TRACKER: "1.0.0",
    CONFIG_TYPE.LAYOUT: "1.0.0",
}
```

Do you think this would be better?

nipoppy/config/schema.py
```diff
@@ -0,0 +1,32 @@
...
+) -> None:
+    """Validate SCHEMA_VERSION after model validation."""
+    if _version_parts(schema_version) > _version_parts(
+        current_version.value
+    ):  # noqa: E501
+        raise ConfigError(
+            f"{current_version.name} config uses schema version {schema_version}, which"
```
michellewang:
```suggestion
            f"{current_version.name.capitalize()} config uses schema version {schema_version}, which"
```

nipoppy/config/schema.py
michellewang:
```
> from packaging.version import Version
> Version("1") == Version("1.0.0")
True
```

michellewang: Not built-in though so we would need an "extra" dependency (most likely we already have it)

nipoppy/data/examples/sample_global_config.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/layouts/layout-0.1.0.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/layouts/layout-0.2.x.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/layouts/layout-0.3.x.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/layouts/layout-bids-study.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/layouts/layout-default.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/data/template_pipeline/tracker.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

docs/source/how_to_guides/tracking/mriqc-23.1.0-tracker_config.json
```diff
@@ -1,4 +1,5 @@
 {
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

docs/source/how_to_guides/user_guide/inserts/mriqc-23.1.0-tracker_config.json
```diff
@@ -0,0 +1,7 @@
+{
+    "SCHEMA_VERSION": "1.0.0",
```
michellewang: I think we can put that at the end

nipoppy/config/main.py
```diff
@@ -112,6 +113,13 @@ def validate_after(self):
 class Config(_SchemaWithContainerConfig):
     """Schema for dataset configuration."""
 
+    SCHEMA_VERSION: str = Field(
+        default=DEFAULT_SCHEMA_VERSION,
```
michellewang: Default should be based on `CURRENT_SCHEMA_VERSION` (though see other comment about that)

nipoppy/config/pipeline.py
```diff
@@ -74,9 +75,10 @@ class BasePipelineConfig(_SchemaWithContainerConfig, ABC):
     )
     PIPELINE_TYPE: Optional[PipelineTypeEnum] = None
     SCHEMA_VERSION: str = Field(
+        default=DEFAULT_SCHEMA_VERSION,
```
michellewang: Default should be based on `CURRENT_SCHEMA_VERSION`

nipoppy/config/tracker.py
```diff
@@ -7,12 +7,21 @@
...
+from nipoppy.config.schema import DEFAULT_SCHEMA_VERSION, check_schema_version
+from nipoppy.env import CURRENT_SCHEMA_VERSION
 from nipoppy.exceptions import ConfigError
 
 
 class TrackerConfig(BaseModel):
     """Schema for tracker configuration."""
 
+    SCHEMA_VERSION: str = Field(
+        default=DEFAULT_SCHEMA_VERSION,
```
michellewang: Default should be based on `CURRENT_SCHEMA_VERSION`

nipoppy/layout.py
```diff
@@ -53,6 +59,13 @@ class LayoutConfig(BaseModel):
     """Relative paths for the dataset layout."""
 
     model_config = ConfigDict(extra="forbid")
+    SCHEMA_VERSION: str = Field(
+        default=DEFAULT_SCHEMA_VERSION,
```
michellewang: Default should be based on `CURRENT_SCHEMA_VERSION`

nipoppy/layout.py
```diff
@@ -200,6 +222,7 @@ def __init__(
 
         # load the config
         config = LayoutConfig(**load_json(fpath_config))
+        self.SCHEMA_VERSION = config.SCHEMA_VERSION
```
michellewang: Why is this needed?

nipoppy/pipeline_validation.py
```diff
@@ -34,17 +35,56 @@
 class PipelineValidationError(LayoutError): ...  # noqa E701
 
 
+def check_schema_version_exist(
```
michellewang: I think this check should be done (with `strict=False`) when loading each of the config types (as a `@model_validator(mode="before")` to encourage users to set the schema version everywhere

We would lose the path in the error message but I think that's okay as the validation flow already prints the name of each file being checked

tests/integration/test_zenodo_api.py
```diff
@@ -157,6 +158,7 @@ def test_create_new_record_invalid_token(zenodo_api: ZenodoAPI, metadata: dict):
         )
 
 
+# Keep schema_version:1 since there're no uploaded records with the new version yet.
```
michellewang: Given that this is just a test of the search keyword capability, it doesn't matter what the version is. So even when there are new records with the new schema version, we wouldn't necessarily need to change it

tests/unit/config/test_pipeline.py
```diff
@@ -174,10 +181,10 @@ def test_error_pipeline_type(valid_data, extra_data, pipeline_class, valid):
         pipeline_class(**data)
 
 
-def test_error_schema_version():
+def test_error_invalid_schema_version():
```
michellewang: Can probably be removed after the switch to `packaging.version.Version`

tests/unit/config/test_tracker.py
```diff
@@ -29,6 +33,16 @@ def test_no_extra_field():
...
 
 
+def test_schema_version_default():
+    tracker_config = TrackerConfig(PATHS=["path1"])
+    assert tracker_config.SCHEMA_VERSION == DEFAULT_SCHEMA_VERSION
+
+
+def test_schema_version_newer():
+    with pytest.raises(ValueError, match="newer than the schema version"):
+        TrackerConfig(PATHS=[Path("path1")], SCHEMA_VERSION="999.0.0")
```
michellewang: Previous test was using `sys.maxsize`

michellewang: Don't need to wrap with `Path`

tests/unit/test_layout.py
```diff
@@ -38,6 +39,18 @@ def test_config_path_infos():
...
 
 
+def test_schema_version_default():
+    layout = DatasetLayout("my_dataset")
+    assert layout.config.SCHEMA_VERSION == DEFAULT_SCHEMA_VERSION
+
+
+def test_schema_version_newer():
+    data = DatasetLayout("my_dataset").config.model_dump()
+    data["SCHEMA_VERSION"] = "999.0.0"
```
michellewang: The other test uses `sys.maxsize`

michellewang (changes requested):
I think we will need to have different versions for each of the three pipeline types, as we might add a field to one but not all of them.

`nipoppy.config.schema.py` needs a bit of rethinking. I think the current implementation is not the cleanest because enums should be used for choices, not mappings, in my opinion.

We should probably add the warning about missing `SCHEMA_VERSION` fields everywhere, not just in the pipeline validation function.

nipoppy/config/schema.py
mathdugre: Updated in b82113a97b97909f5f1054066fbc57fde5a2886c .
Let me know if you have feedback on the new one.

nipoppy/env.py
mathdugre: I thinks it should be singular. Otherwise, it could imply that each key have multiple current versions, which would be illogical (to me).

nipoppy/config/schema.py
mathdugre: This inferred the schema version when not present in files. Going onward, this should only happen for the existing config file (prior to versioning). Therefore, it represent the earliest version of schema that Nipoppy would be aware of.
I will change the name to `EARLIEST_SCHEMA_VERSION` to clarify.

nipoppy/config/schema.py
mathdugre: I'm not sure to understand this comment. Could you elaborate?

michellewang: It was my mistake, I mistook `CURRENT_SCHEMA_VERSION` as a default value instead of type annotation

nipoppy/env.py
mathdugre: Updated in be88bfdc58cf404f7fadd4c21e7a4281f375ed24

nipoppy/pipeline_validation.py
mathdugre: Can you explain the benefits of doing the validation "before". I'm not as familiar with this part of Pydantic

tests/integration/test_zenodo_api.py
mathdugre: Calrified in c9b4c7511ab40243099568972f8349f5b99898a2

tests/unit/config/test_tracker.py
mathdugre: We should, otherwise there's a mismatch in type annotations.

tests/unit/config/test_pipeline.py
mathdugre: I think we should leave it even after we change to the `packaging.version.Version`. The test shouldn' rely on the internal logic.

mathdugre:
We can discuss more the order for the `SCHEMA_VERSION` in the config file, but the rest is ready for review.

nipoppy/env.py
```diff
@@ -15,8 +15,28 @@
 )
 DISCORD_URL = "https://discord.gg/2VMKFRpjkm"
 
-# pipeline config schema version
-CURRENT_SCHEMA_VERSION = "1"
+BUG_REPORT_URL = (
+    "https://github.com/nipoppy/nipoppy/issues/new/choose?template=1-bug.yml"
+)
+DISCORD_URL = "https://discord.gg/2VMKFRpjkm"
```
michellewang: Duplicated lines

```suggestion
```

nipoppy/env.py
```diff
@@ -15,8 +15,28 @@
...
+    PIPELINE = "pipeline"
+    TRACKER = "tracker"
+    LAYOUT = "layout"
+
+
+SCHEMA_VERSION_INFO = {
+    ConfigType.STUDY: {"current": "1.0.0"},
+    ConfigType.PIPELINE: {"current": "1.0.0"},
+    ConfigType.TRACKER: {"current": "1.0.0"},
+    ConfigType.LAYOUT: {"current": "1.0.0"},
```
michellewang: Unless we can think of a scenario where we would need a patch version, I would keep only major and minor versions

```suggestion
    ConfigType.STUDY: {"current": "1.0"},
    ConfigType.PIPELINE: {"current": "1.0"},
    ConfigType.TRACKER: {"current": "1.0"},
    ConfigType.LAYOUT: {"current": "1.0"},
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,41 @@
...
+    ConfigError
+        If schema version is newer than the one supported by this version of Nipoppy.
+    """
+    current_version = get_current_schema_version(config_type)
+    try:
+        is_newer = Version(schema_version) > Version(current_version)
+    except InvalidVersion as exception:
+        raise ConfigError(f"Invalid schema version: {schema_version}") from exception
+
+    if is_newer:  # noqa: E501
```
michellewang: No need for `# noqa`

```suggestion
    if is_newer:
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,41 @@
...
+        raise ConfigError(
+            f"{config_type.value.capitalize()} config uses schema version "
+            f"{schema_version}, which is newer than the schema version supported by "
+            f"this version of Nipoppy ({current_version}). Please upgrade Nipoppy."
```
michellewang:
```suggestion
            f"{config_type.value.capitalize()} config uses schema version "
            f"{schema_version}, which is newer than the latest schema version supported"
            f" by this version of Nipoppy ({current_version}). Please upgrade Nipoppy."
```

nipoppy/layout.py
```diff
@@ -200,6 +226,7 @@ def __init__(
...
```
michellewang: Still don't see where this is used

```suggestion
```

nipoppy/pipeline_validation.py
michellewang: If "before", it allows us to modify the input dict before instantiation of the Pydantic model. Which could make sense conceptually if we make the schema version required but omit the default. Then for now, if `SCHEMA_VERSION` doesn't exist in the user data, we fix it in the "before" validation. But once we remove that it will fail

michellewang: I do still think we should have a warning if `SCHEMA_VERSION` doesn't exist explicitly in all four Pydantic models

docs/source/how_to_guides/docker/fmriprep_config.json
```diff
@@ -33,5 +33,5 @@
         "TEMPLATEFLOW_HOME": "Path to the directory where TemplateFlow will store templates (can be empty)"
     },
     "PIPELINE_TYPE": "processing",
-    "SCHEMA_VERSION": "1"
+    "SCHEMA_VERSION": "1.0.0"
```
michellewang:
```suggestion
    "SCHEMA_VERSION": "1.0"
```

docs/source/how_to_guides/pipeline_create/data/config.json
```diff
@@ -19,5 +19,5 @@
         }
     ],
     "PIPELINE_TYPE": "processing",
-    "SCHEMA_VERSION": "1"
+    "SCHEMA_VERSION": "1.0.0"
```
michellewang:
```suggestion
    "SCHEMA_VERSION": "1.0"
```

docs/source/how_to_guides/tracking/mriqc-23.1.0-tracker_config.json
michellewang:
```suggestion
    "SCHEMA_VERSION": "1.0",
```

michellewang: For consistency we should have all of them at the top of all of them at the bottom

docs/source/how_to_guides/tracking/mriqc_config.json
```diff
@@ -26,5 +26,5 @@
         "TEMPLATEFLOW_HOME": "Path to the directory where TemplateFlow will store templates (can be empty)"
     },
     "PIPELINE_TYPE": "processing",
-    "SCHEMA_VERSION": "1"
+    "SCHEMA_VERSION": "1.0.0"
```
michellewang:
```suggestion
    "SCHEMA_VERSION": "1.0"
```

tests/integration/test_zenodo_api.py
```diff
@@ -157,6 +158,8 @@ def test_create_new_record_invalid_token(zenodo_api: ZenodoAPI, metadata: dict):
         )
 
 
+# Ensure that records with the specified schema_version exist on Zenodo before changing
+#  the test parameters.
```
michellewang: Not sure what this comment is for

tests/unit/config/test_pipeline.py
michellewang: Then should we test the same thing for all config types?

tests/unit/config/test_pipeline.py
```diff
@@ -186,6 +196,21 @@ def test_error_schema_version():
...
+        ValidationError,
+        match=(
+            ".* config uses schema version.*which is newer than the schema version "
+            "supported by this version of Nipoppy.*Please upgrade Nipoppy."
+        ),
+    ):
+        BasePipelineConfig(
+            NAME="my_pipeline",
+            VERSION="1.0.0",
+            SCHEMA_VERSION=str(sys.maxsize),  # large int for current machine
```
michellewang: I would prefer consistency in the max schema versions (for `Config` the test uses `999.0.0`)

tests/unit/test_layout.py
```diff
@@ -38,6 +40,18 @@ def test_config_path_infos():
...
+def test_schema_version_default():
+    layout = DatasetLayout("my_dataset")
+    assert layout.config.SCHEMA_VERSION == EARLIEST_SCHEMA_VERSION
+
+
+def test_schema_version_newer():
+    data = DatasetLayout("my_dataset").config.model_dump()
+    data["SCHEMA_VERSION"] = str(sys.maxsize)
+    with pytest.raises(ValidationError, match="newer than the schema version"):
+        LayoutConfig(**data)
```
michellewang: These should not use `DatasetLayout`, just `LayoutConfig`.

Can use `from nipoppy.utils.utils import FPATH_DEFAULT_LAYOUT` to get the default layout data in a fixture

michellewang (changes requested):
Looks good, last few things to do:

1. Change `"1.0.0"` to `"1.0"` everywhere
2. Add validation so that warning is emitted if `SCHEMA_VERSION` doesn't exist when loading any of the config files
3. Update tests for layout config to use `LayoutConfig` instead of `DatasetLayout`

tests/integration/test_zenodo_api.py
michellewang: I still got confused about this...

nipoppy/config/pipeline.py
```diff
@@ -105,6 +129,7 @@ def validate_after(self):
         Validate the pipeline configuration after creation.
 
         Specifically:
+        - Check schema version compatibility.
```
michellewang: This should be removed since the validation is no longer happening in this method.

```suggestion
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def get_earliest_schema_version(
+    config_type: ConfigType,
+) -> str:
+    """Get default factory for schema version field in config objects.
+
+    This is used to warn users when they are using a config file that does not include a
+     schema version field.
+    """
```
michellewang: I find the docstring not very clear.

```suggestion
    """Get the earliest schema version for a config type.

    If set as the default_factory in a Pydantic model, this will warn users when 
    they are using a config file that does not include a schema version field.
    """
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,102 @@
...
+        f"{SCHEMA_VERSION_FIELD} field with an explicit version, but it is"
+        f" missing. Latest schema version is {latest_version}."
+    )
+    logger.warning(
+        "Setting schema version to the earliest known version"
+        f" ({EARLIEST_SCHEMA_VERSION}). This will raise an error in a future"
+        " version of Nipoppy. To fix this warning, please add the following field"
+        " to your config file:\n"
+        f'"{SCHEMA_VERSION_FIELD}": "{EARLIEST_SCHEMA_VERSION}"'
+    )
```
michellewang: Do you feel the two warnings are kind of redundant? Maybe only the second one is enough?

nipoppy/config/schema.py
```diff
@@ -0,0 +1,102 @@
...
+        else:
+            current_version = get_current_schema_version(config_type)
+            logger.warning(
+                f"Pipeline configuration file {fpath_config} is missing "
+                f"{SCHEMA_VERSION_FIELD} field. Assuming version "
+                f"{current_version}, but this will raise an error in "
+                "a future version of Nipoppy. To fix this warning, please add the "
+                "following field to your pipeline configuration file:\n"
+                f'"{SCHEMA_VERSION_FIELD}": "{current_version}"'
+            )
```
michellewang, line 74: This should be deleted as it is already being emitted by `get_earliest_schema_version`

nipoppy/config/tracker.py
```diff
@@ -35,6 +66,7 @@ def validate_after(self):
         Validate the configuration after instantiation.
 
         Specifically:
+        - Check schema version compatibility.
```
michellewang: Remove?

```suggestion
```

tests/integration/test_zenodo_api.py
```diff
@@ -7,7 +7,9 @@
 from nipoppy.zenodo_api import ZenodoAPI, ZenodoAPIError
 
 from ..conftest import datetime_fixture  # noqa F401
-from ..conftest import TEST_PIPELINE
+from ..conftest import (
+    TEST_PIPELINE,
+)
```
michellewang, line 12: This can be one line?

tests/unit/config/test_main.py
```diff
@@ -78,6 +83,24 @@ def test_no_extra_fields(valid_config_data):
         Config(**valid_config_data, NOT_A_FIELD="x")
 
 
+def test_schema_version_default(valid_config_data):
+    config = Config(**valid_config_data)
+    assert config.SCHEMA_VERSION == EARLIEST_SCHEMA_VERSION
```
michellewang: We could add a check for the warning(s)

nipoppy/config/schema.py
michellewang: Hm, though it is nice to have the path to the config file printed like that. Makes it easier to edit. But still having what is essentially the same warning twice is not good:

```
[2026-07-10 14:15:08] WARNING [yellow]Tracker config must include SCHEMA_VERSION field with an explicit version, but it is missing. Latest schema version is 1.0.[/]
[2026-07-10 14:15:08] WARNING [yellow]Setting schema version to the earliest known version (1.0). This will raise an error in a future version of Nipoppy. To fix this warning, please add the following field to your config file:
"SCHEMA_VERSION": "1.0"[/]
[2026-07-10 14:15:08] WARNING [yellow]Pipeline configuration file /data/origami/michelle/scratch/test_schema_version/pipelines/16876793/tracker.json is missing SCHEMA_VERSION field. Assuming version 1.0, but this will raise an error in a future version of Nipoppy. To fix this warning, please add the following field to your pipeline configuration file:
"SCHEMA_VERSION": "1.0"[/]
```

nipoppy/config/schema.py
michellewang: In any case, the first warning should not say "must include" since it's not (yet) an error if `SCHEMA_VERSION` is not set

tests/unit/config/test_pipeline.py
```diff
@@ -95,6 +99,11 @@ def test_fields_missing_required(model_class, data):
         model_class(**data)
 
 
+def test_pipeline_schema_version_default():
+    config = BasePipelineConfig(NAME="my_pipeline", VERSION="1.2.3")
```
michellewang, line 108: We could test the warning(s) as well here

tests/unit/config/test_main.py
```diff
@@ -78,6 +83,24 @@ def test_no_extra_fields(valid_config_data):
...
+    with pytest.raises(
+        ValidationError,
+        match="Invalid schema version:",
+    ):
+        Config(**valid_config_data, SCHEMA_VERSION="invalid_version")
+
+
+def test_schema_version_newer(valid_config_data):
+    with pytest.raises(ValidationError, match="newer than the latest schema version"):
+        Config(**valid_config_data, SCHEMA_VERSION="999.0.0")
```
michellewang: This could be merged into a single parametrized test function

tests/unit/config/test_pipeline.py
```diff
@@ -186,6 +195,21 @@ def test_error_schema_version():
         )
 
 
+def test_error_outdated_schema_version():
+    with pytest.raises(
+        ValidationError,
+        match=(
+            ".* config uses schema version.*which is newer than the latest schema "
+            "version supported by this version of Nipoppy.*Please upgrade Nipoppy."
```
michellewang: Merge?

tests/unit/config/test_tracker.py
```diff
@@ -29,6 +34,24 @@ def test_no_extra_field():
...
+    with pytest.raises(
+        ValidationError,
+        match="Invalid schema version:",
+    ):
+        TrackerConfig(PATHS=["path1"], SCHEMA_VERSION="invalid_version")
+
+
+def test_schema_version_newer():
+    with pytest.raises(ValueError, match="newer than the latest schema version"):
+        TrackerConfig(PATHS=[Path("path1")], SCHEMA_VERSION="999.0.0")
```
michellewang: Merge?

tests/unit/config/test_tracker.py
```diff
@@ -29,6 +34,24 @@ def test_no_extra_field():
         TrackerConfig(not_a_field="a")
 
 
+def test_schema_version_default():
+    tracker_config = TrackerConfig(PATHS=["path1"])
```
michellewang: We could test the warning(s) as well here

tests/unit/config/test_tracker.py
```diff
@@ -1,10 +1,15 @@
...
 
+from pathlib import Path
+
 import pytest
+from pydantic import ValidationError
 
+from nipoppy.config.schema import EARLIEST_SCHEMA_VERSION
 from nipoppy.config.tracker import TrackerConfig
 
 FIELDS_STEP = [
```
michellewang: This should be renamed to `FIELDS_TRACKER` I think

tests/unit/test_layout.py
```diff
@@ -33,9 +34,38 @@ def create_invalid_dataset(dpath_root: Path, paths_to_delete: list[str]):
...
+
+
+def test_config_path_infos(layout_config):
+    assert all(
+        [isinstance(path_info, PathInfo) for path_info in layout_config.path_infos]
+    )
+
+
+def test_schema_version_default(layout_config):
+    assert layout_config.SCHEMA_VERSION == EARLIEST_SCHEMA_VERSION
```
michellewang: We could test the warning(s) as well here

tests/unit/test_layout.py
```diff
@@ -33,9 +34,38 @@ def create_invalid_dataset(dpath_root: Path, paths_to_delete: list[str]):
...
+        LayoutConfig(**config_params)
+
+
+def test_schema_version_newer(layout_config):
+    data = layout_config.model_dump()
+    data["SCHEMA_VERSION"] = "999.0.0"
+    with pytest.raises(
+        ValidationError, match="newer than the latest schema version supported"
+    ):
+        LayoutConfig(**data)
```
michellewang: Merge?

Also `test_error_invalid_schema` should use the `layout_config` fixture, no?

tests/unit/test_layout.py
```diff
@@ -33,9 +34,38 @@ def create_invalid_dataset(dpath_root: Path, paths_to_delete: list[str]):
             shutil.rmtree(path_to_delete, ignore_errors=True)
 
 
-def test_config_path_infos():
-    config = DatasetLayout("my_dataset").config
-    assert all([isinstance(path_info, PathInfo) for path_info in config.path_infos])
+@pytest.fixture
+def layout_config() -> LayoutConfig:
+    return LayoutConfig(**load_json(FPATH_DEFAULT_LAYOUT))
```
michellewang: It might be better to have the fixture to be the dict, and do the `LayoutConfig` instantiation inside each test

tests/unit/config/test_main.py
michellewang: These tests are added for every config type, which is quite repetitive. It would be better to have dedicated tests in `unit/config/test_schema.py`, and then in `test_{main,pipeline,tracker,layout}.py` just check that the validator and default factory are set (if that is possible).

tests/unit/test_pipeline_validation.py
```diff
@@ -53,6 +54,25 @@ def test_load_pipeline_config_file():
...
+def test_load_pipeline_config_file_warns_no_schema_version(
+    valid_config_data: dict, caplog: pytest.LogCaptureFixture
+):
+    fpath_config = DPATH_TEST_DATA / "pipeline_config-no-schema-version.json"
+
+    _load_pipeline_config_file(fpath_config)
+
+    assert any(
+        "is missing SCHEMA_VERSION field" in record.message for record in caplog.records
+    )
```
michellewang: These should be moved to `tests/unit/config/test_schema.py`. And we could have `test_load_pipeline_config_file` mock `ensure_config_file_schema_version_exists` and check that it is called with `strict` being passed

michellewang (changes requested):
There are up to 3 warning messages printed per config file load right now, which is a lot

Some of the tests are quite similar and could be refactored. I think we should have dedicated tests for `get_earliest_schema_version` and `ensure_config_file_schema_version_exists` in a new file `tests/unit/config/schema.py`. And in the other files we just test that those functions are used when expected

nipoppy/config/schema.py
mathdugre: Modified the log messages

nipoppy/config/schema.py
mathdugre: Modified the log messages

tests/unit/test_layout.py
mathdugre: I think it's fine as is, but no strong opinion

nipoppy/config/schema.py
```diff
@@ -0,0 +1,76 @@
...
+def ensure_schema_support(
+    schema_version: str,
+    config_type: ConfigType,
+) -> str:
+    """Validate schema version is supported by this version of Nipoppy.
+
+    Raises
+    ------
+    ConfigError
+        If schema version is newer than the one supported by this version of Nipoppy.
```
michellewang:
```suggestion
        If schema version is invalid or newer than the one supported by this version of Nipoppy.
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,76 @@
...
+        raise ConfigError(
+            f"{config_type.value.capitalize()} config uses schema version "
+            f"{schema_version}, which is newer than the latest schema version supported"
+            f" by this version of Nipoppy ({current_version}). Please upgrade Nipoppy."
+        )
+
+    return schema_version
+
+
+def ensure_config_file_schema_version_exists(
```
michellewang: What do you think of renaming to `ensure_schema_version_exists`?

nipoppy/config/schema.py
```diff
@@ -0,0 +1,76 @@
...
+    fpath_config: Path, config_type: ConfigType, strict: bool = False
+) -> str:
+    """Check if the current schema version for pipelines exists."""
```
michellewang:
```suggestion
    """Check if the schema version field is set."""
```

nipoppy/config/schema.py
```diff
@@ -0,0 +1,76 @@
...
+                f"Pipeline configuration file {fpath_config} must include "
+                f"{SCHEMA_VERSION_FIELD} field with an explicit version, but it is"
+                " missing"
+            )
+        else:
+            logger.warning(
+                f"{fpath_config} is missing the required {SCHEMA_VERSION_FIELD} field; "
+                f"assuming version {EARLIEST_SCHEMA_VERSION}. This will become an "
+                f"error in a future Nipoppy release. Add the following to the config: "
+                f'"{SCHEMA_VERSION_FIELD}": "{EARLIEST_SCHEMA_VERSION}"'
```
michellewang: Personal preference which you might disagree with...

```suggestion
                f"error in a future Nipoppy release. To silence this warning, add the following to the config: "
                f'"{SCHEMA_VERSION_FIELD}": "{EARLIEST_SCHEMA_VERSION}"'
```

nipoppy/pipeline_validation.py
```diff
@@ -220,6 +227,9 @@ def _check_pipeline_files(
                 )
                 fpath_tracker_config = dpath_bundle / step.TRACKER_CONFIG_FILE
                 _check_tracker_config_file(fpath_tracker_config)
+                ensure_config_file_schema_version_exists(
+                    fpath_tracker_config, ConfigType.TRACKER, strict=strict
+                )
```
michellewang: Should this be moved inside `_check_tracker_config_file`?

tests/unit/workflows/test_runner.py
```diff
@@ -119,6 +119,39 @@ def runner(tmp_path: Path, mocker: pytest_mock.MockFixture) -> ProcessingRunner:
...
 
 
+def test_run_setup_validates_pipeline_bundle(
+    runner: ProcessingRunner, mocker: pytest_mock.MockFixture
+):
+    runner.pipeline_version = None
+    runner.__dict__.pop("dpath_pipeline_bundle", None)
+    mocked_check_pipeline_bundle = mocker.patch(
+        "nipoppy.workflows.runner.check_pipeline_bundle"
+    )
```
michellewang, line 130: An alternative (more readable in my opinion) way would be this:

```suggestion
    from nipoppy.pipeline_validation import check_pipeline_bundle  # somewhere above
    
    runner.pipeline_version = None
    mocked_check_pipeline_bundle = mocker.patch(
        "nipoppy.workflows.runner.check_pipeline_bundle",
        wraps=check_pipeline_bundle,
    )
```

tests/unit/test_layout.py
```diff
@@ -33,9 +34,21 @@ def create_invalid_dataset(dpath_root: Path, paths_to_delete: list[str]):
...
+
+
+def test_config_path_infos(layout_config):
+    assert all(
+        [isinstance(path_info, PathInfo) for path_info in layout_config.path_infos]
```
michellewang:
```suggestion
        isinstance(path_info, PathInfo) for path_info in layout_config.path_infos
```

tests/unit/test_layout.py
michellewang: I think this was mostly because `test_schema_version_default_schema_version` was slightly convoluted (take existing `LayoutConfig`, dump it to dict, delete key, make a new `LayoutConfig`)

michellewang (changes requested):
Minor comments

tests/unit/workflows/test_runner.py
```diff
@@ -122,11 +122,13 @@ def runner(tmp_path: Path, mocker: pytest_mock.MockFixture) -> ProcessingRunner:
 def test_run_setup_validates_pipeline_bundle(
     runner: ProcessingRunner, mocker: pytest_mock.MockFixture
 ):
-    runner.pipeline_version = None
-    runner.__dict__.pop("dpath_pipeline_bundle", None)
-    mocked_check_pipeline_bundle = mocker.patch(
-        "nipoppy.workflows.runner.check_pipeline_bundle"
-    )
+    from nipoppy.pipeline_validation import check_pipeline_bundle  # somewhere above
```
michellewang: This should be moved to the top of the file... hence the inline comment

michellewang (changes requested):
`tests/unit/workflows/test_runner.py` has an import inside a test that should be moved outside

And I still think the fixture in `tests/unit/test_layout.py` should be the data dict and not the `LayoutConfig`

michellewang (approved):
🧑‍🍳 🎉

michellewang (changes requested):
Sorry, approved but then realized there are merge conflicts from #1027 being merged 😅

michellewang (approved):
🧑‍🍳 🎉

PR #1034, opened 2026-06-07
[ENH] Allow overriding of default global config file through `~/.nipoppy/config.json`

### LINKED ISSUE #872: [ENH] Multi-level Nipoppy configuration file

### New feature

# Context
The current `global_config.json` allows the configuration of a study. However, there are currently no streamlined way of reusing the  part of—or the whole—configuration across multiple studies.

**Question:** How do we handle overlapping configurations between dataset?

# Potential solution
Have a multi-tier configuration:
1. System: `/env/.nipoppy/config`
2. User: `${HOME}/.config/nipoppy/config`
3. Project: `${NIPOPPY_PROJECT_ROOT}/.nipoppy/config`

We would load the config file in order, with project config taking precedence over previous levels.

### Unclear documentation

_No response_

### LINKED ISSUE #911: [ENH] Add user-level "global" config file that gets inherited from
Possibly also system-level one at `/etc/nipoppy/.config`

At `nipoppy init`, the system + user level configs get merged (with user taking precedence) and copied into `<NIPOPPY_ROOT>/global_config.json` (in the future, `<NIPOPPY_ROOT>/.nipoppy/config.json`).

At runtime of other `nipoppy` commands, when the config file gets loaded, apply inheritance from system- and user-level configs as well.

michellewang:
- Closes #872
- Closes #911 

## Changes
- Extracted dataset config creation into `InitWorkflow._create_config_file`, with overrideable user/default config paths for testing and internal use.
- Added fixed user-level config discovery at `~/.nipoppy/config.json`.
- Added `--default-config` to `nipoppy init` to force copying the bundled default global config instead of any user-level config
- Added Pydantic validation via `Config.load(...)` before copying the selected global config file.
- Updated obsolete logger summary message about default config/manifest files being copied

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1034.org.readthedocs.build/en/1034/


nipoppy/workflows/dataset_init.py
```diff
@@ -213,6 +211,38 @@ def handle_bids_source(self) -> None:
         else:
             raise ValueError(f"Invalid mode: {self.mode}")
 
+    def _create_config_file(
+        self,
+        fpath_user_config: StrOrPathLike = FPATH_USER_CONFIG,
+        fpath_default_config: StrOrPathLike = FPATH_SAMPLE_CONFIG,
```
mathdugre: Not sure if we want to pass the default as a parameter here. It feels counter-intuitive that we could change the path to the default config.

tests/unit/workflows/test_init.py
```diff
@@ -174,6 +184,65 @@ def test_run_no_success_message_on_error(
...
 
 
+def test_create_config_file_defaults_to_sample_when_user_config_missing(
+    workflow: InitWorkflow, tmp_path: Path, caplog: pytest.LogCaptureFixture
+):
+    workflow._create_config_file(fpath_user_config=tmp_path / "missing_config.json")
+
+    assert_config_matches(workflow.study.layout.fpath_config, FPATH_SAMPLE_CONFIG)
+    assert "Default config file copied" in caplog.text
+    assert "It may need to be edited to match your dataset" in caplog.text
```
mathdugre, line 201: We should raise an error when the config file is missing, instead of using the default config.
It would generally mean that the user provided a wrong config path.

tests/unit/workflows/test_init.py
```diff
@@ -174,6 +184,65 @@ def test_run_no_success_message_on_error(
...
+    assert "Default config file copied" not in caplog.text
+
+
+def test_create_config_file_default_config_ignores_user_config(
+    dpath_root: Path, tmp_path: Path, caplog: pytest.LogCaptureFixture
+):
+    fpath_user_config = tmp_path / "config.json"
+    fpath_user_config.write_text('{"CUSTOM": {"label": "config"}}\n')
+
+    workflow = InitWorkflow(dpath_root=dpath_root, default_config=True)
```
mathdugre: We coudl use the workflow fixture, then override the value `default_config` value.

mathdugre (changes requested):
Minor comments

tests/unit/workflows/test_init.py
michellewang: We don't allow the user to provide their own config path. We always look for it at `~/.nipoppy/config.json`, so we can't error out if it's not found.

The reason why `_create_config_file` takes the path as arg is basically for testing, but I'll change it to mocking to avoid confusion

mathdugre (approved):
🧑‍🍳

PR #1087, opened 2026-08-04
[MAINT][DOC] Pin doc dependencies

### LINKED ISSUE #1071: [DOC] Pin doc dependencies to avoid ReadTheDocs future builds failing

### Documentation

Docs for versions 0.4.4 and 0.4.5 never got built, and we didn't catch that. Now it's too late since some dependencies are no longer compatible with each other.

We should pin the doc dependencies to be "< next major version" to avoid things like that happening in the future

michellewang:
- Closes #1071

## Changes
- Pin doc dependencies to latest versions that work with 3.10
  - **Except for** `myst-parser` because we need a feature from 5.10 which only supports Python >= 3.11
- Enable dependabot version updates for doc dependencies only
- Use Python 3.14 to build docs in CI

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1087.org.readthedocs.build/en/1087/


mathdugre:
We might want to consider have a lock file (https://packaging.python.org/en/latest/specifications/pylock-toml/)
This can be done with `uv export --format pylock.toml`.

Then in the CI, we can install with the lock file before building the docs.

It might be easier then manually keeping up-to-date the dependencies, while ensuring they are working for the builds.

michellewang:
Having a lockfile is not a bad idea, but wouldn’t we still have to manage it e.g. when dropping a Python version?

michellewang:
Also moving to `uv` for project management is maybe something to be discussed further

mathdugre:
> Having a lockfile is not a bad idea, but wouldn’t we still have to manage it e.g. when dropping a Python version?

Yes, but we won't have to manually resolve each of the dependencies. Possibly only a few key ones.

michellewang:
If using the current implementation, if we update the minimum Python version in `pyproject.toml`, Dependabot will be able to do the version bumps. Not sure how a `pylock.toml` file would be handled

michellewang:
Ok, after thinking it a bit more, I think it's better to do the lockfile approach, since the current approach means we're forced to use older versions of packages which is not ideal

michellewang:
@mathdugre I'm looking into this more. Dependabot doesn't support `pylock.toml` yet, but does support `uv`. It's possible to install via `uv` during the Readthedocs build (https://docs.readthedocs.com/platform/stable/config-file/v2.html#uv). Given that, do we want to use `uv.lock` instead? And switch to `pylock.toml` if/when Dependabot supports it (https://github.com/dependabot/dependabot-core/issues/12094).

Personally I think it's nice to be able to rely on Dependabot for version updates. What do you think?

.github/dependabot.yml
```diff
@@ -5,8 +5,26 @@
...
+
+  - package-ecosystem: "pip"
     directory: "/" # Location of package manifests
+    allow:
+      - dependency-name: "furo"
+      - dependency-name: "mdit-py-plugins"
+      - dependency-name: "myst-parser"
+      - dependency-name: "sphinx*"
+      - dependency-name: "Sphinx*"
+    versioning-strategy: increase
```
michellewang: TODO remove or switch to `uv`

pyproject.toml
```diff
@@ -88,7 +86,7 @@ version-file = "nipoppy/_version.py"
 source = "vcs"
 
 [tool.codespell]
-skip = '.git,.github,pyproject.toml'
+skip = '.git,.github,pyproject.toml,pylock.toml,uv.lock'
```
michellewang: TODO update after picking one lockfile format

mathdugre:
> @mathdugre I'm looking into this more. Dependabot doesn't support `pylock.toml` yet, but does support `uv`. It's possible to install via `uv` during the Readthedocs build (https://docs.readthedocs.com/platform/stable/config-file/v2.html#uv). Given that, do we want to use `uv.lock` instead? And switch to `pylock.toml` if/when Dependabot supports it ([dependabot/dependabot-core#12094](https://github.com/dependabot/dependabot-core/issues/12094)).
> 
> Personally I think it's nice to be able to rely on Dependabot for version updates. What do you think?

This seems like a reasonable approach.

michellewang:
It turns out Dependabot updates are limited since it will only resolve to the lowest Python version accepted by `pyproject.toml`. We might want to do a custom CI workflow for updating the lockfile periodically (say every month). In that case we can go with either `uv.lock` or `pylock.toml`. One caveat is that `pip install -r pylock.toml` shows a warning saying that `pylock.toml` support is experimental and may be removed in the future.

pyproject.toml
```diff
@@ -44,8 +44,7 @@ doc = [
     "mdit-py-plugins",
     # Simplify after dropping Python 3.10 support
     "myst-parser>=5.1.0 ; python_version >= '3.11'",
-    "myst-parser ; python_version < '3.11'",
-    "sphinx",
+    "myst-parser ; python_version < '3.11'",    "sphinx",
```
mathdugre: Formatting looks weird here. `"sphinx"` should be on a new line.

mathdugre (changes requested):
Very minor change, otherwise LGTM 🧑‍🍳

mathdugre (approved):

PR #1088, opened 2026-08-05
[MAINT] Reorganize structure of the project files

### LINKED ISSUE #922: [MAINT] File organization and Public API

mathdugre:
- Closes #922

## Checklist

### All PRs
- [x] Tests pass
- [x] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1088.org.readthedocs.build/en/1088/


michellewang (commented):
Looks like the tests and docs are failing :/

mathdugre:
> Looks like the tests and docs are failing :/

My bad, forgot to commit the latest changes before pushing 🤦

docs/source/conf.py
```diff
@@ -199,9 +199,15 @@
...
     ("py:class", "typing_extensions.Self"),
     ("py:class", "httpx.Client"),
+    ("py:class", "NipoppyLogger"),
+]
+
+nitpick_ignore_regex = [
+    ("py:class", r"nipoppy\.core\._.*"),
+    ("py:func", r"nipoppy\.core\._.*"),
+    ("py:class", r"nipoppy\.workflows\.pipeline\._utils\..*"),
 ]
```
michellewang, line 211: Is there a way to avoid the warnings without blanket ignoring entire modules?

nipoppy/core/_models/config/main.py
michellewang, line 1: Should we call this `study.py` instead of `main.py`?

nipoppy/core/_models/config/schema.py
michellewang, line 1: Not sure this file should be in `_models` as it's not defining a model. And because `ensure_schema_version_exists` is used in `pipeline_validation.py` so it's not purely a config thing.

Maybe it could go in `utils/` instead?

nipoppy/core/__init__.py
michellewang, line 1: This feels like it's not saying much... What even is "domain logic" here?

nipoppy/core/layout.py
michellewang, line 1: Did we want to split this file so that the layout model goes under `_models`?

michellewang: I guess if this PR is only file renaming then it would need to be done in a separate PR

nipoppy/data_retriever.py
michellewang, line 1: I think we can keep that one private and just expect users to do `from nipoppy import NipoppyDataRetriever`. This allows more flexibility in the future if we ever want to move the file to a subdirectory.

nipoppy/workflows/_base.py
michellewang, line 1: We probably want to extract the `BaseDatasetWorkflow` into its own file as well (in another PR). The tests are already split across two files.

tests/unit/core/_utils/test_bids.py
michellewang, line 1: Is `tests/unit/core/_utils` missing an `__init__.py` file?

nipoppy/__init__.py
michellewang, line 1: Should we have a `src` directory? I.e. have `src/nipoppy` instead of `nipoppy/`. Is that a better practice?

michellewang (changes requested):
Looks good overall, only a few comments/questions to be resolved. We could have a follow-up PR (or multiple ones) that further splits some files into more files (I made some comments about that).

Docs are failing due to a warning:
```
generating indices... /home/docs/checkouts/readthedocs.org/user_builds/.   nipoppy/checkouts/1088/docs/source/reference/autoapi/nipoppy/core/layout/LayoutConfig.rst:17: WARNING: py:class reference target not found: ConfigType.LAYOUT [ref.class
```

nipoppy/core/layout.py
mathdugre: I would be okay to do it in this PR if it doesn't result in too many changes.

nipoppy/__init__.py
mathdugre: Sure, I would be fine with that.

From this, it seems to be better 🤔 
https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

michellewang:
Closing because we have decided to address this in the future after discussing more in-depth on the whole code architecture/design.
Also there are many merge conflicts already, so we probably need to make these kinds of changes incrementally/module-by-module

PR #945, opened 2026-03-26
[WIP][BLOCKED][ENH] Change the default layout to be BIDS-study-compliant

michellewang:
- Closes #940

Changes proposed in this pull request:
- Make `layout-bids-study.json` the default layout, with 2 changes related to `rawbids` and `phenotype` directories
- Fix tests

Blocked ~by https://github.com/bids-standard/bids-specification/pull/2191~ until the next BIDS release as this currently doesn't pass the BIDS validator. Getting errors:

```
[ERROR] NOT_INCLUDED Files with such naming scheme are not part of BIDS specification. This error is most commonly caused by typos in file names that make them not BIDS compatible. Please consult the specification and make sure your files are named correctly. If this is not a file naming issue (for example when including files not yet covered by the BIDS specification) you should include a ".bidsignore" file in your dataset (see https://github.com/bids-standard/bids-validator#bidsignore for details). Please note that derived (processed) data should be placed in /derivatives folder and source data (such as DICOMS or behavioural logs in proprietary formats) should be placed in the /sourcedata folder.
                /phenotype/
                /phenotype/assessments/

                1 more files with the same issue

        Please visit https://neurostars.org/search?q=NOT_INCLUDED for existing conversations about this issue.

        [ERROR] EXTENSION_MISMATCH Extension used by file does not match allowed extensions for its suffix
                /phenotype/README.md

        Please visit https://neurostars.org/search?q=EXTENSION_MISMATCH for existing conversations about this issue.
```

I'm not sure if READMEs or subdirectories will be allowed in `phenotype`, so we should wait for the BIDS PR to be merged and the BIDS validator to be released before changing any of that.

## Checklist (for reviewers)
_This section is for the PR reviewer_

- [ ] PR has an interpretable title with a prefix (e.g. `[BUG]`, `[DOC]`, `[ENH]`, `[MAINT]`)\
Refer to [NumPy Development Guide](https://numpy.org/doc/stable/dev/development_workflow.html#writing-the-commit-message) for a full list
- [ ] PR links to GitHub issue with mention `Closes #XXXX`
- [ ] Tests pass
- [ ] Checks pass

For new features:
- [ ] Tests have been added

For bug fixes:
- [ ] There is at least one test that would fail under the original bug conditions

----
📚 Documentation preview 📚: https://nipoppy--945.org.readthedocs.build/en/945/

PR #1099, opened 2026-08-12
[ENH] Add comments to default config files and point to substitutions doc

### LINKED ISSUE #884: [ENH] Add TEMPLATING VAR info to the pipeline config files

- [ ] I have searched the existing issues

### New feature

- Related: #883

When we create the default pipeline configuration files with `nipoppy pipeline create` we should include a list of TEMPLATING VARS atop of each file.
This would improve the experience of developing new pipelines.

### Unclear documentation

_No response_

michellewang:
- Closes #884

## Changes
- Add comments to default study config and pipeline config files
   - Includes a top-level comment pointing to https://nipoppy.readthedocs.io/en/latest/reference/substitutions.html, but instead of `latest`, the URL will have the current version of Nipoppy
- Move `copy_template` to `fileops` module
- Update doc workflow to apply the version substitution on the sample study config file. Otherwise, the config shown in the doc will have `[[NIPOPPY_VERSION]]` which does not look good
- Note: `copy_template` will raise "Unable to replace ..." warnings on files that have other substitutions. I silenced this as it would be very annoying to the user. The tests will explicitly check that the expected version substitution has been done
- Removed the pre-commit hook that checked JSON files, as it would fail on JSON5 files. Given that all JSON5 files are expected to be loaded somewhere in the test suite I don't think the hook is critical

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1099.org.readthedocs.build/en/1099/


tests/unit/workflows/pipeline_store/test_create.py
```diff
@@ -127,55 +154,6 @@ def test_create_from_descriptor(workflow: PipelineCreateWorkflow):
     )
 
 
-def test_create_from_descriptor_preserves_json5(
```
michellewang, line 130: Removed as it kind of got folded into `test_create`: if JSON5 was not preserved, `test_create` would fail

.pre-commit-config.yaml
```diff
@@ -12,7 +12,6 @@ repos:
     hooks:
     -   id: check-ast
     -   id: check-case-conflict
-    -   id: check-json
```
mathdugre, line 15: We should rename the file to have a `.json5` extension instead of remove the `json` check.
It will also help IDEs detect the correct file format.

docs/scripts/sample_study_config.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python
+"""Script to automatically generate the sample study config file shown in the docs."""
+
+from pathlib import Path
+
+import nipoppy.utils.fileops
+from nipoppy.env import PROGRAM_VERSION
+from nipoppy.utils.utils import FPATH_SAMPLE_CONFIG
+
+DPATH_INSERTS = Path(__file__).parent / ".." / "source" / "_inserts"
```
mathdugre: We can use `parents[1]` instead of  `.parent / ".."`
```suggestion
DPATH_INSERTS = Path(__file__).parents[1] / "source" / "_inserts"
```

nipoppy/data/examples/sample_global_config.json5
```diff
@@ -16,13 +18,13 @@
...
-        "# - venv:  source <PATH_TO_VENV>/bin/activate",
-        "# - conda: source ~/.bashrc; conda activate <ENV_NAME>",
-        "# ========== Set environment variables ==========",
+        // (These lines can all be removed if not using HPC functionality.)
+        // ========== Activate Python environment ==========
+        // Here we need the command to activate your Python environment in an
+        // HPC job, for example:
+        // - venv:  source <PATH_TO_VENV>/bin/activate
+        // - conda: source ~/.bashrc; conda activate <ENV_NAME>
+        // ========== Set environment variables ==========
```
mathdugre, line 27: 🍒 Could be a multi-line comment block (i.e. `/* comment */`).

Same for the header comment.

nipoppy/data/examples/sample_global_config.json
```diff
@@ -1,6 +1,8 @@
+// String substitutions will be applied when this file is loaded,
```
mathdugre: Could use active voice instead; e.g. "Nipoppy applies string substitutions at runtime."

nipoppy/utils/fileops.py
```diff
@@ -45,6 +46,33 @@ def copy(source: Path, target: Path, dry_run=False, exist_ok: bool = False):
...
+    path_dest
+        Destination file path
+    **template_kwargs
+        Keyword arguments passed to process_template_str for substitution
+    """
+    logger.debug(f"Copying template {path_source} to {path_dest}")
+    if not dry_run:
+        with open(path_source, "r") as f:
+            content = process_template_str(f.read(), **template_kwargs)
+        mkdir(Path(path_dest).parent, dry_run=dry_run)
```
mathdugre: This should be outside the ` if not dry_run:` scope. This will let the `mkdir` function log it's operation in dry_run mode.

mathdugre: We don't have to convert to Path() since the type accepted by the function is already a Path

nipoppy/utils/fileops.py
```diff
@@ -45,6 +46,33 @@ def copy(source: Path, target: Path, dry_run=False, exist_ok: bool = False):
...
+        with open(path_dest, "w") as f:
+            f.write(content)
```
mathdugre: The read-write operation can be simplified:

```py
path_dest.write_text(
    process_template_str(path_source.read_text(),
    **template_kwargs
)
```

nipoppy/utils/fileops.py
```diff
@@ -45,6 +46,33 @@ def copy(source: Path, target: Path, dry_run=False, exist_ok: bool = False):
             shutil.copytree(src=source, dst=target, dirs_exist_ok=exist_ok)
 
 
+def copy_template(
+    path_source: Path,
+    path_dest: Path,
```
mathdugre: I would prefer using only `source` and `dest` to be consistent with the other functions in the module.

nipoppy/utils/fileops.py
```diff
@@ -45,6 +46,33 @@ def copy(source: Path, target: Path, dry_run=False, exist_ok: bool = False):
             shutil.copytree(src=source, dst=target, dirs_exist_ok=exist_ok)
 
 
+def copy_template(
```
mathdugre, line 49: We probably want to check if the dest/target file already exist, similar to the `copy` function above.

We could actually reuse the copy function from above and only do the string substitution in-place afterwards.

nipoppy/workflows/pipeline_store/create.py
```diff
@@ -71,19 +72,30 @@ def create_bundle(
...
+        fileops.copy_template(
+            TEMPLATE_PIPELINE_PATH.joinpath("invocation_header.txt"),
+            fpath_invocation,
+            version=PROGRAM_VERSION,
+            dry_run=self.dry_run,
         )
-        fileops.copy(
+        # then append the actual example invocation
+        with open(fpath_invocation, "a") as file_invocation:
+            file_invocation.write(boutiques.example(str(descriptor_path)))
```
mathdugre: We can use pathlib syntax
```suggestion
        fpath_invocation.open("a").write(boutiques.example(str(descriptor_path)))
```

nipoppy/workflows/pipeline_store/create.py
mathdugre, line 45: Note for opnening issue:
The `type_` arg could be remove in favor of using the `self.type_` from the class. Likewise for `source_descriptor`.

tests/unit/workflows/pipeline_store/test_create.py
```diff
@@ -4,20 +4,29 @@
...
 
-def _has_same_content(a: Path, b: Path) -> bool:
-    """Check if two files are the same."""
-    return a.read_text().strip() == b.read_text().strip()
+def _has_same_JSON_content(a: Path, b: Path) -> bool:
+    """Check if two files have the same JSON content."""
+    return load_json(a, allow_json5=True) == load_json(b, allow_json5=True)
+
+
+def _JSON5_comment_correct(path: Path) -> bool:
```
mathdugre: Should rename to `_has_correct_JSON5_comment` to be consistent with the other helper function above.

mathdugre (changes requested):
Few minor changes, otherwise LGTM

.pre-commit-config.yaml
michellewang: Renaming all of these files to `.json5` in the layout is a big change if we want to accept both (which we definitely should, otherwise this is a breaking layout change). Would need further discussion. Also, VS Code doesn't call their settings file `settings.json5`/`settings.jsonc`, it's `settings.json` (but of course they control the IDE behaviour).

The renaming would be an issue for `global_config.json` and the pipeline `config.json` files. For other pipeline config files it doesn't matter because the paths are set inside `config.json`.

docs/scripts/sample_study_config.py
michellewang: Is this better practice or a personal preference on your side? To me `.parent / ".."` is more explicit

nipoppy/workflows/pipeline_store/create.py
michellewang: I can change to `with path.open("a")` but will keep the 2-line context manager structure since otherwise the file is not closed (as far as I can tell)

docs/scripts/sample_study_config.py
mathdugre: I tend to prefer `parent[1]` or `parent.parent`, but I'm not sure what's the best practice.

nipoppy/workflows/pipeline_store/create.py
mathdugre: Yes, you're right, indeed

nipoppy/data/examples/sample_global_config.json5
michellewang: If multiline, I would format it like this:
```javascript
        /*
         * (These lines can all be removed if not using HPC functionality.)
         * ========== Activate Python environment ==========
         * Here we need the command to activate your Python environment in an
         * HPC job, for example:
         * - venv:  source <PATH_TO_VENV>/bin/activate
         * - conda: source ~/.bashrc; conda activate <ENV_NAME>
         * ========== Set environment variables ==========
         */
```
which is 2 extra lines per comment. Not sure if it's worth it.

Also VS code will use `//` when toggling comments for a multiline selection which is more convenient in my opinion

docs/scripts/sample_study_config.py
michellewang: Changed to `parent.parent`

michellewang:
Dev meeting decisions
- Change global config path to `.json5` in the BIDS-study layout PR #945
- Start (long-ish) deprecation cycle for `config.json` -> `config.json5` #1105

michellewang:
@mathdugre: I renamed all the internal JSON5 files to have the `.json5` extension. `nipoppy pipeline create` now writes `{invocation,hpc,tracker}.json5`. For the `global_config.json` in the layout and the pipelines' `config.json`, we can handle those later as per the comment above.

nipoppy/workflows/pipeline_store/create.py
```diff
@@ -70,19 +81,30 @@ def create_bundle(
...
             boutiques.create(str(descriptor_path))
 
-        target.joinpath("invocation.json").write_text(
-            boutiques.example(str(descriptor_path))
+        invocation_path = target.joinpath(pipeline_step_config.INVOCATION_FILE)
+        # copy the 'header' (top-level comments)
+        fileops.copy_template(
+            TEMPLATE_PIPELINE_PATH.joinpath("invocation_header.txt"),
+            invocation_path,
+            version=PROGRAM_VERSION,
```
mathdugre: It would be easier to distinguish which parameters are the kwargs if they were passed as a dictionary. 

We could have a `substitution` arg in the `copy_template` function instead. Then we would have something like this:

```py
fileops.copy_template(
            TEMPLATE_PIPELINE_PATH.joinpath("invocation_header.txt"),
            invocation_path,
            dry_run=self.dry_run,
            substitutions={
                       "version": PROGRAM_VERSION,
             }
)
```

tests/unit/utils/test_fileops.py
```diff
@@ -215,6 +215,46 @@ def test_substitution(self, tmp_path: Path):
...
+    def test_copy_template_creates_parent_directory(self, tmp_path: Path):
+        """Test that the parent directory of the destination is created."""
+        template_file = tmp_path / "template.txt"
+        template_file.write_text("Hello, [[NIPOPPY_NAME]]!")
+
+        dest_file = tmp_path / "nonexistent_dir" / "output.txt"
+
+        fileops.copy_template(template_file, dest_file, name="World")
+
+        assert dest_file.is_file()
```
mathdugre: The file read right after would fail if `dest_file` is not a file so this check probably not needed

mathdugre (changes requested):
Minor change + 🍒 nitpick

mathdugre (approved):
LGTM

PR #1101, opened 2026-08-14
[MAINT] Simplify the get_pipeline_tag

mathdugre:
## Changes
Refactoring of the `get_pipeline_tag` to remove redundancy in the conditionals. Instead, it uses a filter.

## Checklist

### All PRs
- [x] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1101.org.readthedocs.build/en/1101/


nipoppy/utils/utils.py
```diff
@@ -60,52 +58,45 @@ def get_pipeline_tag(
...
-    return sep.join(components)
-
-
-def load_json(
-    fpath: StrOrPathLike,
-    *,
-    allow_json5: bool = False,
-    **kwargs,
-) -> dict | list:
+    """Join pipeline submission parameters, filtering out None values."""
```
michellewang, line 63: Docstring is incorrect. This function has nothing to do with "submission parameters".

Arguably the "filtering out None values" is not super useful

michellewang (changes requested):

nipoppy/utils/utils.py
mathdugre: We should discuss a better name for these parameters. I don't think pipeline tag is suitable either, since it also combines the participant and session ids which are not related to the pipeline.

michellewang: I'm okay with calling this something other than "pipeline tag", but I feel it should still be some sort of "tag". Maybe `run_tag`?

mathdugre:
from dev meeting: We can keep the current func name and rename later on while doing a larger refactoring

nipoppy/utils/utils.py
michellewang: @mathdugre can we rephrase the docstring to "Generate a tag"?

michellewang (changes requested):

nipoppy/utils/utils.py
mathdugre: I reverted to the previous docstring. We can modify it along with the func name in a later PR, once we agree on the naming of those args.

michellewang (approved):
🧑‍🍳

PR #1104, opened 2026-08-21
[ENH] add --community option to upload workflow

### LINKED ISSUE #642: [ENH] Prevent immediate Zenodo upload if file contents/metadata are exactly the same

### New feature

- We can compare the hashes for the file contents.
- Not sure if metadata checking is necessary -- we need to make sure to **not** check the date since that will probably be different
- Allow user to override with `--force`

### Unclear documentation

_No response_

### LINKED ISSUE #772: [DOC] Add instruction for adding pipeline to Nipoppy community on Zenodo

mathdugre:
- Closes #642 
- Closes #772

## Changes
- Add a `--community` flag to the `nipoppy upload` command, allowing user to request inclusion to the Nipoppy community via the CLI.
- Skip upload when the local metadata and files match the latest record from Zenodo.
Note: The community is always done, even when the pipeline upload is skipped.

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1104.org.readthedocs.build/en/1104/


nipoppy/cli/pipeline_catalog.py
```diff
@@ -194,6 +194,11 @@ def pipeline_validate(**params):
     is_flag=True,
     help="Ignore safeguard warnings and upload anyway. Use with caution.",
 )
+@click.option(
+    "--community",
+    is_flag=True,
+    help="Request inclusion of the pipeline in the Nipoppy Zenodo community.",
+)
```
michellewang, line 201: `--community` is showing after `--sandbox` in the help:

Image

I think it should be before

nipoppy/cli/pipeline_catalog.py
michellewang, line 181: The type for this should not be string. Should be like the other commands: `click.Path(path_type=Path, exists=True, file_okay=False, resolve_path=True)`

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -143,19 +186,29 @@ def run_main(self):
...
         doi = self.zenodo_api.upload_record(
             input_dir=pipeline_dir,
             record_id=self.record_id,
             metadata=metadata,
             default_preview_filename=DatasetLayout.fname_pipeline_config,
+            community_id=community_id,
         )
         logger.success(f"Pipeline successfully uploaded at {doi}")
+        if community_id is not None:
+            logger.success("Nipoppy community inclusion request submitted.")
```
michellewang: Should we have an additional message if `community_id` is `None` to say they can rerun with `--community` and `--zenodo-id <...>` to request inclusion in the Nipoppy community?

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -130,6 +150,29 @@ def run_main(self):
...
+
+        if (
+            latest_record_id is not None
+            and not self.force
+            and self._is_same_record(
+                record_id=latest_record_id,
+                input_dir=pipeline_dir,
+            )
+        ):
+            logger.success(
```
michellewang: I don't think this should be `success`. Maybe warning or info

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -130,6 +150,29 @@ def run_main(self):
...
+                "Pipeline files are unchanged; skipping upload for "
+                f"zenodo.{latest_record_id}."
```
michellewang: Is there a reason why we always have the `zenodo.` prefix? I feel it is unnecessary

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -130,6 +150,29 @@ def run_main(self):
...
+            )
+            if community_id is not None:
+                self.zenodo_api.request_community_inclusion(
+                    latest_record_id, community_id
+                )
+                logger.success("Nipoppy community inclusion request submitted.")
```
michellewang: If the record is already in the Nipoppy community and I still use `--community`, I get `"Nipoppy community inclusion request submitted."` even though that is not true.

Image

Instead we should just have the "... is already in community ..." warning message

docs/source/how_to_guides/pipeline_create/index.md
```diff
@@ -221,6 +221,21 @@ nipoppy pipeline upload \
   <PIPELINE_DIR>
 ```
 
+::::{tip}
```
michellewang, line 227: I would put this box below the `note` box about `--zenodo-id`

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -81,6 +84,21 @@ def _get_pipeline_metadata(
...
         return metadata
 
+    def _is_same_record(self, record_id: str, input_dir: Path) -> bool:
+        """Check whether local files match a published Zenodo record."""
+        local_files = {
+            file.name: _get_file_md5(file) for file in sorted(input_dir.iterdir())
+        }
+        remote_files = {
+            entry["key"]: (
+                entry["checksum"].lower()
```
michellewang: We could have `entry["checksum"].lower().removeprefix("md5:")` instead of the `if`/`else`. It's what we do in `ZenodoAPI` and will make the dict comprehension more readable.

And then we don't need to add `md5:` in `_get_file_md5`

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -130,6 +150,29 @@ def run_main(self):
...
+            and self._is_same_record(
+                record_id=latest_record_id,
+                input_dir=pipeline_dir,
+            )
+        ):
+            logger.success(
+                "Pipeline files are unchanged; skipping upload for "
+                f"zenodo.{latest_record_id}."
+            )
+            if community_id is not None:
```
michellewang: Having two places where the community inclusion can happen is not very clean.

Instead of:
- If files don't need upload
    - warn user
    - do the community request
    - return
- upload files
- do the community request

We could have:
- If files don't need upload:
    - warn user
- else
    - upload files
    - do the community request

Which would avoid duplication. We could remove the `community_id` handling part of `ZenodoAPI.upload_record` as well which is better for maintenance.

nipoppy/zenodo_api.py
```diff
@@ -380,16 +385,76 @@ def _get_api_endpoint(self, community_id: str | None = None) -> str:
...
-                f"Failed to get metadata for zenodo.{record_id}: {response.json()}"
+                f"Failed to get record zenodo.{record_id}: {response.json()}"
+            )
+
+        return response.json()
+
+    def get_record_files(self, record_id: str) -> dict:
+        """Get the file manifest of a published Zenodo record."""
+        record_id = self._process_record_id(record_id)
+        response = self.client.get(f"/records/{record_id}/files")
```
michellewang: I notice that here we don't include the API endpoint itself. Should we remove it from the other calls that use the client? E.g. replace this:

```python
response = self.client.get(
            f"{self.api_endpoint}/records/{record_id}",
        )
```

by this:
```python
response = self.client.get(
            f"/records/{record_id}",
        )
```

nipoppy/zenodo_api.py
```diff
@@ -272,6 +266,7 @@ def upload_record(
         metadata: dict,
         record_id: str | None = None,
         default_preview_filename: str | None = None,
+        community_id: str | None = None,
```
michellewang: See other comment. I don't think we should couple the community request with this

nipoppy/zenodo_api.py
```diff
@@ -380,16 +385,76 @@ def _get_api_endpoint(self, community_id: str | None = None) -> str:
...
+        record_id = self._process_record_id(record_id)
+        response = self.client.get(f"/records/{record_id}/files")
+        if response.status_code != 200:
+            raise ZenodoAPIError(
+                f"Failed to get files for zenodo.{record_id}: {response.json()}"
+            )
+        return response.json()
+
+    def get_community_id(self, community: str) -> str:
+        """Resolve a community slug or ID in the active Zenodo environment."""
```
michellewang: What does "active Zenodo environment" mean?

tests/unit/workflows/pipeline_store/test_upload.py
```diff
@@ -151,6 +154,73 @@ def test_is_same_pipeline(pipeline_config, zenodo_metadata, expected):
...
+    workflow.zenodo_api.get_record_files.return_value = {
+        "default_preview": "different-preview.json",
+        "entries": [
+            {
+                "key": file_to_upload.name,
+                "checksum": (
+                    # Zenodo prepends "md5:" to the checksum of the preview file
+                    _get_file_md5(file_to_upload).removeprefix("md5:")
+                    if index == 0
+                    else _get_file_md5(file_to_upload)
```
michellewang: `str.removeprefix()` does nothing if the prefix doesn't exist, so we don't need the `if`/`else`

michellewang: Actually I'm not sure I understand this logic and how it is consistent with the comment `# Zenodo prepends "md5:" to the checksum of the preview file`. Doesn't the `if`/`else` here strip the prefix for the first file? And none of these files are the preview file which is called `different-preview.json`

tests/unit/workflows/pipeline_store/test_upload.py
```diff
@@ -183,6 +253,103 @@ def test_upload_same_pipeline(
...
+    assert "files are unchanged; skipping upload" in caplog.text
+    if community:
+        workflow.zenodo_api.get_community_id.assert_called_once_with("nipoppy")
+        workflow.zenodo_api.request_community_inclusion.assert_called_once_with(
+            latest_record_id, community_id
+        )
+        assert "Nipoppy community inclusion request submitted" in caplog.text
+    else:
+        workflow.zenodo_api.get_community_id.assert_not_called()
+        workflow.zenodo_api.request_community_inclusion.assert_not_called()
```
michellewang: Would it be very annoying to split the community upload checks into different tests? I think we said we wanted to keep logic minimal in the tests

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -130,6 +150,29 @@ def run_main(self):
                     f"{', '.join(potential_duplicates)}",
                 )
 
+        community_id = (
+            self.zenodo_api.get_community_id("nipoppy") if self.community else None
```
michellewang: We could use `nipoppy.env.ZENODO_COMMUNITY_ID` here. `nipoppy pipeline search` uses that

tests/unit/workflows/pipeline_store/test_upload.py
michellewang: Actually, the community upload should probably not be tested here at all. Related to my other comments about refactoring so that there is only one place where the request is made

tests/unit/workflows/pipeline_store/test_upload.py
```diff
@@ -183,6 +253,103 @@ def test_upload_same_pipeline(
...
+    mocker: pytest_mock.MockerFixture,
+):
+    metadata = {"metadata": {}}
+    community_id = "nipoppy-community-id"
+    mocker.patch.object(workflow, "_get_pipeline_metadata", return_value=metadata)
+    mocker.patch(
+        "nipoppy.workflows.pipeline_store.upload.check_pipeline_bundle",
+    )
+    workflow.assume_yes = True
+    workflow.force = True
```
michellewang: Not strictly necessary?

tests/unit/workflows/pipeline_store/test_upload.py
```diff
@@ -183,6 +253,103 @@ def test_upload_same_pipeline(
...
+    )
+
+    workflow.run_main()
+
+    is_same_record.assert_called_once_with(
+        record_id=latest_record_id,
+        input_dir=TEST_PIPELINE,
+    )
+    workflow.zenodo_api.upload_record.assert_not_called()
+    confirm.assert_not_called()
```
michellewang: Not sure this is needed if we already check that `upload_record` is not called. But fine to keep as-is

tests/unit/test_zenodo_api.py
michellewang, line 1: We should add tests for the new/refactored functions `get_record` and `get_record_files`

.gitignore
```diff
@@ -40,6 +40,7 @@ docs/source/reference/schemas/*.json
 env/
 venv/
 .venv/
+.venv
```
michellewang, line 42: What's the difference between `.venv` and `.venv/`? You have a file called `.venv`?

tests/unit/test_zenodo_api.py
michellewang, line 1: We should have integration tests in `tests/integration/test_zenodo_api.py` for the community request feature

michellewang (changes requested):
The new behaviour is mostly good. I think the implementation in the workflow can be simplified, but let me know if I missed something. Missing integration tests.

nipoppy/cli/pipeline_catalog.py
mathdugre: Can you open a new issue for this? It's unrelated to this PR.

michellewang: I think it's reasonable to address in this PR since it touches the `nipoppy upload` command... Just like the `--community` flag showing up after the `--sandbox` flag is technically due to a previous PR adding `--community` to `nipoppy search`

.gitignore
mathdugre: It's for the `.venv` symlink created by `uv`. I can make the change in a new PR if you prefer

michellewang: No it's fine to leave it here. Does `.venv` cover both file and directory cases? If yes then we could just have `.venv`. If not both is fine but maybe a comment to avoid future confusion

nipoppy/workflows/pipeline_store/upload.py
mathdugre: Changing to warning

nipoppy/zenodo_api.py
mathdugre: Yes, I think we should remove them, but in another PR.

docs/source/how_to_guides/pipeline_create/index.md
mathdugre: Sure. Do you think repeating the command is necessary or only having the paragraph would be okay?

nipoppy/zenodo_api.py
mathdugre: Removed it

nipoppy/workflows/pipeline_store/upload.py
mathdugre: We have to use "nipoppy" here if we want it to work on sandbox as well.

.gitignore
mathdugre: .venv seems to cover both cases. Will remove the previous .venv/

nipoppy/workflows/pipeline_store/upload.py
mathdugre: We put them in the past, but I can remove them going forward

nipoppy/workflows/pipeline_store/upload.py
mathdugre: I removed the `md5:` prefix across the whole zenodo api code.

tests/unit/workflows/pipeline_store/test_upload.py
mathdugre: Removed `md5:` prefix across the code base.

nipoppy/cli/pipeline_catalog.py
mathdugre: Sure, will do it in this PR

nipoppy/zenodo_api.py
mathdugre: Did it in this PR in the end

mathdugre:
Only this point need confirmation, otherwise ready for another review. @michellewang 
https://github.com/nipoppy/nipoppy/pull/1104#discussion_r3935073642

docs/source/how_to_guides/pipeline_create/index.md
michellewang: I think full command is fine, but it's up to you

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -96,28 +141,28 @@ def run_main(self):
             raise WorkflowError from e
 
         if self.record_id:
-            self.record_id = self.record_id.removeprefix("zenodo.")
-            current_metadata = self.zenodo_api.get_record_metadata(self.record_id)
+            # If a record ID is provided, get it's latest Zenodo version and metadata
```
michellewang: Typo
```suggestion
            # If a record ID is provided, get its latest Zenodo version and metadata
```

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -81,6 +84,48 @@ def _get_pipeline_metadata(
...
+        """Confirm upload with the user."""
+        if self.assume_yes:
+            logger.debug("Assuming yes to all prompts (--assume-yes flag).")
+            return
+
+        if not CONSOLE_STDOUT.is_interactive or not CONSOLE_STDOUT.is_terminal:
+            raise ExecutionError(
+                "Non-interactive terminal detected."
+                " Use the --assume-yes flag to bypass this prompt."
+            )
```
michellewang: Should we move this inside `CONSOLE_STDOUT.confirm`? E.g. if we call `CONSOLE_STDOUT.confirm(message, flag)` where if `flag` is not `None` we raise the error. Since it could be useful in other commands as well

nipoppy/workflows/pipeline_store/upload.py
```diff
@@ -81,6 +84,48 @@ def _get_pipeline_metadata(
...
+        if self.community:
+            self.zenodo_api.request_community_inclusion(
+                record_id=record_id,
+                community_id=self.zenodo_api._get_community_id("nipoppy"),
+            )
+        else:
+            logger.info(
+                "Use the --community flag to request inclusion in the Nipoppy Zenodo "
+                "community."
+            )
```
michellewang, line 121: For better user experience, we should also add the Zenodo ID since just rerunning with `--community` causes an error

```suggestion
            logger.info(
                f'Rerun with "--community --zenodo-id {record_id}" to request inclusion in the Nipoppy Zenodo '
                "community."
            )
```

nipoppy/zenodo_api.py
```diff
@@ -380,16 +370,76 @@ def _get_api_endpoint(self, community_id: str | None = None) -> str:
...
-        return response.json()["metadata"]
+        return response.json()
+
+    def _get_files_checksum(self, record_id: str) -> dict[str, str]:
+        """Get the MD5 checksums of the files in a Zenodo record."""
+        files = self._get_record(record_id)["files"]
+        return {file["key"]: file["checksum"].removeprefix("md5:") for file in files}
+
+    def _get_community_id(self, community: str) -> str:
+        """Resolve a community slug or ID to is uuid."""
```
michellewang: Typo
```suggestion
        """Resolve a community slug or ID to its uuid."""
```

nipoppy/zenodo_api.py
```diff
@@ -380,16 +370,76 @@ def _get_api_endpoint(self, community_id: str | None = None) -> str:
...
+            )
+            return
+
+        if response.status_code != 200 or not response_json.get("processed") or errors:
+            raise ZenodoAPIError(
+                f"Failed to request inclusion of zenodo.{record_id} in community "
+                f"{community_id}: {response_json}"
+            )
+
+        self.logger.info(
```
michellewang: Log as success?

```suggestion
        self.logger.success(
```

tests/unit/workflows/pipeline_store/test_upload.py
```diff
@@ -350,22 +331,77 @@ def test_force_bypasses_unchanged_check(
...
-@pytest.mark.no_xdist
-def test_confirm_upload_no(
-    workflow: PipelineUploadWorkflow,
-    caplog: pytest.LogCaptureFixture,
-    mocker: pytest_mock.MockerFixture,
-):
-    mocker.patch(
-        "nipoppy.workflows.pipeline_store.upload.CONSOLE_STDOUT.confirm",
-        return_value=False,
+class TestConfirmUpload:
```
michellewang, line 333: Some of these could be moved to the console tests if we integrate the checks in the console `confirm` instead

And I understand the use of a class here but I feel we should do a big refactor to classes for tests instead of having some as functions, some as classes. Or at least by module, e.g. the fileops tests are all classes

michellewang (changes requested):
Mostly minor changes, and a question about the location of the console interactivity checks: I think that should go in `_Console.confirm`

nipoppy/workflows/pipeline_store/upload.py
mathdugre: Moved the logic to the _Console.confirm function. However, I'm not sure I understand the rest.

tests/unit/workflows/pipeline_store/test_upload.py
mathdugre: I think for new tests we should use classes when it makes sense. This will prevent some extra work in the future

nipoppy/zenodo_api.py
mathdugre: Reverting as we cannot guarantee the logger used by the ZenodoAPI has a `success` function defined.

nipoppy/workflows/pipeline_store/upload.py
michellewang: So I was thinking of the case where there would be no `--assume-yes` flag exposed in the CLI or if the flag was something different (e.g. if in a command we need to prompt the user for confirmation for two different things twice, then we may want separate flags for yes/no?). Probably I am overthinking this, but we should at least check how `_Confole.confirm` is called in the current workflows and make sure this new logic doesn't break things.

michellewang (approved):
@mathdugre all looks good. See my reply to the comment thread about `_Console.confirm`. Curious to see what you think, but we can open a new issue if we decide that makes sense

PR #1128, opened 2026-09-08
[MAINT] Stop using the Boutiques Python API

michellewang:
- Closes none

## Changes
- Only use the `bosh` CLI via Python subprocess
- Update `nipoppy.workflows.base._run_command` to allow output capturing, which is used for the default invocation file creation and for validation error reporting
    - Not entirely satisfied with the implementation of the validation error reporting. Ideally we would rely on the existing logger functionalities, but because we always log both stdout and stderr that means there will be an `OK` logged with e.g., valid descriptors
- Currently working with latest official Boutiques. Need to test with [github.com/childmindresearch/boutiques_next](https://github.com/childmindresearch/boutiques_next)

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1128.org.readthedocs.build/en/1128/

### DIFF
--- a/nipoppy/pipeline_validation.py
+++ b/nipoppy/pipeline_validation.py
@@ -4,7 +4,6 @@ import json
 import logging
 from pathlib import Path
 
-import boutiques
 from pydantic_core import ValidationError
 
 from nipoppy.config.hpc import HpcConfig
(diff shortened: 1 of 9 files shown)


michellewang:
Closing in favour of #1167

PR #1131, opened 2026-09-08
[MAINT] Harmonize null checks to use is None or is not None

### LINKED ISSUE #937: [MAINT] Harmonize use of `if X is not None` vs. ` if X`

mathdugre:
- Closes #937 

**Question**
Should we still use the long form (`is not None`) when check `getattr(..., None)`?
e.g. changes nipoppy/workflows/dataset_init.py

**AI use**: GPT-5.6 Luna max did the changes and I reviewed them.

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1131.org.readthedocs.build/en/1131/

### DIFF
--- a/nipoppy/cli/cli.py
+++ b/nipoppy/cli/cli.py
@@ -108,7 +108,7 @@ def cli():
     """Organize and process neuroimaging-clinical datasets."""
 
 
-if cli.commands.get("gui"):
+if cli.commands.get("gui") is not None:
     cli.commands["gui"].hidden = True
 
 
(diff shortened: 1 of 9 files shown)


michellewang (approved):
🧑‍🍳

PR #1132, opened 2026-09-08
[BUG] Fail early when manifest is empty and using BIDS source

### LINKED ISSUE #897: [BUG] `nipoppy init --bids-source` should print error if generated manifest is empty

mathdugre:
- Closes #897

## Changes
- Raise error when manifest is empty from bids initialization.
- Init workflow now creates the bids and check the manifest first in the workflow.
- fileops.copy creates the parent dir to prevent failure.

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1132.org.readthedocs.build/en/1132/

### DIFF
--- a/nipoppy/workflows/dataset_init.py
+++ b/nipoppy/workflows/dataset_init.py
@@ -317,7 +317,7 @@ class InitWorkflow(BaseDatasetWorkflow):
         manifest = Manifest(df).validate()
         if manifest.empty:
             raise WorkflowError(
-                "Cannot initialize an empty manifest: no subjects found in BIDS source "
+                "No subjects found in BIDS source "
                 f"directory {self.bids_source}. Expected {BIDS_SUBJECT_PREFIX}* "
                 "directories directly inside it."
             )
--- a/nipoppy/workflows/dataset_init.py
+++ b/nipoppy/workflows/dataset_init.py
@@ -16,7 +16,7 @@ from nipoppy.env import (
     PipelineTypeEnum,
     StrOrPathLike,
 )
-from nipoppy.exceptions import FileOperationError
+from nipoppy.exceptions import FileOperationError, WorkflowError
 from nipoppy.logger import get_logger
 from nipoppy.tabular.manifest import Manifest
 from nipoppy.utils import fileops
@@ -79,11 +79,26 @@ class InitWorkflow(BaseDatasetWorkflow):
         """
         self._validate_study_root()
 
+        if self.bids_source is None:
+            fileops.copy(
+                FPATH_SAMPLE_MANIFEST,
+                self.study.layout.fpath_manifest,
+                exist_ok=True,
+                dry_run=self.dry_run,
+            )
+            logger.warning(
+                f"Sample manifest file copied to {self.study.layout.fpath_manifest}. "
+                "It should be edited to match your dataset."
+            )
+        else:
+            self._handle_bids_source()
+            self._init_manifest_from_bids_dataset()
+
         # create directories
         fileops.mkdir(self.dpath_root / NIPOPPY_DIR_NAME, dry_run=self.dry_run)
         for dpath in self.study.layout.get_paths(directory=True, include_optional=True):
             if self.bids_source is not None and dpath == self.study.layout.dpath_bids:
-                self._handle_bids_source()
+                continue
             elif (
                 self.container_store is not None
                 and dpath == self.study.layout.dpath_containers
@@ -106,21 +121,6 @@ class InitWorkflow(BaseDatasetWorkflow):
 
         # config file
         self._create_config_file()
-
-        # manifest
-        if self.bids_source is not None:
-            self._init_manifest_from_bids_dataset()
-        else:
-            fileops.copy(
-                FPATH_SAMPLE_MANIFEST,
-                self.study.layout.fpath_manifest,
-                exist_ok=True,
-                dry_run=self.dry_run,
-            )
-            logger.warning(
-                f"Sample manifest file copied to {self.study.layout.fpath_manifest}. "
-                "It should be edited to match your dataset."
-            )
 
         # copy dataset description file if specified in layout
         if getattr(self.study.layout, "fpath_bids_dataset_description", None):
@@ -315,6 +315,12 @@ class InitWorkflow(BaseDatasetWorkflow):
         df[Manifest.col_visit_id] = df[Manifest.col_session_id]
 
         manifest = Manifest(df).validate()
+        if manifest.empty:
+            raise WorkflowError(
+                "Cannot initialize an empty manifest: no subjects found in BIDS source "
+                f"directory {self.bids_source}. Expected {BIDS_SUBJECT_PREFIX}* "
+                "directories directly inside it."
+            )
         manifest.save_with_backup(
             self.study.layout.fpath_manifest, dry_run=self.dry_run
         )
--- a/tests/unit/workflows/test_init.py
+++ b/tests/unit/workflows/test_init.py
@@ -1,6 +1,7 @@
 """Tests for the dataset init workflow."""
 
 import os
+import re
 import shutil
 import subprocess
 from collections.abc import Generator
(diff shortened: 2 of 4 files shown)


nipoppy/workflows/dataset_init.py
```diff
@@ -315,6 +315,12 @@ def _init_manifest_from_bids_dataset(self) -> None:
         df[Manifest.col_visit_id] = df[Manifest.col_session_id]
 
         manifest = Manifest(df).validate()
+        if manifest.empty:
+            raise WorkflowError(
+                "Cannot initialize an empty manifest: no subjects found in BIDS source "
```
michellewang: No manifest is created, so I think the error can be just about the empty BIDS dir instead:

```suggestion
                "No subjects found in BIDS source "
```

tests/unit/workflows/test_init.py
```diff
@@ -155,10 +169,32 @@ def assert_config_matches(fpath_actual: Path, fpath_expected: Path):
...
+    dpath_root: Path,
+    caplog: pytest.LogCaptureFixture,
+    fname_layout: str,
+    dry_run: bool,
+):
+    workflow = InitWorkflow(
+        dpath_root=dpath_root,
+        fpath_layout=DPATH_LAYOUTS / fname_layout,
+        dry_run=dry_run,
+    )
```
michellewang, line 187: We should use the `workflow` fixture and set `workflow.fpath_layout = DPATHS_LAYOUT / fname_layout` and other args. Unless that doesn't work? And also remove `dpath_root` from the args and just use `workflow.dpath_root` inside the test

tests/unit/workflows/test_init.py
```diff
@@ -451,7 +493,12 @@ def test_init_bids(
     - handle_bids_source is called
     - README has been created
     """
-    workflow.bids_source = fake_bids_root
+    workflow = InitWorkflow(
+        dpath_root=dpath_root,
+        bids_source=fake_bids_root,
+        mode=mode,
+        fpath_layout=DPATH_LAYOUTS / fname_layout,
+    )
```
michellewang, line 501: Same comment here? It's what we do in the tests for all the other workflows

tests/unit/workflows/test_init.py
```diff
@@ -465,6 +512,41 @@ def test_init_bids(
     assert "Sample manifest file copied" not in caplog.text
 
 
+@pytest.mark.parametrize("mode", ["copy", "move", "symlink"])
+@pytest.mark.parametrize("nested", [False, True], ids=["empty", "nested-bids"])
```
michellewang, line 516: I feel the `nested` is not very useful here. Testing empty dir is enough, no?

tests/unit/workflows/test_init.py
```diff
@@ -465,6 +512,41 @@ def test_init_bids(
     assert "Sample manifest file copied" not in caplog.text
 
 
+@pytest.mark.parametrize("mode", ["copy", "move", "symlink"])
```
michellewang, line 515: Not useful/needed here since we have separate tests for the different modes

michellewang (changes requested):
I propose a different error message, and the tests have redundancy with existing tests and should be using the workflow fixture

tests/unit/workflows/test_init.py
mathdugre: Currently, the layout cannot change after instantiation of the InitWorkflow.

tests/unit/workflows/test_init.py
mathdugre: Same as above

mathdugre:
The current version of [_init_manifest_from_bids_dataset](https://github.com/mathdugre/nipoppy/blob/f994a50e22a52845fe8dc6df5dc57a7c28ed31bb/nipoppy/workflows/dataset_init.py#L237) will add an entry to the manifest for an empty datatype folder (e.g. `./bids/sub-1/ses-01/anat).

@michellewang , do we still want this behavior, or should we check some files exist in that dir?

michellewang:
> The current version of [_init_manifest_from_bids_dataset](https://github.com/mathdugre/nipoppy/blob/f994a50e22a52845fe8dc6df5dc57a7c28ed31bb/nipoppy/workflows/dataset_init.py#L237) will add an entry to the manifest for an empty datatype folder (e.g. `./bids/sub-1/ses-01/anat).
> 
> @michellewang , do we still want this behavior, or should we check some files exist in that dir?

@mathdugre I think it makes sense to check that files exist. Can be done in either this PR or another one

nipoppy/workflows/dataset_init.py
```diff
@@ -284,28 +284,29 @@ def _init_manifest_from_bids_dataset(self) -> None:
...
+                        x.name
+                        for x in (dpath_participant_session).iterdir()
+                        if x.is_dir() and any(child.is_file() for child in x.iterdir())
+                    ]
+                )
+
+                if len(datatypes) == 0:
+                    # Skip datatype folder without any files in it.
+                    # e.g. empty anat folder
+                    continue
```
michellewang: Should we emit a warning if that happens?

tests/unit/workflows/test_init.py
```diff
@@ -512,36 +512,29 @@ def test_init_bids(
     assert "Sample manifest file copied" not in caplog.text
 
 
-@pytest.mark.parametrize("mode", ["copy", "move", "symlink"])
-@pytest.mark.parametrize("nested", [False, True], ids=["empty", "nested-bids"])
-def test_init_bids_empty_manifest(
+@pytest.mark.parametrize(
+    "nested", ["", "sub-01/ses-1/anat"], ids=["empty", "nested-bids"]
```
michellewang: We could use `"subdirs"` instead of `"nested"` here for clarity

tests/unit/workflows/test_init.py
```diff
@@ -465,6 +512,34 @@ def test_init_bids(
...
+)
+def test_init_manifest_from_empty_bids_fails(
+    workflow: InitWorkflow,
+    dpath_root: Path,
+    caplog: pytest.LogCaptureFixture,
+    nested: str,
+):
+    bids_source = dpath_root / "bids" / nested
+    bids_source.mkdir(parents=True)
+    # workflow.study.layout.dpath_bids = bids_source
```
michellewang:
```suggestion
```

michellewang (changes requested):
Minor suggestions and a question

nipoppy/workflows/dataset_init.py
mathdugre: Good idea. I'll replace the comment with a logged warning

tests/unit/workflows/test_init.py
```diff
@@ -512,9 +512,7 @@ def test_init_bids(
     assert "Sample manifest file copied" not in caplog.text
 
 
-@pytest.mark.parametrize(
-    "nested", ["", "sub-01/ses-1/anat"], ids=["empty", "nested-bids"]
-)
+@pytest.mark.parametrize("nested", ["", "sub-01/ses-1/anat"], ids=["empty", "subdirs"])
```
michellewang: What I meant is that I feel `nested` is not a very clear variable name. Do you think `subdirs=""` / `subdirs="sub-01/ses-1/anat"` is better?

michellewang (changes requested):
One last thing, I think I wasn't clear in a previous comment

michellewang (approved):
🧑‍🍳

PR #1148, opened 2026-09-16
[MAINT] Use fileops module across the code base.

### LINKED ISSUE #851: [MAINT] Harmonize use of `Path.mkdir()` calls within the codebase (favor fileops class)
There are a few instances of `Path.mkdir()` calls within the codebase, in:
- `container.py`
- `config/main.py`
- `utils/utils.py`
- `workflows/pipeline.py`
- `workflows/pipeline_store/create.py`

Would it make sense to use `fileops` in some of those as well?

_Originally posted by @michellewang in https://github.com/nipoppy/nipoppy/pull/834#discussion_r2569442292_

mathdugre:
- Closes #851 

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1148.org.readthedocs.build/en/1148/


mathdugre:
We might want to consider creating a new `fileops.unlink` function.

Currently used here:
- https://github.com/mathdugre/nipoppy/blob/e8610a2b37aab6ef5df4cd0e414d3ef93a559cf1/nipoppy/workflows/services/hpc.py#L257
- https://github.com/mathdugre/nipoppy/blob/e8610a2b37aab6ef5df4cd0e414d3ef93a559cf1/nipoppy/utils/utils.py#L211

If it's relevant, we should open a new issue to create the function and make the changes across the code.

nipoppy/logger.py
```diff
@@ -115,7 +115,9 @@ def add_file_handler(self, file: Path) -> Self:
         # Only one file handler allowed
         self._cleanup_handler(self._file_handler)
 
-        file.parent.mkdir(parents=True, exist_ok=True)
+        from nipoppy.utils import fileops
```
michellewang, line 118: Why is the import in the middle of the method and not at the top? It's not a heavy module so I think it's better to import at the top even if only used once.

nipoppy/utils/utils.py
```diff
@@ -122,10 +122,12 @@ def save_json(obj: dict, fpath: StrOrPathLike, **kwargs):
     **kwargs :
         Keyword arguments to pass to json.dump
     """
+    from nipoppy.utils import fileops
```
michellewang, line 125: Same comment about importing at the top of the file

michellewang (changes requested):
Looks good except for the imports that are inside functions/methods instead of at the top of the file

nipoppy/logger.py
mathdugre: It needs to be a local import to prevent circular import

nipoppy/utils/utils.py
mathdugre: Same as above.

michellewang (changes requested):
Okay for the import statements staying where they are. Could you add brief comments explaining that this is to avoid circular imports?

michellewang (approved):
🧑‍🍳

Issue #460, opened 2025-01-08
[ENH] Use manifest's `datatype` column to filter participants to run


michellewang:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### New feature

So that e.g. diffusion pipelines won't be run for participants/sessions without `dwi` data

Then we would also be able to use the `UNAVAILABLE` status in the tracker

PR #1150, opened 2026-09-16
[ENH] filtering by `--datatype` for `nipoppy status`

### LINKED ISSUE #348: [ENH] Make doughnut workflow check datatype availability in BIDS directory

- [X] I have searched the existing issues

### New feature

- Would be a subset of what is in the respective `datatype` column in the manifest based on actual data on disk
- Related to #235 (though with less flexibility)

### Unclear documentation

_No response_

mathdugre:
- Closes #348 

## Changes
- Add a `--datatype` flag to the `nipoppy status` command. This allows filtering by datatype, using the manifest as the source of truth for datatypes per subject-session pairs 

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1150.org.readthedocs.build/en/1150/


michellewang:
Was not aware that `Poppy status` was a command 😉

docs/source/reference/cli_reference/status.rst
```diff
@@ -4,5 +4,10 @@
 .. note::
    This command calls the :py:class:`nipoppy.workflows.dataset_status.StatusWorkflow` class from the Python :term:`API` internally.
 
+Use ``--datatype`` to restrict the status table to imaging participant-session
+pairs whose manifest ``datatype`` list contains the exact requested BIDS
+datatype, for example ``nipoppy status --datatype dwi``. The manifest is the
+source of truth for the selection.
+
```
michellewang: Redundant with the CLI help

```suggestion
```

nipoppy/cli/__init__.py
```diff
@@ -30,8 +30,8 @@ def exception_handler(workflow):
             "Troubleshooting: Review your configuration fields and value types."
         )
     except SystemExit as e:
-        workflow.return_code = e.code or ReturnCode.UNKNOWN_FAILURE
-        logger.error(e)
+        workflow.return_code = ReturnCode.UNKNOWN_FAILURE if e.code is None else e.code
+
```
michellewang, line 34: Why did `logger.error(e)` get removed?

nipoppy/cli/cli.py
```diff
@@ -285,6 +286,15 @@ def extract(**params):
 
 @cli.command()
 @dataset_option
+@click.option(
+    "--datatype",
+    type=str,
+    help=(
+        "Only include imaging manifest rows whose datatype list contains this exact "
+        "BIDS datatype. Curation and processing counts use the selected "
+        "participant-session pairs."
```
michellewang, line 295:
```suggestion
        "Only include imaging manifest rows whose datatype list contains this exact "
        "datatype."
```

nipoppy/workflows/dataset_status.py
```diff
@@ -102,6 +105,25 @@ def _check_manifest(self, status_df: pd.DataFrame) -> pd.DataFrame:
...
+                    lambda datatypes: (
+                        isinstance(datatypes, list) and self.datatype in datatypes
+                    )
+                )
+            ]
+            self._participant_session_pairs = set(
+                imaging_manifest[
+                    [manifest.col_participant_id, manifest.col_session_id]
+                ].itertuples(index=False, name=None)
+            )
```
michellewang, line 120: Would it be better to use `imaging_manifest.get_participants_sessions()` here instead? It's basically the same logic

nipoppy/workflows/dataset_status.py
```diff
@@ -102,6 +105,25 @@ def _check_manifest(self, status_df: pd.DataFrame) -> pd.DataFrame:
 
         # filter participants with imaging data
         imaging_manifest = manifest.get_imaging_subset()
```
michellewang, line 107: We should probably update `Manifest.get_imaging_subset()` to accept a `datatype` arg instead. Could be useful e.g. for #460

nipoppy/workflows/dataset_status.py
```diff
@@ -102,6 +105,25 @@ def _check_manifest(self, status_df: pd.DataFrame) -> pd.DataFrame:
...
+            self._participant_session_pairs = set(
+                imaging_manifest[
+                    [manifest.col_participant_id, manifest.col_session_id]
+                ].itertuples(index=False, name=None)
+            )
+            if imaging_manifest.empty:
+                logger.warning(
+                    f"No imaging manifest rows matched datatype '{self.datatype}'."
+                )
+                raise SystemExit(ReturnCode.SUCCESS.value)
```
michellewang, line 125: I don't think we should do `SystemExit` in a workflow since it will be very annoying if someone is only using the Python API. It might make more sense to raise a `WorkflowError`?

nipoppy/workflows/dataset_status.py
```diff
@@ -36,6 +38,8 @@ def __init__(
             _skip_logfile=True,
         )
         self.col_pipeline = "pipeline"
+        self.datatype = datatype
+        self._participant_session_pairs: set[tuple[str, str]] | None = None
```
michellewang, line 42: Would it be better to explicitly pass the `participant_session_pairs` around instead of keeping track of it in the object state?

nipoppy/workflows/dataset_status.py
```diff
@@ -36,6 +38,8 @@ def __init__(
...
 
     def run_main(self):
         """Check the status of the dataset and report.
```
michellewang, line 45: We might want to update the docstring a bit. Curation status is always available. And the counts are not per BIDS datatype (even after this PR I don't think I would qualify it as "per BIDS datatype")

nipoppy/workflows/dataset_status.py
michellewang, line 54: What do you think of having the logic be more explicit, something like this:

```python
participant_session_pairs = self._get_participant_session_pairs(datatype)
status_df = self._check_manifest(status_df, participant_session_pairs)
status_df, curation_cols = self._check_curation_status_table(status_df, participant_session_pairs)
status_df, processing_cols = self._check_processing_status_table(status_df, participant_session_pairs)
```

I feel that would would be cleaner? But really now I am leaning towards a rewrite so that we build a large, long-form dataframe with `participant_id`, `session_id`, `status_col`, then filter it by the datatype and use `groupby`/`count` to get the counts. Do you think that would work/be better?

michellewang (changes requested):
I am not a fan of the implementation in the `StatusWorkflow`... Most of that is due to the existing code though. Do you think it would be worth reimplementing to make the filtering easier? See more specific comment below

PR #1152, opened 2026-09-16
[ENH] Add --skip-container option for pipeline install

### LINKED ISSUE #1147: [BUG] No way to refuse container installation in `nipoppy pipeline install`

### Describe the bug

The tests on `neurodatascience/nipoppy-pipeline-configs` are failing from #1104 being merged: https://github.com/neurodatascience/nipoppy-pipeline-configs/actions/runs/35102683386/job/104815791344

The cause is because the test [mocks stdin to give "n" as reply to the confirmation prompt](https://github.com/neurodatascience/nipoppy-pipeline-configs/blob/ed4d0e28043a9eb4f1e52a651c4336c2b7d50c15/tests/test_pipelines.py#L69), but now the environment is being detected as non-interactive so `_Console.confirm` will error out. @mathdugre what do you think of allowing `--assume-no` in addition to `--assume-yes` as CLI option here? The alternative is to change the test logic, but I think it could be useful to have `--assume-no` at least in this case.

### To reproduce

Run tracker tests in https://github.com/neurodatascience/nipoppy-pipeline-configs/blob/ed4d0e28043a9eb4f1e52a651c4336c2b7d50c15/tests/test_pipelines.py#L240

### Environment

Latest dev version of Nipoppy

mathdugre:
- Closes #1147 

## Changes
- Add `--skip-container` flag to the `nipoppy pipeline install` command.

## Checklist

### All PRs
- [ ] Tests pass
- [ ] Checks pass

### If new feature
- [ ] Tests have been added

### If bug fix

----
📚 Documentation preview 📚: https://nipoppy--1152.org.readthedocs.build/en/1152/


nipoppy/cli/cli.py
```diff
@@ -86,6 +86,7 @@ def decorator(f):
             "options": [
                 "--force",
                 "--assume-yes",
+                "--skip-container",
```
michellewang: I would tend to put this in "Command-specific"...

nipoppy/cli/pipeline_catalog.py
```diff
@@ -124,6 +124,11 @@ def pipeline_create(**params):
     is_flag=True,
     help="Overwrite existing pipeline directory if it exists.",
 )
+@click.option(
+    "--skip-container",
+    is_flag=True,
+    help="Skip downloading the pipeline container.",
+)
```
michellewang, line 134: Should we add here that `--skip-container` is mutually exclusive with `--assume-yes`? For documentation

michellewang: Or I guess we can be less strict and if both are given, we emit a warning and skip the container download

michellewang (commented):
Mostly looks fine. Some minor questions

nipoppy/cli/cli.py
mathdugre: Sure, no preference on my side

nipoppy/cli/pipeline_catalog.py
mathdugre: I think I prefer adding it here. It will simplify the logic in the implementation.

nipoppy/cli/pipeline_catalog.py
```diff
@@ -124,6 +124,14 @@ def pipeline_create(**params):
     is_flag=True,
     help="Overwrite existing pipeline directory if it exists.",
 )
+@click.option(
+    "--skip-container",
+    is_flag=True,
+    help=(
+        "Skip downloading the pipeline container "
+        "(mutually exclusive with --container-store)."
```
michellewang:
```suggestion
        "(mutually exclusive with --assume-yes)."
```

nipoppy/workflows/pipeline_store/install.py
```diff
@@ -39,8 +39,14 @@ def __init__(
         fpath_layout: StrOrPathLike | None = None,
         verbose: bool = False,
         dry_run: bool = False,
+        skip_container: bool = False,
     ):
         """Initialize the workflow."""
+        if assume_yes and skip_container:
+            raise WorkflowError(
```
michellewang, line 46: Set `self.return_code = ReturnCode.INVALID_COMMAND`?

michellewang (changes requested):
Minor comments

nipoppy/cli/pipeline_catalog.py
michellewang: @mathdugre this hasn't been resolved

michellewang (changes requested):
Just a leftover comment from the previous review that hasn't actually been resolved

michellewang (approved):
🧑‍🍳

